PR #2, opened 2015-12-09
Rdfstore

sandervd:


web/modules/custom/rdf_entity/src/Entity/RdfEntitySparqlStorage.php
```diff
@@ -0,0 +1,433 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\Core\Entity\ContentEntityNullStorage.
```
pfrenssen: `RdfEntitySparqlStorage`.

web/modules/custom/rdf_entity/src/Form/OverviewRdf.php
```diff
@@ -0,0 +1,116 @@
...
+use Drupal\Core\Entity\EntityManagerInterface;
+use Drupal\Core\Form\FormBase;
+use Drupal\Core\Extension\ModuleHandlerInterface;
+use Drupal\Core\Form\FormStateInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Provides terms overview form for a taxonomy vocabulary.
+ */
+class OverviewRdf extends FormBase {
```
pfrenssen: It looks like we can throw away this class. It was initially copied from the Taxonomy module, but it doesn't seem to do anything useful.

PR #10, opened 2016-02-09
ISAICP-2303

pfrenssen:
https://webgate.ec.europa.eu/CITnet/jira/browse/ISAICP-2303


pfrenssen:
I cherry-picked one of the commits from ISAICP-2191, so that the composer dependencies can be built without errors - one of our patches has been accepted upstream in the Coder module, so the patch needed to be removed from `composer.json`.

PR #16, opened 2016-02-23
Isaicp 2205

idimopoulos:


composer.json
```diff
@@ -26,6 +26,7 @@
         "drupal/devel": "dev-8.x-1.x",
         "drupal/drupal-extension": "dev-assert-not-button as 3.1.6",
         "drupal/phing-drush-task": "1.0",
+        "drupal/search_api": "8.1.0-alpha12",
```
pfrenssen: This hardcodes the alpha12 release. If you put a tilde in front of the version string, then we will also benefit from updates when they are released: `~8.1.0-alpha12`.

This should also be moved to the "require" section. It is now in "require-dev" which is intended for development environments only. Search API is something we need on acceptance, staging and production as well.

composer.json
```diff
@@ -54,6 +55,11 @@
             "web/profiles/contrib/{$name}": ["type:drupal-profile"],
             "web/themes/contrib/{$name}": ["type:drupal-theme"],
             "drush/contrib/{$name}": ["type:drupal-drush"]
+        },
+        "patches": {
+            "drupal/search_api": {
+                "Issue #2671228: Support for non Numeric IDs": "https://www.drupal.org/files/issues/support_for_non_numeric-2671228-2_1.patch"
```
pfrenssen: A newer version of your patch is available on that issue, let's use the latest version.

tests/features/collection/collection_overview.feature
```diff
@@ -0,0 +1,16 @@
...
+      | Open Data         | Facilitate access to data sets | Madame Sharn    | http://joinup.eu/coll1     |
+      | Connecting Europe | Reusable tools and services    | Ptaclusp IIb    | http://joinup.eu/coll2     |
+    And The collections are indexed
+    When I visit the collection overview page
+    Then I should see the link "eHealth"
+    And I should see the text "Supports health-related fields"
+    And I should see the link "Open Data"
+    And I should see the text "Facilitate access to data sets"
+    And I should see the link "Connecting Europe"
+    And I should see the text "Reusable tools and services"
```
pfrenssen: Missing newline at the end.

tests/features/collection/collection_overview.feature
```diff
@@ -0,0 +1,16 @@
+@api
+Feature: Collections Overview
+  Scenario: View collection overview
+    Given collections:
+      | name              | description                    | author          | uri                        |
+      | eHealth           | Supports health-related fields | Arnold Sideways | http://drupal.org/notfound |
+      | Open Data         | Facilitate access to data sets | Madame Sharn    | http://joinup.eu/coll1     |
+      | Connecting Europe | Reusable tools and services    | Ptaclusp IIb    | http://joinup.eu/coll2     |
+    And The collections are indexed
+    When I visit the collection overview page
```
pfrenssen: Great stuff, this is really good from the BDD perspective, it describes perfectly the intention of the user. In the original test that was proposed in ticket ISAICP-2205, this was saying "When I visit '/collections'" which is not so good, since it implies that users have to browse the site by memorizing URLs.

:+1:

web/modules/custom/collection/collection.behat.inc
```diff
@@ -83,6 +92,7 @@ class CollectionSubContext extends DrupalSubContextBase implements DrupalSubCont
       'uri' => 'id',
       'name' => 'label',
       'author' => 'author',
+      'description' => 'field_ar_description',
```
pfrenssen: Why is this called `field_ar_description`? What is "ar"?

web/modules/custom/collection/collection.behat.inc
```diff
@@ -157,6 +167,36 @@ class CollectionSubContext extends DrupalSubContextBase implements DrupalSubCont
   }
 
   /**
+   * Indexes items of entity type: collection.
+   *
+   * In addiction, views cache data are being cleared.
```
pfrenssen: s/addition/addiction/

web/modules/custom/collection/collection.behat.inc
```diff
@@ -157,6 +167,36 @@ class CollectionSubContext extends DrupalSubContextBase implements DrupalSubCont
...
+   *
+   * @throws \Exception
+   *   When no items are indexed.
```
pfrenssen: This is not really correct, the exception is thrown when the index doesn't exist, not when there are no items to index.

web/modules/custom/collection/collection.behat.inc
```diff
@@ -157,6 +167,36 @@ class CollectionSubContext extends DrupalSubContextBase implements DrupalSubCont
...
+    $index = \Drupal\search_api\Entity\Index::load($id);
+    if (empty($index)) {
+      throw new \Exception("Collection index not found.");
+    }
+    $index->reindex();
+    $index->indexItems();
+
+    // Clear the views cache data.
+    $view_tags = \Drupal\views\Views::getView($views_id)->getCacheTags();
+    \Drupal\Core\Cache\Cache::invalidateTags($view_tags);
```
pfrenssen: I don't understand this. Why do we manually need to clear the Views cache when new items are indexed? This should be taken care of automatically.

So in real life when items are indexed they do not appear in Views until you manually clear the cache? That seems like a bug to me.

web/profiles/joinup/config/install/search_api.index.collections.yml
```diff
@@ -0,0 +1,73 @@
+uuid: ac8e6731-1174-4f27-ab00-de719168b34b
```
pfrenssen: Don't commit the UUID. Are you using the Config Devel module to export config? That should take care of this automatically.

This config item should be added to the list in `joinup.info.yml`.

web/profiles/joinup/config/install/search_api.index.collections.yml
```diff
@@ -0,0 +1,73 @@
...
+    label: ID
+    datasource_id: 'entity:rdf_entity'
+    property_path: id
+    type: string
+  label:
+    label: Title
+    datasource_id: 'entity:rdf_entity'
+    property_path: label
+    type: string
+  field_ar_description:
```
pfrenssen: Here is that weird field again, this seems to belong to a different entity or bundle. This should be `field_collection_description`.

web/profiles/joinup/config/install/search_api.server.collections_search_server.yml
```diff
@@ -0,0 +1,18 @@
+uuid: 9fe2d8ff-5921-4ec5-a400-c5351a1499f4
```
pfrenssen: Don't export the UUID. Do add it to `joinup.info.yml`.

web/profiles/joinup/config/install/search_api.server.collections_search_server.yml
```diff
@@ -0,0 +1,18 @@
...
+langcode: en
+status: true
+dependencies:
+  module:
+    - search_api_db
+id: collections_search_server
+name: 'Collections Search Server'
```
pfrenssen: This seems to imply we are dedicating an entire server to just indexing collections. I need to refresh my memory on how Search API works, but I think this server can be used for indexing the entire project, so we can call it the "Joinup Search Server".

web/profiles/joinup/config/install/search_api.server.collections_search_server.yml
```diff
@@ -0,0 +1,18 @@
...
+description: ''
+backend: search_api_db
+backend_config:
+  database: 'default:default'
+  database_text: ''
+  min_chars: 3
+  partial_matches: false
```
pfrenssen: Partial matches is something we will probably need in the future, but you can leave this as is for now, it is out of scope of this ticket.

web/profiles/joinup/config/install/views.view.collection.yml
```diff
@@ -0,0 +1,449 @@
+uuid: e5498df7-0435-45dc-820a-10889814f600
```
pfrenssen: UUID bad, adding to `joinup.info.yml` good.

web/profiles/joinup/config/install/views.view.collection.yml
```diff
@@ -0,0 +1,449 @@
...
+      display_extenders: {  }
+      path: collection
+    cache_metadata:
+      max-age: 0
+      contexts:
+        - 'languages:language_content'
+        - 'languages:language_interface'
+        - url.query_args
+      tags:
+        - 'config:field.storage.rdf_entity.field_ar_description'
```
pfrenssen: I did not read this whole thing. Will instead try it out in the interface :)

composer.json
idimopoulos: I only put it in require-dev because this is a temporary solution. In the future, a requirement of the search_api_solr package will be included and this is going to download the search_api as well. I can move it but it is indeed meant for the development environment.

web/modules/custom/collection/collection.behat.inc
idimopoulos: 'ar' stands for Asset Repository which is another RDF enitty. We should change this but this is the name of the description field right now.

web/modules/custom/collection/collection.behat.inc
idimopoulos: I am not sure on this, it is just the way it worked for me. By default, I found that views save a result cache in "cache_data" table using the "config:views:view:<view_machine_name>" as tags. Even by disabling the cache in views settings, did not work for me when indexing items.
All in all, the test was successful at the first run, but the results would be cached so a manual cache clear or re-saving the view was needed. 
If you have any suggestion on this I am open but my tests would only work with this piece of code.

composer.json
pfrenssen: I can give some more background information. This is used by our continuous integration setup. ContinuousPHP only installs the `require-dev` packages for running the tests. When deploying they are running Composer with the `--no-dev` flag which omits `require-dev` for production builds. We are using these to deploy to the acceptance environments on Amazon. So this means everything which is in `require-dev` will not be there for doing UAT.

web/profiles/joinup/config/install/search_api.server.collections_search_server.yml
idimopoulos: Multiple servers can be created through UI and each having a separate index or indexes. So We can have multiple instances of local database servers or have a central one with multiple Indexes. It is all a matter of how we define it. What do you think?

web/modules/custom/collection/collection.behat.inc
pfrenssen: The tests should prove that it works as expected. This test implies that it is expected that the views cache is cleared manually after indexing data. That's not right, it should happen automatically. Here the test is adapted to the bugs, it should be the other way around.

Here's a good way of thinking about this: normally you should write your tests first, and then implement the functionality. If you didn't know beforehand that there was a caching problem after indexing, would you have written the test like this?

Now we need to be practical too. If this is a bug in Search API then it might take too much time to fix it right now. But then we need to make sure that there is a ticket for this on drupal.org, a ticket for us to follow up on it, and a clear comment here that explains why this workaround is needed, with links to the respective tickets.

I just tested it in the interface and the problem is very real. If I add 3 collections and visit the overview then I see the 3 items. If I add a 4th and go back to the overview then I don't see the new item, I see a cached version that contains the 3 old items. I'm afraid that it will likely be rejected by the client if it stays like this.

I think the most realistic approach for now is to disable caching on the view.

PR #18, opened 2016-02-25
Isaicp 2192

pfrenssen:
https://webgate.ec.europa.eu/CITnet/jira/browse/ISAICP-2192


tests/features/collection/og.feature
```diff
@@ -79,3 +80,27 @@ Feature: Organic Groups integration
...
+      | http://joinup.eu/collection/dog  | logo.png | Überwaldean Land Eels        | Read up on all about <strong>dogs</strong>  | http://dogtime.com/dog-breeds/profiles  | 1       | 28-01-1995 12:05  | The Afghan Hound is elegance personified.                                                                           | 0                  | 1           |                    |       |
+      | http://joinup.eu/collection/cat  | logo.png | Folk Dance and Song Society  | Cats are cool!                              | http://mashable.com/category/cats/      | 0       | 28-01-1995 12:06  | The domestic cat (Felis catus or Felis silvestris catus) is a small usually furry domesticated carnivorous mammal.  | 1                  | 0           |                    |       |
+    And users:
+      | name             | roles         |
+      | Collection admin | administrator |
+      | Madame Sharn     |               |
+      | Goodie Whemper   |               |
+    # Administrators can edit the collection.
+    # @fixme Run this as facilitator or collection owner.
+    Given I am logged in as "Collection admin"
```
pfrenssen: Add a link here to ISAICP-2190. We will add the roles in that issue, so this can be fixed very soon.

tests/features/collection/og.feature
```diff
@@ -79,3 +80,27 @@ Feature: Organic Groups integration
...
+    When I go to the homepage of the "Überwaldean Land Eels" collection
+    Then I should see the link "Edit"
+
+    # Edit a collection.
+    When I go to the "Überwaldean Land Eels" collection edit form
+    Then the following fields should be present "Title, Description, Abstract, Contact information, Owner, Policy domain, Topic, Spatial coverage, Affiliates, Closed collection, eLibrary creation, Moderated"
+    And I fill in "Title" with "Überwaldean Sea Eels"
+    And I press the "Save" button
+    Then I should see the text "Überwaldean Sea Eels"
```
pfrenssen: It's better to use the test "Then I should see the *heading* ..." instead - this will match on the page title which is a bit more reliable test to see that the collection was correctly saved.

web/modules/custom/collection/collection.behat.inc
```diff
@@ -288,9 +284,87 @@ class CollectionSubContext extends DrupalSubContextBase implements DrupalSubCont
     }
     // Also remove the image entities that were attached to the collections.
     foreach ($this->images as $image) {
-      file_unmanaged_delete($image->getFileUri());
```
pfrenssen: :+1:

web/modules/custom/collection/collection.behat.inc
```diff
@@ -288,9 +284,87 @@ class CollectionSubContext extends DrupalSubContextBase implements DrupalSubCont
...
+   *
+   * @param string $fields
+   *    Fields.
+   *
+   * @throws \Exception
+   *   Thrown when a column name is incorrect.
+   *
+   * @Then /^(?:|the following )fields should be present? "(?P<fields>[^"]*)"$/
+   */
+  public function assertFieldsPresent($fields) {
```
pfrenssen: This (and the `::assertFieldsNotPresent()` counterpart) is useful outside the scope of collections. I would move it to DrupalContext.

web/modules/custom/collection/collection.info.yml
```diff
@@ -11,19 +11,3 @@ dependencies:
...
-  - field.field.collection.collection.field_collection_logo
-  - field.field.collection.collection.field_collection_metadata_url
-  - field.field.collection.collection.field_collection_pre_moderation
-  - field.field.collection.collection.field_collection_schedule
-  - field.storage.collection.field_collection_closed
-  - field.storage.collection.field_collection_create_elibrary
-  - field.storage.collection.field_collection_logo
-  - field.storage.collection.field_collection_metadata_url
-  - field.storage.collection.field_collection_pre_moderation
-  - field.storage.collection.field_collection_schedule
```
pfrenssen: It seems like you have to be introduced to the awesomeness of the `config_devel` module. I'll show you a demo soon, you'll love it :)

web/modules/custom/collection/config/install/core.entity_form_display.rdf_entity.collection.default.yml
```diff
@@ -23,9 +23,10 @@ dependencies:
     - datetime
     - image
     - link
+    - rdf_entity
     - text
 _core:
-  default_config_hash: 7J5kiNcWwzxFV3dpv_JwbNVd6IlpWCqGBp7Y-FzPWGM
+  default_config_hash: fbC_Go0HP9SCUjTRv94g0KHkTip0Bor8FeOJzkymoBk
```
pfrenssen: This is a recent change in core, these `default_config_hash` parameters should not be committed since they are environment specific.

This is just as information, it does not affect the functionality of this ticket so it doesn't need to be addressed now. config_devel takes care of this.

composer.json
```diff
@@ -55,6 +55,11 @@
             "web/profiles/contrib/{$name}": ["type:drupal-profile"],
             "web/themes/contrib/{$name}": ["type:drupal-theme"],
             "drush/contrib/{$name}": ["type:drupal-drush"]
+        },
+        "patches": {
+            "drupal/core": {
+                "Enlarge file_usage entity id length.": "https://gist.githubusercontent.com/sandervd/884027d8fa7d5a14165a/raw/03d3460888dba36f37dc2f74ef36128dc4cf4a3d/file_usage_entity_id_length.patch"
+            }
```
pfrenssen: Let's contribute this patch on drupal.org.

pfrenssen:
I see that the description is a multivalue field. From earlier discussions it transpired that this is listed as such in the functional specification because this is intended to be translatable. In reality it should be a single value field.

Many of the fields are multivalue, so there might be more instances where these are actually intended to be translatable single value fields. We should ask PWC about this.

tests/features/collection/og.feature
```diff
@@ -79,3 +80,26 @@ Feature: Organic Groups integration
...
+    Given I am logged in as "Collection admin"
+    When I go to the homepage of the "Überwaldean Land Eels" collection
+    Then I should see the link "Edit"
+
+    # Edit a collection.
+    When I go to the "Überwaldean Land Eels" collection edit form
+    Then the following fields should be present "Title, Description, Abstract, Contact information, Owner, Policy domain, Topic, Spatial coverage, Affiliates, Closed collection, eLibrary creation, Moderated"
+    And I fill in "Title" with "Überwaldean Sea Eels"
+    And I press the "Save" button
+    Then I should see the heading "Überwaldean Sea Eels"
```
idimopoulos: A new line should be added.

pfrenssen:
Das ist schade.

web/modules/custom/rdf_entity/src/Plugin/Field/FieldWidget/AutoUriWidget.php
```diff
@@ -0,0 +1,88 @@
...
+      $prefix = $this->getSetting('prefix');
+      if (empty($prefix)) {
+        throw new Exception('Auto URI widget has is un-configured: Prefix not set.');
+      }
+      // If needed, append a trailing slash.
+      if (substr($prefix, -1) != '/') {
+        $prefix .= '/';
+      }
+      $uuid = new Php();
+      $element['value']['#value'] = $prefix . $uuid->generate();
```
idimopoulos: Does this mean that the uuid will be generated every time a user refreshes the page as well?
Shouldn't this be moved to submit form instead (or validate)?

web/modules/custom/rdf_entity/src/Plugin/Field/FieldWidget/DateTimeCreateUpdateWidget.php
```diff
@@ -0,0 +1,117 @@
...
+    $form_object = $form_state->getFormObject();
+    /** @var \Drupal\Core\Entity\EntityInterface $entity */
+    $entity = $form_object->getEntity();
+    $is_new = empty($entity->id());
+
+    $behaviour = $this->getSetting('behaviour');
+    if (($behaviour == 'create' && $is_new) || $behaviour == 'modified') {
+      $element['value'] = [
+        '#type' => 'value',
+        '#value' => date('c'),
```
idimopoulos: This is also being set when a user accesses the page, right? So, we are not recording the precise time of the create/update like this (am I wrong?).
Shouldn't this be moved to submit form instead (or validate)?

PR #21, opened 2016-02-29
Isaicp 2314

sandervd:


tests/features/user/api.feature
```diff
@@ -0,0 +1,21 @@
...
+      | First name           | Leonardo                                     |
+      | Family name          | di ser Piero da Vinci                        |
+      | Photo                | leonardo.jpg                                 |
+      | Professional profile | invention, painting, sculpting, architecture |
+      | Organisation         | Verrocchio's workshop                        |
+     # @Fixme unimplemented.
+     # | Professional domain  |                                              |
+     # | Nationality          |                                              |
+     # | Social network       |                                              |
+    Then I should have a "Leonardo Da Vinci" user
```
idimopoulos: New line does not exist at the end of the file.

web/profiles/joinup/config/install/user.role.user.yml
```diff
@@ -0,0 +1 @@
+{  }
```
idimopoulos: New line is shown to be required at the end of the file.

idimopoulos: Also, why do we need an empty user role file?

web/profiles/joinup/joinup.info.yml
```diff
@@ -78,8 +98,11 @@ config_devel:
   - og.og_membership_type.og_membership_type
   - og.settings
   - system.theme
+  - user.flood
+  - user.mail
   - user.role.administrator
   - user.role.anonymous
   - user.role.authenticated
   - user.role.moderator
   - user.role.user
```
idimopoulos: What is the use of the "user" role since we are using authenticated for this purpose?

web/profiles/joinup/joinup.behat.inc
```diff
@@ -0,0 +1,163 @@
...
+   * Test images.
+   *
+   * @var \Drupal\file\Entity\File[]
+   */
+  protected $images = [];
+
+  /**
+   * Creates a user with data provided in a table.
+   *
+   * The following fields are required:
```
idimopoulos: Maybe change this to table format as it is in modules/custom/collection/collection.behat.inc marking the required fields with an asterisk or something similar.
This is only to keep the same way of writing comments and keep it generic rather than personalized comments.

web/profiles/joinup/joinup.behat.inc
```diff
@@ -0,0 +1,163 @@
+<?php
+
+/**
+ * @file
+ * Contains \UserSubContext.
```
idimopoulos: Maybe this class should move to a different file or be renamed to JoinupSubContext in order to give the opportunity to add code for more core functionalities such as User entity.

web/profiles/joinup/joinup.behat.inc
```diff
@@ -0,0 +1,163 @@
...
+   * @param array $values
+   *   An optional associative array of values, keyed by property name.
+   *
+   * @return \Drupal\user\Entity\User
+   *   A new user entity.
+   *
+   * @throws \Exception
+   *   Thrown when a given image is not found.
+   */
+  public function createUser(array $values) {
```
idimopoulos: You are overriding the createUser function here. Is this because it does not handle the image upload correctly?

web/profiles/joinup/joinup.behat.inc
```diff
@@ -0,0 +1,163 @@
...
+      throw new \Exception("Unable to load expected user " . $username);
+    }
+  }
+
+  /**
+   * Remove any created users.
+   *
+   * @AfterScenario
+   */
+  public function cleanUsers() {
```
idimopoulos: There is already a cleanUsers inherited here but it expectes stdObjects rather than User objects.

PR #22, opened 2016-02-29
Isaicp 2191

pfrenssen:
https://webgate.ec.europa.eu/CITnet/jira/browse/ISAICP-2191


web/profiles/joinup/config/install/user.role.authenticated.yml
```diff
@@ -7,8 +7,7 @@ weight: 1
 is_admin: false
 permissions:
   - 'access content'
-  - 'create collection rdf entity'
```
idimopoulos: You removed this permission. Please, also update the authentication.feature behat tests because it fails while testing for access for the authenticated user.

web/profiles/joinup/src/Plugin/Block/AddContentBlock.php
```diff
@@ -0,0 +1,39 @@
...
+class AddContentBlock extends BlockBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function build() {
+    $build = [
+      'collection' => [
+        '#type' => 'link',
+        '#title' => $this->t('Add collection'),
```
idimopoulos: Maybe change this to Propose collection since the title of the page is "Propose collection".

web/profiles/joinup/src/Plugin/Block/AddContentBlock.php
```diff
@@ -0,0 +1,39 @@
...
+        '#url' => Url::fromRoute('collection.propose_form'),
+        '#attributes' => ['class' => ['button', 'button--small']],
```
idimopoulos: what is the reason of the 'button--small' class? Maybe we shouldn't pass classes yet because we don't have a theme. I can understand that the specs speak of a button but can we depend on future class names yet?

web/profiles/joinup/joinup.info.yml
```diff
@@ -57,17 +58,21 @@ config_devel:
...
   - core.date_format.html_month
   - core.date_format.html_time
   - core.date_format.html_week
-  - core.date_format.html_yearless_date
   - core.date_format.html_year
+  - core.date_format.html_yearless_date
   - core.date_format.long
   - core.date_format.medium
   - core.date_format.short
+  - core.entity_form_display.rdf_entity.collection.propose
```
idimopoulos: Can we move this to collection module install config files? Isn't it better to keep it with the rest of the collection config files?

web/profiles/joinup/src/Plugin/Block/AddContentBlock.php
pfrenssen: Yeah it's fine, these are two of the core classes from the "Classy" theme. We can expect these to be supported in whatever theme is going to be developed for us. The standard practice is to base new themes on the "Classy" theme. Core uses the same practice, I ripped this code from `BlockListBuilder::buildBlocksForm()`.

web/profiles/joinup/config/install/user.role.authenticated.yml
pfrenssen: Thanks! Very important. I've fixed it and added tests to ensure that nobody can access the form to create a collection any more.

web/profiles/joinup/src/Plugin/Block/AddContentBlock.php
pfrenssen: Good idea. Fixed it and updated tests accordingly.

web/profiles/joinup/joinup.info.yml
pfrenssen: This is here because the "Propose" mechanism is not exclusive to collections, in the future we will also be able to propose interoperability solutions etc.

To facilitate this, I've created a new form display mode on the RDF entity that is called "propose" (ref. `joinup_entity_type_alter()`). In the future we will add the form displays for interoperability solutions alongside this.

PR #23, opened 2016-03-01
Isaicp 2249

idimopoulos:


tests/features/collection_page/api.feature
```diff
@@ -0,0 +1,23 @@
+@api
+Feature: Collection Page API
+  In order to manage collection pages programmatically
+  As a backend developer
+  I need to be able to use the Node API to handle the "Collection Page" bundle
```
pfrenssen: Node API? :) What's that? :)

Just change it to something like "I need to be able to create, edit and delete collection pages through the API"

tests/features/collection_page/api.feature
```diff
@@ -0,0 +1,23 @@
...
+      | create elibrary | facilitators                |
+      | schedule        | daily                       |
+      | metadata url    | http://joinup.eu/my/foo     |
+      | uri             | http://joinup.eu/my/foo     |
+    And content of type "Collection Page":
+      | title             | body                                     | groups audience         |
+      | Dummy Page        | This is some dummy content like foo:bar. | http://joinup.eu/my/foo |
+     # @Fixme unimplemented.
+     # | Exclude          |                                              |
+     Then I should have a collection page titled "Dummy Page"
```
pfrenssen: Missing newline at the end.

tests/features/collection_page/api.feature
```diff
@@ -0,0 +1,23 @@
...
+      | closed          | 0                           |
+      | create elibrary | facilitators                |
+      | schedule        | daily                       |
+      | metadata url    | http://joinup.eu/my/foo     |
+      | uri             | http://joinup.eu/my/foo     |
+    And content of type "Collection Page":
+      | title             | body                                     | groups audience         |
+      | Dummy Page        | This is some dummy content like foo:bar. | http://joinup.eu/my/foo |
+     # @Fixme unimplemented.
+     # | Exclude          |                                              |
```
pfrenssen: What is this Exclude thing? If this still needs to be implemented then please create a followup issue and link it here. It's a good idea to link every todo to a ticket number.

web/modules/custom/collection_page/collection_page.behat.inc
```diff
@@ -0,0 +1,82 @@
...
+   */
+  protected $bundle = 'collection_page';
+
+  /**
+   * Creates content of a given type provided in the form.
+   * @codingStandardsIgnoreStart
+   * | title    | author     | status | created           |
+   * | My title | Joe Editor | 1      | 2014-10-17 8:00am |
+   * | ...      | ...        | ...    | ...               |
+   * @codingStandardsIgnoreEnd
```
pfrenssen: This doesn't seem to violate coding standards, those indicators are probably not needed.

web/modules/custom/collection_page/collection_page.behat.inc
```diff
@@ -0,0 +1,82 @@
...
+   * @Given content of type "Collection Page":
+   */
+  public function createNodes(TableNode $nodes) {
+    foreach ($nodes->getHash() as $hash) {
+      $node = (object) $hash;
+      $node->type = $this->bundle;
+      $this->nodeCreate($node);
+      $this->collectionPages[] = $node->nid;
+    }
+  }
```
pfrenssen: We don't need this,.it is already covered by the Behat Drupal Extension, use the step `Given :type content` instead.

web/modules/custom/collection_page/collection_page.behat.inc
```diff
@@ -0,0 +1,82 @@
...
+   */
+  public function assertCollectionPage($title) {
+    $actual = \Drupal::entityQuery('node')
+      ->condition('title', $title)
+      ->count()
+      ->execute();
+    if (!$actual) {
+      throw new \Exception("Entity titled '$title' was not found.");
+    }
+  }
```
pfrenssen: This is very useful! Can you rework it so the node type is a parameter? Then we can reuse it for our other node types.

`Then I should have a :type node titled :title`

Put it in our `\Drupal\joinup\Context\DrupalContext`.

web/modules/custom/collection_page/collection_page.behat.inc
```diff
@@ -0,0 +1,82 @@
...
+   * Remove any created collections.
+   *
+   * @AfterScenario
+   */
+  public function cleanCollections() {
+    // Remove any collection_pages that were created.
+    $storage_handler = \Drupal::entityManager()->getStorage('node');
+    $entities = $storage_handler->loadMultiple($this->collectionPages);
+    $storage_handler->delete($entities);
+  }
```
pfrenssen: This is probably also not needed any more if we use the standard step from BDE.

web/modules/custom/collection_page/collection_page.info.yml
```diff
@@ -0,0 +1,12 @@
...
+description: Custom content type
+core: 8.x
+package: Joinup
+dependencies:
+  - og
+  - node
+  - field
+  - path
+  - text
+  - menu_ui
```
pfrenssen: Missing newline at the end.

web/modules/custom/collection_page/collection_page.info.yml
```diff
@@ -0,0 +1,12 @@
+name: collection_page
+type: module
+description: Custom content type
```
pfrenssen: Provide a more descriptive description :)

web/modules/custom/collection_page/config/install/core.entity_view_display.node.collection_page.teaser.yml
```diff
@@ -0,0 +1,27 @@
+langcode: en
```
pfrenssen: We don't have a use for this "teaser" display mode. This is enabled because we copied it from the standard profile. Let's remove it now, otherwise we will carry this unused display mode along for all eternity :)

You can remove it by deleting the `core.entity_view_mode.node.teaser` config file from the install profile.

web/modules/custom/collection_page/config/install/field.field.node.collection_page.og_group_ref.yml
```diff
@@ -0,0 +1,29 @@
+langcode: en
```
pfrenssen: Nice to see that this is an OG content type. Is this actually working? Do you think it's worthwhile to test this at this point? Or is there perhaps a future ticket that plans a full UAT test for this?

web/modules/custom/collection_page/config/install/node.type.collection_page.yml
```diff
@@ -0,0 +1,28 @@
...
+  - node.type.collection_page
+  - field.storage.user.og_group_ref
+  - field.storage.node.body
+  - field.field.node.collection_page.body
+  - field.field.node.collection_page.og_group_ref
+  - core.base_field_override.node.collection_page.promote
+  - core.base_field_override.node.collection_page.status
+  - core.entity_form_display.node.collection_page.default
+  - core.entity_view_display.node.collection_page.default
+  - core.entity_view_display.node.collection_page.teaser
```
pfrenssen: This is interesting, I haven't seen these exports of config_devel data in individual config items before. I wonder what this is used for.

web/profiles/joinup/config/install/block.block.bartik_local_actions.yml
```diff
@@ -3,8 +3,6 @@ status: true
 dependencies:
   theme:
     - bartik
-_core:
-  default_config_hash: 13GQpeITIJsp1kyPniXtWZfyFH87vb1xxJCHifL4UeE
```
pfrenssen: Good cleanup :+1:

web/profiles/joinup/config/install/core.entity_view_mode.node.rss.yml
```diff
@@ -0,0 +1,9 @@
+langcode: en
```
pfrenssen: RSS? I don't think I've seen this in the functional specs yet. It's also here because we copied the standard profile. Let's remove this display mode for now, we can still add it back later if needed in a dedicated ticket.

web/profiles/joinup/config/install/core.entity_view_mode.node.search_index.yml
```diff
@@ -0,0 +1,9 @@
+langcode: en
```
pfrenssen: Remove this display mode too, we don't have search yet.

web/profiles/joinup/config/install/field.storage.user.og_group_ref.yml
```diff
@@ -2,8 +2,8 @@ langcode: en
 status: true
 dependencies:
   module:
-    - collection
     - og
+    - rdf_entity
```
pfrenssen: Good catch!

web/profiles/joinup/config/install/filter.format.basic_html.yml
```diff
@@ -6,8 +6,6 @@ dependencies:
 name: 'Basic HTML'
 format: basic_html
 weight: 0
-roles:
-  - authenticated
```
pfrenssen: Authenticated users are not allowed to use the Basic HTML filter format any more? This is unrelated to this ticket.

I'm guessing this has been exported like this, so this indicates a problem in our configuration. Still I would leave this unchanged. If this pops up again when re-exporting config then we can take a closer look.

web/profiles/joinup/config/install/filter.format.full_html.yml
```diff
@@ -6,8 +6,6 @@ dependencies:
 name: 'Full HTML'
 format: full_html
 weight: 1
-roles:
-  - administrator
```
pfrenssen: Same here, we shouldn't be touching the filter formats in scope of this ticket. Let's revert this.

web/profiles/joinup/config/install/filter.format.restricted_html.yml
```diff
@@ -4,8 +4,6 @@ dependencies: {  }
 name: 'Restricted HTML'
 format: restricted_html
 weight: 0
-roles:
-  - anonymous
```
pfrenssen: Same here.

web/profiles/joinup/config/install/user.role.user.yml
```diff
@@ -0,0 +1 @@
+{  }
```
pfrenssen: Hmm ok that's an improvement I guess :D

web/profiles/joinup/config/install/views.view.collections.yml
```diff
@@ -0,0 +1,452 @@
+langcode: en
```
pfrenssen: I'm confused about this. Did this appear in the export? This is a view showing collections. I think this belongs to a different ticket, probably ISAICP-2205. It doesn't belong in the Joinup install profile either, this should be in the Collections module.

We have to be careful. If we delete it here then it's possible the view will disappear when this is merged with ISAICP-2205.

Can you investigate what happened?

web/modules/custom/collection_page/collection_page.behat.inc
idimopoulos: I know this is a duplicate but I implemented it this way only because the machine name of the entity bundle "Collection Page" is collection_page, so the "Given collection_page content:" would not be so user friendly.
I moved this to a generic behat file in the joinup profile so that we can store there functions that are useful and reusable.

web/modules/custom/collection_page/collection_page.behat.inc
idimopoulos: Moved it to the new file included in joinup profile along with the other function.

web/modules/custom/collection_page/collection_page.behat.inc
idimopoulos: Removed but left the file collection_page.behat.inc for future use.

web/modules/custom/collection_page/config/install/field.field.node.collection_page.og_group_ref.yml
idimopoulos: This is as it is originally exported by Drupal and was created by the OG UI.
It works for the time being as a configuration. We cannot actually test it in the scope of this ticket as the only thing this ticket provides is the definition. 
We will test it though during the "Create custom page" ticket.

web/profiles/joinup/config/install/filter.format.basic_html.yml
idimopoulos: Restored.

web/profiles/joinup/config/install/views.view.collections.yml
idimopoulos: Hmm if I am not mistaking, this was my unintentional doing. I wanted to just rename the file from views.view.collection to views.view.collections due to our conversation on whether the title should be collection or collections. But the 2205 had already passed I think (or I didn't push this change). But now there are two files (one collection and one collections) and one must be deleted because the one overrides the other.

web/modules/custom/collection_page/config/install/field.field.node.collection_page.og_group_ref.yml
pfrenssen: :+1:

web/modules/custom/collection_page/collection_page.behat.inc
pfrenssen: I see you managed to make it work after all. Great job!

tests/src/Context/DrupalContext.php
```diff
@@ -27,6 +28,53 @@ public function loggedIn() {
   }
 
   /**
+   * Checks the existence of a content page.
+   *
+   * @param int $title
+   *   The title of the page.
```
pfrenssen: This is a string, not an integer. The documentation for the `$type` parameter is also missing.

tests/src/Context/DrupalContext.php
```diff
@@ -27,6 +28,53 @@ public function loggedIn() {
...
+   *
+   * @throws \Exception
+   *   Thrown when the bundle type is not found.
+   */
+  public function getBundleFromLabel($bundle_label) {
+    foreach (NodeType::loadMultiple() as $type) {
+      if ($bundle_label == $type->label()) {
+        return $type->id();
+      }
+    }
```
pfrenssen: Retrieving all node types and looping over them to find the right one is not very efficient. I understand the philosophy though, we will only have very few node types in this project so this won't take much time at all. Still it would be better to simply to a query that retrieves the node type by label.

A second point is that this is hard coded on node types. We will need to do the same for RDF types, and in the future even more types. If we would make this generic now, then we don't need to duplicate code and we can save time in the future.

We can even take it further. Since an entity type in itself is _also an entity_, we can make this more generic: let it match on _any entity_ that matches a certain label and entity type. Since the node title is also a label this code can then also be used to look up a node with a certain title. We can hit three birds with one stone!

That would make the code that does this in `assertContentPageByTitle()` above redundant, after all it first retrieves the node type by looking up the entity that matches the "type" label, and then the node by looking up the entity that matches the "title" label. That would reduce the code above to the following:

```
$bundle = $this->getEntityFromLabel('node_type', $type);
$node = $this->getEntityFromLabel('node', $label, $bundle);
```

PR #26, opened 2016-03-08
Isaicp 2312

sandervd:


web/profiles/joinup/config/install/user.role.user.yml
```diff
@@ -0,0 +1 @@
+{  }
```
idimopoulos: Why do we need this file at all?

tests/features/user/register.feature
```diff
@@ -0,0 +1,40 @@
+@api
+Feature: User registration
+  As a user I must be able to register to the site and complete my user profile.
+
+  Scenario: User can find the register page
+    Given I am an anonymous user
+    When I am on the homepage
+    And I click "Log in"
+    Then I am at "/user/login"
+
```
idimopoulos: No need for extra line.

tests/features/user/register.feature
```diff
@@ -0,0 +1,40 @@
...
+    When I click "Create new account"
+    Then I am at "/user/register"
+
+
```
idimopoulos: No need for extra line.

tests/features/user/register.feature
```diff
@@ -0,0 +1,40 @@
...
+      | Family name        | User                  |
+    Then I should see the success message "A welcome message with further instructions has been sent to your email address."
+
+  Scenario: A moderator can register a user
+    Given users:
+      | name            | roles          |
+      | Mr. Moderator   | Moderator     |
+    And I am logged in as "Mr. Moderator"
+    When I am on the homepage
+    Then I click "People"
```
idimopoulos: There should be a 'Then I click "Manage"' first because when the login occurs for the moderator, the People link is hidden by JS. The profile links are there at first. The test passes because the link exists in the page but it is hidden but this will fail in UAT probably.

tests/features/user/register.feature
```diff
@@ -0,0 +1,40 @@
...
+    Then I click "Add user"
+    Then I am at "admin/people/create"
+
+    Given the following user registration at "admin/people/create":
+      | Email address      | superuser@example.org |
```
idimopoulos: Nothing important but can we have the same column width for all elements?

tests/src/Context/DrupalContext.php
```diff
@@ -75,4 +75,23 @@ public function assertFieldsNotPresent($fields) {
...
+  /**
+   * Checks the users existence.
+   *
+   * @param string $username
+   *   The username of the user.
+   *
+   * @throws \Exception
+   *   Thrown when the user is not found.
+   *
+   * @Then I should have a :username user
```
idimopoulos: I think it is better to be "I should have an :username user" because the 'a' refers to the word 'user' so it is better as 'an' than 'a' (don't curse me too much.)

web/profiles/joinup/joinup.behat.inc
```diff
@@ -0,0 +1,174 @@
...
+    $storage = \Drupal::entityManager()->getStorage('user');
+    $query = $storage->getQuery();
+    $result = $query->sort('uid', 'DESC')->range(0, 1)->execute();
+    $uid = array_pop($result);
+    /** @var \Drupal\user\UserInterface $user */
+    $user = $storage->load($uid);
+
+    // Check if latest registered user is created during this request.
+    $user_creation_time = $user->getCreatedTime();
+    $request_time = \Drupal::requestStack()->getCurrentRequest()->server->get('REQUEST_TIME');
```
idimopoulos: This can go to the beginning of the method. The way I see it, this has a dependency that our code will go faster than the db transaction, right?
If there is a delay on the server it might at some point fail (correct me if I'm wrong). 
If we set it to the beginning of the method, then it will always work.

web/profiles/joinup/joinup.info.yml
```diff
@@ -70,11 +70,31 @@ config_devel:
   - core.date_format.long
   - core.date_format.medium
   - core.date_format.short
+  - core.entity_form_display.user.user.default
+  - core.entity_form_display.user.user.register
+  - core.entity_form_mode.user.register
+  - core.entity_view_display.user.user.default
+  - core.entity_view_mode.user.compact
```
idimopoulos: I think we can remove this. We do not need it for now.

PR #28, opened 2016-03-15
Isaicp 2237

idimopoulos:


composer.json
```diff
@@ -120,7 +120,7 @@
         {
             "_comment": "Amitai's fork is the main development repository for Organic Groups. Remove this once a D8 release is made on drupal.org.",
             "type": "vcs",
-            "url": "git@github.com:amitaibu/og"
+            "url": "git@github.com:ec-europa/og"
```
pfrenssen: * The comment here is no longer correct. Mention exactly which pull requests are merged in this branch.
* This branch includes https://github.com/amitaibu/og/pull/140 which at the moment only contains a test. The fixes that were present in the PR in the past have all been merged back into the main branch, so we actually don't need this any more.
* Commit https://github.com/amitaibu/og/commit/44b2702f16bfe1d9ef34bdf0c43e5e742b494014 indicates that this issue is currently blocked on ISAICP-2311. That unfortunately means that this cannot go through QA before ISAICP-2311 is finished.
* This branch contains https://github.com/amitaibu/og/pull/166 which is not complete - it does not have a test.

drush/README.md
```diff
@@ -1 +0,0 @@
-This directory contains commands, configuration and site aliases for Drush. See http://packages.drush.org/ for a directory of Drush commands installable via Composer.
```
pfrenssen: Unrelated change. Revert this. This file is part of the upstream drupal-project and we should not mess with it.

tests/features/custom_page/collection.custom_page.feature
```diff
@@ -0,0 +1,44 @@
...
+  As a collection member
+  I need to be able to add "Custom Page" content through UI.
+  Background:
+    Given users:
+    | name            | mail                  | roles         | pass |
+    | Rufus Drumknott | member@example.com    |               | test |
+    And the following collections:
+    | uri                             | name           | description               | logo     |  owner |
+    | http://joinup.eu/collection/foo | Foo Collection | This is a foo collection. | logo.png |        |
+    | http://joinup.eu/collection/bar | Bar Collection | This is a bar connection. | logo.png |        |
```
pfrenssen: Owner is empty? Is this a required field? If it is not then this column can be removed. If it is then we have a bug :)

tests/features/custom_page/collection.custom_page.feature
```diff
@@ -0,0 +1,44 @@
+@api
+Feature: "Add custom page" visibility options.
+  In order to manage custom pages
+  As a collection member
+  I need to be able to add "Custom Page" content through UI.
+  Background:
+    Given users:
+    | name            | mail                  | roles         | pass |
+    | Rufus Drumknott | member@example.com    |               | test |
```
pfrenssen: Is it normal to have the roles empty? Also if you don't mind it would be nice to not reuse test data from other tests (like the username "Rufus Drumknott"). It's better to test with unique data so we have more chances of catching problems.

tests/features/custom_page/collection.custom_page.feature
```diff
@@ -0,0 +1,44 @@
...
+    And I go to the homepage of the "Foo Collection" collection
+    Then I should see the link "Add custom page"
+    When I click "Leave this collection"
+    And I press the "Confirm" button
+    Then I should not see the link "Add custom page"
+    When I go to the homepage of the "Bar Collection" collection
+    Then I should not see the link "Add custom page"
+    But I should see the button "Join this collection"
+    When I press the "Join this collection" button
+    Then I should see the link "Add custom page"
```
pfrenssen: Why is this tested twice? Did you perhaps have a problem with the button being cached?

tests/features/custom_page/collection.custom_page.feature
```diff
@@ -0,0 +1,44 @@
...
+
+  Scenario: Add custom page as a collection member.
+    When I am logged in as "Rufus Drumknott"
+    And I go to the homepage of the "Foo Collection" collection
+    Then I should see the link "Add custom page"
+    When I click "Add custom page"
+    Then I should see the text "Add custom page"
```
pfrenssen: This test is a bit meaningless. This will simply check if the text "Add custom page" is visible anywhere on the page. This might be a link, a menu item, even a string somewhere in the body text. I suppose you are checking here to see that the page title is "Add custom page" - for that case you can use "Then I should see the heading :heading".

tests/src/Context/DrupalContext.php
```diff
@@ -156,4 +159,92 @@ public function assertUserExistence($username) {
...
+      throw new \Exception("The entity " . $member->getEntityTypeId()
+        . " is not a group content.");
+    }
+
+    if ($member->og_group_ref->getValue()[0]['target_id'] != $group_id) {
+      throw new \Exception("Entity \"" . $member->label()
+        . "\" is not a member of entity with id \"" . $group_id . "\"");
+    }
+  }
+
```
pfrenssen: This stuff is great, we should contribute it to OG.

web/modules/custom/collection/src/Form/JoinCollectionForm.php
```diff
@@ -128,6 +129,10 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
       ->setState(OgMembershipInterface::STATE_ACTIVE)
       ->save();
 
+    // @todo: when og gets its own caching system, use its context instead.
+    // Invalidate cache so that the add content block can be rebuilt.
+    Cache::invalidateTags(['block_view']);
```
pfrenssen: If you have a @todo there should always be a link to a ticket. Otherwise this is not tracked and then it will never be solved.

This workaround is wrong for several reasons:
1. This is a Form class, it is not concerned with where it is being displayed. You should not have any code here relating to blocks. Having this here makes the form no longer reusable.
2. The 'block_view' tag sounds very generic to me. Are you clearing *all* the blocks with this? Don't do that! :)
3. This needs more research. I don't see why OG would not have the necessary cache contexts, after all it is using generic entities.

web/modules/custom/collection/src/Form/LeaveCollectionConfirmForm.php
```diff
@@ -113,6 +114,10 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
     $memberships = Og::membershipStorage()->loadMultiple($membership_ids);
     Og::membershipStorage()->delete($memberships);
 
+    // @todo: when og gets its own caching system, use its context instead.
+    // Invalidate cache so that the add content block can be rebuilt.
+    Cache::invalidateTags(['block_view']);
```
pfrenssen: For the same reason as listed above this workaround should be removed. It does not belong here.

web/modules/custom/custom_page/config/install/core.entity_form_display.node.custom_page.collection_custom_page.yml
```diff
@@ -0,0 +1,52 @@
...
+    weight: 2
+    settings:
+      display_label: false
+    third_party_settings: {  }
+  sticky:
+    type: boolean_checkbox
+    settings:
+      display_label: true
+    weight: 3
+    third_party_settings: {  }
```
pfrenssen: Are all these options necessary? Is it mentioned in the functional specification that there should be fields visible for the paths, promote to frontpage and sticky content? I don't remember this being mentioned in the discussions. If this is offered to the end user then we should also make it work. At the moment we don't have a frontpage or stickied items yet so it's better to hide them now. If we need these functionalities we can enable it again in separate tickets.

web/modules/custom/custom_page/config/install/core.entity_form_display.node.custom_page.default.yml
```diff
@@ -15,20 +16,28 @@ mode: default
...
       placeholder: ''
     third_party_settings: {  }
+  og_group_ref:
+    type: og_complex
+    weight: 0
+    settings:
+      match_operator: CONTAINS
+      size: 60
+      placeholder: ''
+    third_party_settings: {  }
```
pfrenssen: I would definitely hide the OG Complex field. Nobody should have access to this, we manage the group content through the "Add custom page" button instead.

web/modules/custom/custom_page/src/Controller/CustomPageController.php
```diff
@@ -0,0 +1,71 @@
...
+use Drupal\Core\Controller\ControllerBase;
+use Drupal\og\Og;
+use Drupal\user\Entity\User;
+
+// @todo: Fix the description.
+/**
+ * Class CustomPageController.
+ *
+ * @package Drupal\custom_page\Controller
+ */
```
pfrenssen: Indeed :) This description needs to be updated :)

web/modules/custom/custom_page/src/Controller/CustomPageController.php
```diff
@@ -0,0 +1,71 @@
...
+   * Controller for the base form .
+   *
+   * @param \Drupal\rdf_entity\RdfInterface $rdf_entity
+   *   The collection rdf_entity.
+   *
+   * @return array
+   *   Return the form array to be rendered.
+   */
+  public function add($rdf_entity) {
+    // @todo: Find why value is not filtered.
```
pfrenssen: Every @todo should have a link to a ticket where this is going to be solved.

web/modules/custom/custom_page/src/Controller/CustomPageController.php
```diff
@@ -0,0 +1,71 @@
...
+    $node = $this->entityTypeManager()->getStorage('node')->create(array(
+      'type' => 'custom_page',
+      'og_group_ref' => $rdf_entity->Id()
+    ));
+
+    // @todo: Change form name to include '_form' suffix.
+    $form = $this->entityFormBuilder()->getForm($node, 'collection_custom_page');
```
pfrenssen: This form ID still needs to be changed.

web/modules/custom/custom_page/src/Controller/CustomPageController.php
```diff
@@ -0,0 +1,71 @@
...
+
+    return $form;
+  }
+
+  /**
+   * Handles accessibility to the custom page add form through collection pages.
```
pfrenssen: "Handles access" :) Accessibility is more related to allowing people with disabilities to use the website.

web/modules/custom/custom_page/src/Controller/CustomPageController.php
```diff
@@ -0,0 +1,71 @@
...
+      return AccessResult::forbidden();
+    }
+
+    // @todo: Fix the visibility to include og membership role dependency after ISAICP-2362.
+    $user = User::load($account->id());
+    if(!(Og::isMember($rdf_entity, $user))){
+      return AccessResult::forbidden();
+    }
+
+    return AccessResult::allowed();
```
pfrenssen: Shouldn't this be `AccessResult::neutral()` as per the discussion in https://github.com/amitaibu/og/pull/166?

web/modules/custom/custom_page/src/Form/CollectionCustomPageForm.php
```diff
@@ -0,0 +1,42 @@
...
+
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\node\NodeForm;
+
+/**
+ * Class CollectionCustomPageForm.
+ *
+ * @package Drupal\custom_page\Form
+ */
+class CollectionCustomPageForm extends NodeForm {
```
pfrenssen: It seems like this form is useless. It doesn't do anything. It can be removed.

web/modules/custom/rdf_entity/src/Entity/Query/Sparql/Query.php
```diff
@@ -154,7 +162,12 @@ public function condition($property, $value = NULL, $operator = NULL, $langcode
       case 'label-=':
         preg_match('/\((.*?)\)/', $value, $matches);
         $matching = array_pop($matches);
-        $ids = "(<$matching>)";
+        if($matching){
+          $ids = "(<$matching>)";
+        }
+        else{
+          $ids = '("' . $value . '")';
+        }
```
pfrenssen: Coding standards are not correct here. Did the bot accept this? Then we should look into why this file is not checked for coding standards.

web/modules/custom/rdf_entity/src/Entity/Rdf.php
```diff
@@ -314,7 +314,7 @@ public function setName($name) {
   protected function urlRouteParameters($rel) {
     $uri_route_parameters = [];
 
-    if ($rel != 'collection') {
+    if ($rel != 'rdf_entity') {
```
pfrenssen: Haha!! Nice one :)

web/profiles/joinup/joinup.behat.inc
```diff
@@ -105,28 +105,27 @@ class JoinupSubContext extends DrupalSubContextBase implements DrupalSubContextI
...
     // Submit form.
     $this->getSession()->getPage()->pressButton("Create new account");
 
     // Get the last registered user.
-    $storage = \Drupal::entityManager()->getStorage('user');
-    $query = $storage->getQuery();
-    $result = $query->sort('uid', 'DESC')->range(0, 1)->execute();
-    $uid = array_pop($result);
+    $results = \Drupal::entityManager()->getStorage('user')
+        ->loadByProperties(['name' => $username]);
```
pfrenssen: What if the username is empty? This logic doesn't seem valid to me. Isn't the username a required field? I would throw an exception if the username is missing.

web/profiles/joinup/src/Plugin/Block/AddContentBlock.php
```diff
@@ -18,12 +22,55 @@
  *  admin_label = @Translation("Add content"),
  * )
  */
-class AddContentBlock extends BlockBase {
+class AddContentBlock extends BlockBase
+{
```
pfrenssen: This is not correct according to the coding standards. We definitely need to look at the automated coding standards check. Something seems to be wrong there.

web/profiles/joinup/src/Plugin/Block/AddContentBlock.php
```diff
@@ -18,12 +22,55 @@
...
+   *   A configuration array containing information about the plugin instance.
+   * @param string $plugin_id
+   *   The plugin ID for the plugin instance.
+   * @param string $plugin_definition
+   *   The plugin implementation definition.
+   */
+  public function __construct(array $configuration, $plugin_id, $plugin_definition)
+  {
+    parent::__construct($configuration, $plugin_id, $plugin_definition);
+    $this->currentRouteMatch = \Drupal::routeMatch();
```
pfrenssen: Don't access the global `\Drupal` class in OO code. Use dependency injection instead.

web/profiles/joinup/src/Plugin/Block/AddContentBlock.php
```diff
@@ -18,12 +22,55 @@
...
+
+    // @todo: This should be restricted to collection rdf_entities only.
+    // Retrieve the collection from the route.
+    $this->collection = $this->currentRouteMatch->getParameter('rdf_entity');
```
pfrenssen: This has a todo, seems easy enough to fix.

drush/README.md
idimopoulos: I don't even know why this is part of this ticket.. I didn't do it! I cannot even find the thing.

tests/features/custom_page/collection.custom_page.feature
idimopoulos: Name changed. roles column removed. It is normal because the method that creates users does not allow the authenticated role to be passed. So if we want a user with only authenticated role assigned to him, we are passing an empty value. But it can work without it so I'm removing the column.

tests/features/custom_page/collection.custom_page.feature
idimopoulos: Owner is an autocomplete field in the collection form but does not appear in the propose collection page. It is not mandatory (at least for now). Column removed.

tests/features/custom_page/collection.custom_page.feature
idimopoulos: The reason of this test is to prove that the button is cached but varies per user 'and' per collection. So the meaning is that when a state changes for one collection (Foo), then if you visit the other collection (Bar), they do not share the same cached block. It is actually a test that the caching occurs correctly.

tests/features/custom_page/collection.custom_page.feature
idimopoulos: Yeah good poing. I guess even if the title is not "Add cutom page" then the button's name is and it will still pass.

tests/features/custom_page/collection.custom_page.feature
pfrenssen: Ok that makes sense, thanks for the explanation.

tests/features/custom_page/collection.custom_page.feature
pfrenssen: Great, that's what I thought. Maybe you can provide a little comment explaining that this is the reason.

composer.json
idimopoulos: Comment changed. One of the patches included, solve temporarily the issue that gets it blocked by ISAICP-2311 which is the delete functionality. Can't we use this for now?
After all, the patch takes care of behat's incapability of cleaning the memberships because of the lack of functionality currently being built on 2311.
Anyway, bottom line, in order to unblock, I altered the method that cleans the collections after the behat scenarios to also delete the memberships at CollectionSubContext:cleanCollections();.
Is this acceptable until 2311 is in place?

web/modules/custom/custom_page/src/Controller/CustomPageController.php
idimopoulos: I am not 100% sure but I think that allowed is the way to go here. The discussion on the 166 was about a field level access overriding the neutral access returned by all other handlers. 
Also, node is returning a neutral access in order to be better integrated with the rest of the modules.
But in our case, we are overriding the default controller and form in order to alter it a bit and we provide our custom method to handle access through the route. In addition to that, we are not calling for any invocations nor are we part of any core flow since we overrode the access method. That makes this method the "high bidder". 
So, considering that if all elements return a neutral access, then the access is denied, the high bidder is the one who should take the responsibility to allow the access.

Also, changing this to a neutral denies access to the page (because there is no method specifically allowing access and since no method explicitly cares for the access, the system does not care that much and denies it as it should) which backs up all the above a bit. :)

web/modules/custom/custom_page/src/Controller/CustomPageController.php
idimopoulos: Indeed! fixed.

web/modules/custom/custom_page/src/Form/CollectionCustomPageForm.php
idimopoulos: Yup. Not needed any more. Removed

web/profiles/joinup/joinup.behat.inc
idimopoulos: Indeed it is required. Removed the first line.

web/profiles/joinup/src/Plugin/Block/AddContentBlock.php
idimopoulos: Todo removed. In the end, there will be buttons associated with other rdf_entities as well (e.g. solutions). No need to filter per rdf_entity bundle here.

pfrenssen: That makes sense, thanks!

web/profiles/joinup/joinup.behat.inc
pfrenssen: Reading the code this still won't work as expected when the username is not provided.

idimopoulos: Well, I don't think we need to worry about this. This code runs after a registration page is submitted. If the username is not passed at this time, this would be a bug of core I guess (or some other implementation that allowed it). Should we still catch an exception for that possibility?

pfrenssen: This code runs whenever someone uses this step:

```
@Given the following user registration at :location:
```

If you don't supply the username in the table then this method will not work as expected. That's fully the responsibility of this method.

I also don't really see the need for this step at all. Behat has everything built in to register a user through the UI. You just navigate to the page, fill in the fields, press the button and check that the success message is displayed. This is not good from a UAT perspective.

composer.json
```diff
@@ -10,7 +10,7 @@
         "cweagans/composer-patches": "~1.0",
         "drupal/console": "~0.10",
         "drupal/core": "~8.0.3",
-        "drupal/og": "dev-port-og-complex-widget-test",
+        "drupal/og": "dev-userGroupsFix",
```
pfrenssen: My previous remarks here were not fixed. This includes https://github.com/amitaibu/og/pull/166 which is not complete, it still needs a test.

drush/README.md
pfrenssen: This is also not fixed yet.

tests/src/Context/DrupalContext.php
```diff
@@ -156,4 +152,92 @@ public function assertUserExistence($username) {
...
+   * @throws \Exception
+   *   Thrown when an entity is not found or when an entity set as group is not
+   *   a group.
+   *
+   * @Given (the following )user memberships:
+   */
+  public function givenUserMembership(TableNode $membership_table) {
+    foreach ($membership_table->getColumnsHash() as $membership_row) {
+      foreach ($membership_row as $key => $value) {
+        $values[$key] = $value;
```
pfrenssen: The `$values` array is not initialized.

tests/src/Context/DrupalContext.php
```diff
@@ -156,4 +152,92 @@ public function assertUserExistence($username) {
...
+      }
+
+      // Load group.
+      $group = \Drupal::entityTypeManager()
+        ->getStorage($values['group_type'])
+        ->load($values['group_id']);
+      if (empty($group)) {
+        throw new \Exception("Entity " . $values['group_id'] . " not found.");
+      }
```
pfrenssen: This loads every group one by one in a loop. It would be more efficient to load all at once using `loadMultiple()`.

tests/src/Context/DrupalContext.php
```diff
@@ -156,4 +152,92 @@ public function assertUserExistence($username) {
...
+      /** @var \Drupal\og\OgMembership $membership */
+      $membership = Og::membershipStorage()->create(Og::membershipDefault());
+      $membership
+        ->setEntityType($group->getEntityTypeId())
+        ->setEntityid($group->id())
+        ->setState(OgMembershipInterface::STATE_ACTIVE);
+      if ($member->getEntityTypeId() == 'user') {
+        $membership->setUser($member->id());
+      }
+      $membership->save();
```
pfrenssen: This only works for users. This has recently changed. All entity types other than users now use field storage.

tests/src/Context/DrupalContext.php
```diff
@@ -156,4 +152,92 @@ public function assertUserExistence($username) {
...
+    if (empty($member)) {
+      throw new \Exception("The entity titled $member_title was not found.");
+    }
+    if (!(Og::isGroupContent($member->getEntityTypeId(), $member->bundle()))) {
+      throw new \Exception("The entity " . $member->getEntityTypeId() . " is not a group content.");
+    }
+
+    if ($member->og_group_ref->getValue()[0]['target_id'] != $group_id) {
+      throw new \Exception("Entity \"" . $member->label() . "\" is not a member of entity with id \"" . $group_id . "\"");
+    }
```
pfrenssen: This assumes that the field name of the groups audience field is always called `og_group_ref` but this is not always the case, in can have any name. You also need to special case users, since they do not use field storage.

Anyway it's not necessary to do all this stuff, you can use `Og::isMember()` to do the check. But from a quick look it seems this is not yet updated to discern between users and other entities either :(

web/modules/custom/custom_page/custom_page.module
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * @var \Drupal\og\Entity\OgMembership $og_membership
+ *  The membership that has just been inserted.
+ */
+function custom_page_og_membership_insert(OgMembership $og_membership) {
+  $account = \Drupal::currentUser();
+  $current_user = User::load($account->id());
+  $tag = 'rdf_entity:' . $current_user->getEntityTypeId() . ':' . $current_user->id();
+
+  Cache::invalidateTags([$tag]);
```
pfrenssen: This is not reliable. You cannot expect that the currently logged in user is the one whose membership is being inserted. Administrators will also be able to let users join or leave their groups, and we can do this by direct API calls (e.g. through drush commands, cron jobs, import tasks, ...) in which case no user will be logged in.

In all these cases the wrong cache will be flushed, and the caches that are actually affected will not be flushed.

It also seems that this is flushing way too much, it flushes the entire cache tag of the user. We only need to flush the caches of the elements that are actually depending on the membership.

web/modules/custom/custom_page/src/Controller/CustomPageController.php
```diff
@@ -0,0 +1,71 @@
...
+   * @return \Drupal\Core\Access\AccessResult
+   *   The access result object.
+   */
+  public function access() {
+    $rdf_entity = \Drupal::routeMatch()->getParameter('rdf_entity');
+    $account = \Drupal::currentUser();
+
+    if ($account->isAnonymous()) {
+      return AccessResult::forbidden();
+    }
```
pfrenssen: This is strange, anonymous users are not allowed to access the form, but everyone else is. It's better to use a permission for this.

web/profiles/joinup/src/Plugin/Block/AddContentBlock.php
pfrenssen: Just saw that the @todo is still there. If the parameter is going to contain any RDF entity, then change the name from `$this->collection` to `$this->rdfEntity`.

web/profiles/joinup/src/Plugin/Block/AddContentBlock.php
```diff
@@ -33,7 +80,66 @@ public function build() {
       ],
     ];
 
+    // This check has to occur here so that the link can be cached correctly
+    // for each page.
+    if (
+      !($this->account->isAnonymous())
```
pfrenssen: The link access is handled by `CustomPageController::access()` so you don't need to check it here.

web/profiles/joinup/src/Plugin/Block/AddContentBlock.php
```diff
@@ -33,7 +80,66 @@ public function build() {
...
+      && $this->currentRouteMatch->getRouteName() == 'entity.rdf_entity.canonical'
+      && $this->collection->bundle() == 'collection'
+    ) {
+      $build['custom_page'] = [
+        '#type' => 'link',
+        '#title' => $this->t('Add custom page'),
+        '#url' => Url::fromRoute('custom_page.collection_custom_page.add',
+          ['rdf_entity' => $this->collection->sanitizedId()]),
+        '#attributes' => ['class' => ['button', 'button--small']],
+        '#access' => Og::isMember($this->collection, $this->account),
```
pfrenssen: Same for this, this should be checked by `CustomPageController::access()`.

tests/src/Context/DrupalContext.php
idimopoulos: Yes but this method does that exactly. Checks a user membership.
Also it is defined as such: @Given (the following )user memberships:

web/modules/custom/custom_page/custom_page.module
idimopoulos: This is actually a really nice (in terms of debugging) notice! :) I will fix this after the QA to use the uid of the membership.

web/profiles/joinup/src/Plugin/Block/AddContentBlock.php
```diff
@@ -21,6 +20,54 @@
...
+   * @param array $configuration
+   *   A configuration array containing information about the plugin instance.
+   * @param string $plugin_id
+   *   The plugin ID for the plugin instance.
+   * @param string $plugin_definition
+   *   The plugin implementation definition.
+   */
+  public function __construct(array $configuration, $plugin_id, $plugin_definition) {
+    parent::__construct($configuration, $plugin_id, $plugin_definition);
+    $this->currentRouteMatch = \Drupal::routeMatch();
```
pfrenssen: Don't inject the routematcher in a block and do direct checks on the route. That creates a hard coupling between the block and the route. If we ever change the route then this doesn't work any more.

Drupal has the concept of Contexts for this. We already have a context to retrieve the collection from the route: `\Drupal\collection\ContextProvider\CollectionRouteContext`. Let's use this instead.

web/profiles/joinup/joinup.behat.inc
pfrenssen: This is actually out of scope for this ticket. This step is not used in the test that is added for this ticket.

PR #30, opened 2016-03-17
Isaicp 2253

idimopoulos:


web/modules/custom/joinup_document/config/install/field.field.node.joinup_document.body.yml
```diff
@@ -0,0 +1,21 @@
+langcode: en
```
sandervd: I prefer to prefix all fields with the bundle name...

web/modules/custom/joinup_document/joinup_document.module
```diff
@@ -0,0 +1,57 @@
...
+ * If the short title field is filled, it has to be unique in a global scope.
+ *
+ * @param array $form
+ *   The form array.
+ * @param \Drupal\core\Form\FormStateInterface $form_state
+ *   The form state object.
+ */
+function joinup_document_form_default_validate(&$form, FormStateInterface $form_state) {
+  // Check the file submission.
+  if (empty($form_state->getValue('field_file')[0]['fids'])
```
sandervd: Should be a !XOR instead of AND

web/profiles/joinup/joinup.info.yml
```diff
@@ -17,6 +17,7 @@ dependencies:
   - file
   - help
   - image
+  - joinup_document
```
sandervd: I'd prefer collection_document as a name for the module, as it better represents the relation to the group.

idimopoulos: Yes but aren't these content types able to attach to other entities (e.g. solutions)?

web/modules/custom/joinup_document/config/install/field.field.node.joinup_document.body.yml
idimopoulos: The reason behind this is that in Drupal 8 now, you can have a same storage for a field that is used across more than one entity type or bundle. The seperate field.field.{entity_type}.{bundle}.{field_name}.yml gives you the ability to alter some settings on this field but without the need to create a different storage.
Now, I do not mind changing them but my thought was that we can at least re use the same fields no matter who creates them so that we can stay consistent. Like when we create a storage for URLs, we wont need a different one for e.g. 6 content types that use this field type.
Because in the end, the entity query knows which data to draw.
If you remember, we somehow got confused when there were the fields for rdf entity and then different for the asset repository (because we did not discuss it) and it got a bit confusing. Now we can keep the name simple and descriptive and reuse a field across all website with different settings.

So, to conclude, I don't mind altering it, it is just search and replace. But I would like your thoughts in this.
P.S., this kind of idea about the storage, derives from a presentation I saw on a drupal con a talking about how and why they changed this in this way.

PR #34, opened 2016-04-01
Isaicp 2399

idimopoulos:


web/modules/custom/solution/solution.module
```diff
@@ -5,20 +5,16 @@
...
-      $output .= '<h3>' . t('About') . '</h3>';
-      $output .= '<p>' . t('The interoperability solution RDF entity.') . '</p>';
-      return $output;
-
-    default:
+function solution_entity_base_field_info_alter(&$fields, EntityTypeInterface $entity_type) {
+  if ($entity_type->id() == 'node') {
+    // Title must be unique.
+    if (!empty($fields['title'])) {
+      $fields['title']->addConstraint('UniqueFieldInBundle', ['bundles' => ['solution']]);
```
idimopoulos: I agree we should go with custom constrains instead of checking separately in the database.

PR #35, opened 2016-04-06
Isaicp 2401

idimopoulos:


tests/features/asset_distribution/add_solution.feature
```diff
@@ -0,0 +1,53 @@
...
+    Given the following collection:
+      | name | Asset Distribution Test                  |
+      | uri  | https://a.distribution/solution/add/test |
+      | logo | logo.png                                 |
+    And the following solution:
+      | name              | Asset random name                        |
+      | uri               | http://joinup.eu/ad/solution/api/bar     |
+      | description       | Some reusable random description         |
+      | elibrary creation | 1                                        |
+      | groups audience   | https://a.distribution/solution/add/test |
```
pfrenssen: This step does not match the behaviour of the functionality:

* The URI cannot be entered through the UI, so this should not be proposed as an option.
* "1" as an option for elibrary creation doesn't mean anything to an end user. This should be a human readable string such as "Only members can publish new content".
* "Groups audience" doesn't make sense from an end user perspective. This is a term from organic groups. This should be "collection" and it should take the human readable title of the collection, not the URI.

tests/features/asset_distribution/add_solution.feature
```diff
@@ -0,0 +1,53 @@
...
+Feature: "Add asset distribution" visibility options.
+  In order to manage distributions
+  As a moderator
+  I need to be able to add "Asset distribution" rdf entities through UI.
+
+  Scenario: "Add distribution" button should only be shown to moderators.
+    Given the following collection:
+      | name | Asset Distribution Test                  |
+      | uri  | https://a.distribution/solution/add/test |
+      | logo | logo.png                                 |
```
pfrenssen: As a user, when creating a collection I cannot determine the URI, so this should not be presented as an option.

tests/features/asset_distribution/add_solution.feature
```diff
@@ -0,0 +1,53 @@
...
+    And the following solution:
+      | name              | Asset random name                        |
+      | uri               | http://joinup.eu/ad/solution/api/bar     |
+      | description       | Some reusable random description         |
+      | elibrary creation | 1                                        |
+      | groups audience   | https://a.distribution/solution/add/test |
+
+    When I am logged in as a "moderator"
+    And I go to the homepage of the "Asset random name" solution
```
pfrenssen: How do I go to the homepage of the "Asset random name" solution? As a UAT tester this is not clear. It would be better to say where to click in the interface.

pfrenssen: I just tried it manually and it is impossible to reach the solution starting from the homepage.

1. There is no link to the collection overview on the homepage. This is fixed in ISAICP-2457 which has been accepted in develop recently.
2. After I merged in the latest develop and saw the link to the collection overview. However when I clicked it I got an error message: "Notice: Undefined index: server in Drupal\search_api_db\Plugin\search_api\backend\Database->getIndexDbInfo() (line 2460 of modules/contrib/search_api/search_api_db/src/Plugin/search_api/backend/Database.php)." I will try again after a full rebuild and reinstall.
2. The list of collections is empty. This will be fixed in ISAICP-2458.
3. When I finally make it to the collection page by typing in the URL manually, there are no links to the available solutions. There is even no evidence at all that this collection has solutions associated with it.
4. When I go to the solution page by typing in the URL manually, I don't see any evidence that there are distributions associated with the solution.

So in short, there are 4 critical issues blocking

tests/features/asset_distribution/api.feature
```diff
@@ -0,0 +1,58 @@
...
+  I need to be able to use the Solution API
+
+  Scenario: Programmatically create a solution
+    Given the following collection:
+      | name              | Asset distribution collection API foo                  |
+      | logo              | logo.png                                               |
+      | moderation        | 1                                                      |
+      | closed            | 1                                                      |
+      | elibrary creation | facilitators                                           |
+      | uri               | http://joinup.eu/collection/asset_distribution-api-foo |
```
pfrenssen: The "moderation" and "closed" values make no sense. The "URI" cannot be entered. It is not possible to pass UAT based on these instructions.

web/modules/custom/asset_distribution/asset_distribution.behat.inc
```diff
@@ -0,0 +1,354 @@
...
+   *   The solution data.
+   *
+   * @codingStandardsIgnoreStart
+   *   Table format:
+   *   Table format:
+   * uri                                      | name              | description                  | access url                               | file     | creation date     | modification date
+   * http://joinup.eu/asset_distribution/foo  | Foo distribution  | This is a foo distribution   | http://base_url.com/path-to-file.zip     | test.zip | 28-01-1995 12:05  |
+   * http://joinup.eu/asset_distribution/bar  | Bar distribution  | This is a bar distribution   | http://external_url.com/path-to-file.zip | test.zip | 28-01-1995 12:06  |
+   * @codingStandardsIgnoreEnd
+   * Fields title and description are mandatory.
```
pfrenssen: There is no "Title" field.

web/modules/custom/asset_distribution/asset_distribution.behat.inc
```diff
@@ -0,0 +1,354 @@
...
+      // Copy the file into the public files folder and turn it into a File
+      // entity before linking it to the collection.
+      $uri = 'public://' . $filename;
+      $destination = file_unmanaged_copy($path, $uri);
+      $file = File::create(['uri' => $destination]);
+      $file->save();
+      $values['field_ad_file'] = [$file->id()];
+
+      $this->files[$file->id()] = $file;
+    }
```
pfrenssen: This code is duplicated from the `CollectionSubContext`. Let's create a trait that contains code related to working with files, so that we can reuse it.

web/modules/custom/asset_distribution/asset_distribution.behat.inc
```diff
@@ -0,0 +1,354 @@
...
+      ->condition('?entity', SparqlArg::uri('http://purl.org/dc/terms/title'), SparqlArg::literal($name))
+      ->range(0, 1);
+    $result = $query->execute();
+
+    if (empty($result)) {
+      throw new \InvalidArgumentException("The asset distribution with the name '$name' was not found.");
+    }
+
+    return Rdf::load(reset($result));
+  }
```
pfrenssen: This is duplicated from CollectionSubContext. Let's put it in a trait that reuses code related to RDF entities.

web/modules/custom/asset_distribution/asset_distribution.behat.inc
```diff
@@ -0,0 +1,354 @@
...
+
+    if (empty($result)) {
+      throw new \InvalidArgumentException("The solution with the name '$name' was not found.");
+    }
+
+    // We need to skip the static cache and load the unchanged entity from the
+    // database.
+    return \Drupal::entityTypeManager()
+      ->getStorage('rdf_entity')
+      ->loadUnchanged(reset($result));
```
pfrenssen: This seems identical to `getAssetDistributionByName()` above, except for the exception message and that it returns an uncached version of the entity.

web/modules/custom/asset_distribution/asset_distribution.behat.inc
```diff
@@ -0,0 +1,354 @@
...
+   * @Then I should have :number asset distribution(s)
+   */
+  public function assertAssetDistributionCount($number) {
+    $actual = \Drupal::entityQuery('rdf_entity')
+      ->condition('?entity', 'rdf:type', '<http://www.w3.org/ns/adms#AssetDistribution>')
+      ->count()
+      ->execute();
+    if ($actual != $number) {
+      throw new \Exception("Wrong number of asset distributions. Expected number: $number, actual number: $actual.");
+    }
```
pfrenssen: Duplicate code, put it in a trait that deals with RDF entities.

web/modules/custom/asset_distribution/asset_distribution.behat.inc
```diff
@@ -0,0 +1,354 @@
...
+   *   Thrown when an asset distribution with the given title doesn't exist.
+   *
+   * @Then the :asset_distribution asset distribution is related to the :solution solution
+   */
+  public function assertAssetRelationship($asset_distribution, $solution) {
+    /** @var \Drupal\rdf_entity\RdfInterface $asset_distribution */
+    $asset_distribution = $this->getAssetDistributionByName($asset_distribution);
+
+    /** @var \Drupal\rdf_entity\RdfInterface $solution */
+    $this->clearStaticCaches();
```
pfrenssen: This seems overkill to clear ALL static caches, you can just clear the entity cache of this specific entity.

web/modules/custom/asset_distribution/asset_distribution.module
```diff
@@ -0,0 +1,30 @@
...
+use \Drupal\Core\Entity\EntityInterface;
+use \Drupal\rdf_entity\RdfInterface;
+
+/**
+ * Implements hook_ENTITY_TYPE_presave().
+ *
+ * Handle the reverse reference of the solution-distribution entities.
+ */
+function asset_distribution_rdf_entity_presave(EntityInterface $entity) {
+  if ($entity->bundle() != 'asset_distribution') {
```
pfrenssen: Also check the entity type.

web/modules/custom/asset_distribution/asset_distribution.module
```diff
@@ -0,0 +1,30 @@
...
+    return;
+  }
+  // Handle reverse reference in the interoperability solution.
+  /** @var \Drupal\rdf_entity\RdfInterface $rdf_entity */
+  $rdf_entity = \Drupal::routeMatch()->getParameter('rdf_entity');
```
pfrenssen: The fact that an RDF entity of type "solution" is present on the route while a distribution is saved is not enough proof that this action is required. This hook will fire every time an entity is saved, so it will trigger every time a distribution is saved while the user is on any solution page.

For example we could add a block or contextual link on the solution page that allows a collection facilitator to do an action on a distribution (e.g. publish it, highlight it, mark it as a stable release, whatever), then this code will trigger.

The intention is of course to make this reference when a distribution is created through the UI. It would be better to use a form alter to prefill this value in a hidden field.

web/modules/custom/asset_distribution/src/Controller/AssetDistributionController.php
```diff
@@ -0,0 +1,65 @@
...
+   *   The access result object.
+   */
+  public function createAssetDistributionAccess(RdfInterface $rdf_entity) {
+    // Check that the passed in RDF entity is a solution, and that the user
+    // has the permission to create distribution entities.
+    // @todo Collection owners and facilitators should also have the right to
+    //   create distributions related to the solutions they manage.
+    // @see https://webgate.ec.europa.eu/CITnet/jira/browse/ISAICP-2450
+    if ($rdf_entity->bundle() == 'solution' && $this->currentUser()
+        ->hasPermission('propose asset_distribution rdf entity')
```
pfrenssen: Is this really a proposal? Meaning that this needs to be validated by a moderator before it is made public?

web/modules/custom/rdf_entity/src/Plugin/Field/FieldWidget/UrlFileSynchronizerWidget.php
```diff
@@ -0,0 +1,142 @@
...
+use Drupal\Core\Entity\EntityFieldManagerInterface;
+use Drupal\Core\Field\FieldDefinitionInterface;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+use Drupal\file\Entity\File;
+use Drupal\link\Plugin\Field\FieldWidget\LinkWidget;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Plugin implementation of the 'url_file_synchronizer_widget' widget.
```
pfrenssen: Please explain what the purpose of this widget is. It's not clear to me even after I read the code.

web/modules/custom/rdf_entity/src/Plugin/Field/FieldWidget/UrlFileSynchronizerWidget.php
```diff
@@ -0,0 +1,142 @@
...
+    $summary[] = t('File fields to mirror: :filefields',
+      [':filefields' => implode(', ', $this->getSetting('file_fields'))]);
+
+    return $summary + parent::settingsSummary();
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function massageFormValues(array $values, array $form, FormStateInterface $form_state) {
```
pfrenssen: Where are the tests for this code?

web/profiles/joinup/src/Plugin/Block/AddContentBlock.php
```diff
@@ -73,9 +84,10 @@ public function build() {
     // caching ourselves.
     /** @var \Drupal\Core\Plugin\Context\Context[] $collection_contexts */
     $collection_contexts = $this->collectionContext->getRuntimeContexts(['collection']);
-    if ($collection_contexts['collection']->hasContextValue()) {
+    if ($collection_contexts && $collection_contexts['collection']->hasContextValue()) {
       $page_url = Url::fromRoute('custom_page.collection_custom_page.add', [
-        'rdf_entity' => $collection_contexts['collection']->getContextValue()->id(),
+        'rdf_entity' => $collection_contexts['collection']->getContextValue()
+          ->id(),
```
pfrenssen: This is not according to coding standards. Either the chained methods should be inline, or they should be on separate lines, not a combination of both.

web/profiles/joinup/src/Plugin/Block/AddContentBlock.php
```diff
@@ -86,7 +98,8 @@ public function build() {
       ];
 
       $solution_url = Url::fromRoute('solution.collection_solution.add', [
-        'rdf_entity' => $collection_contexts['collection']->getContextValue()->id(),
+        'rdf_entity' => $collection_contexts['collection']->getContextValue()
+          ->id(),
```
pfrenssen: Same here, either split or keep it on one line. I would prefer on one line, this is too trivial to split it over multiple lines.

web/modules/custom/rdf_entity/src/Plugin/Field/FieldWidget/FileSizeCalculatorWidget.php
```diff
@@ -0,0 +1,169 @@
...
+use Drupal\Core\Field\FieldDefinitionInterface;
+use Drupal\Core\Field\FieldItemListInterface;
+use Drupal\Core\Field\WidgetBase;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+use Drupal\file\Entity\File;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Plugin implementation of the 'file_size_calculator_widget' widget.
```
pfrenssen: * Please explain the purpose of the widget. It's already evident from the annotation that this is a plugin that has the "file_size_calculator_widget" machine name.
* It doesn't seem like this belongs in the RDF Entity module. This is project code.
* Tests are missing.

tests/features/asset_distribution/add_solution.feature
idimopoulos: The functionality of the URI generation is occurring in the widget of the field. It is not a storage level functionality, nor the definition of field. This means that direct assignment of the URI is possible. Should we convert the behat to match the current UI? We can just remove it from the Acceptance criteria.

tests/features/asset_distribution/api.feature
idimopoulos: Well, all in all, this is just to check the API. If the behat is all about user experience, maybe we should remove the API tests as they are not UATable.

web/modules/custom/asset_distribution/asset_distribution.behat.inc
idimopoulos: Name is the title field.

web/modules/custom/asset_distribution/asset_distribution.behat.inc
idimopoulos: Indeed it is. I once tried to create a generic "getRdfByName" method but It would need a rdf entity id variable to be passed in the behat sentence. And that would not be user friendly (you also said that xD ). So I just created a duplicate to keep it simple.

web/modules/custom/rdf_entity/src/Plugin/Field/FieldWidget/UrlFileSynchronizerWidget.php
idimopoulos: The issue here is that the entity has two fields (relative to this widget). A file field and a URI field.
The user will be able to provide files for this distribution either by URL or by directly uploading the file to the website.
Now, the thing is that the URL field is part of the rdf entity and the file field is not (it is just a helper for the user to be able to upload the files). So in order to keep the rdf entity up to date, we need to have the URLs passed to the URL field. This implies that if a user has filled both fields, the URL field must be updated (same applies if a link or file is removed during an edit).
After discussions, we thought that it would be better to do this formally in a widget than handling it in a hook.

I thought that since we are going to contribute this module, someone else might need something like this so I thought that it should remain to the base entity (plus for reusability reasons, so that we can avoid dependencies on the distribution entitiy).

web/profiles/joinup/src/Plugin/Block/AddContentBlock.php
idimopoulos: This is maybe one thing about the codesniffer.
The issue is that during a git push there is a check. But if there are problems in the coding standards and the user fixes them, the phpcs goes green regardless of whether the changes were committed.

web/modules/custom/rdf_entity/src/Plugin/Field/FieldWidget/FileSizeCalculatorWidget.php
idimopoulos: Same as the other widget, except from being there for possible re-usability, the main idea is that this is to calculate an the file size of files in an entity and put this number to a field. But the field that holds the size is an rdf entity field and the file field is not necessarily. I could even move it to the joinup profile but it was created to solve problems for an rdf entity field so I thought to put it there.

web/modules/custom/asset_distribution/asset_distribution.behat.inc
idimopoulos: Put on a trait along with some other stuff.

web/modules/custom/asset_distribution/asset_distribution.behat.inc
idimopoulos: That was actually leftover. Removed it as it is not needed anymore. I replaced the load function with the loadUnchanged so that we can get the latest version on loading of an entity.

web/modules/custom/asset_distribution/asset_distribution.module
idimopoulos: Indeed. This could complicate things a bit. Added two more checks. One is regarding the route name. So this will be executed only when we are in the route of the adding a distribution through a solution page.
The other check is whether the entity is a new entity. This will ensure that this hook is not fired on entity update which would duplicate the value in the reference list.

web/modules/custom/asset_distribution/src/Controller/AssetDistributionController.php
idimopoulos: I initially thought that all entities go through the moderation process. Apparently this is not! So, changing to 'create' permission.

web/profiles/joinup/src/Plugin/Block/AddContentBlock.php
idimopoulos: Fixed.

web/modules/custom/rdf_entity/src/Plugin/Field/FieldWidget/FileSizeCalculatorWidget.php
idimopoulos: Also, we have planned to add the testing later on when we have support for specific tests in CPHP

web/profiles/joinup/src/Plugin/Block/AddContentBlock.php
pfrenssen: Doesn't matter, I will find them anyway. I am much better at finding coding standards violations than PHPCS ;)

web/modules/custom/rdf_entity/src/Plugin/Field/FieldWidget/FileSizeCalculatorWidget.php
pfrenssen: There is already support for these tests right now, using Travis CI. We don't need to wait.

web/modules/custom/solution/solution.routing.yml
```diff
@@ -1,4 +1,4 @@
-# Route to create a custom page that belongs to an rdf_entity.
+# Route to create a custom page that belongs to an rdf_entity.asset_distribution
```
pfrenssen: This does not seem correct.

web/modules/custom/solution/solution.behat.inc
```diff
@@ -352,8 +308,53 @@ class SolutionSubContext extends DrupalSubContextBase implements DrupalSubContex
...
+   * @param array $fields
+   *    An array of fields keyed by field name.
+   *
+   * @return mixed
+   *    The array with the values converted.
+   *
+   * @throws \Exception
+   *    Throws an exception when a mapped value is not found.
+   */
+  private function convertValueAliases($fields) {
```
pfrenssen: Why is this private? Let's make it protected, there is usually no good reason to make anything private.

web/modules/custom/solution/solution.behat.inc
```diff
@@ -226,18 +204,8 @@ class SolutionSubContext extends DrupalSubContextBase implements DrupalSubContex
...
-      ->range(0, 1);
-    $result = $query->execute();
-
-    if (empty($result)) {
-      throw new \InvalidArgumentException("The solution with the name '$name' was not found.");
-    }
-
-    return Rdf::load(reset($result));
+  public static function getSolutionByName($title) {
+    return self::getRdfEntityByLabel($title, 'solution');
```
pfrenssen: There should be no public static methods in this class. The only public methods should be the step definitions. There is no reason why this class should ever be loaded by external code.

web/modules/custom/asset_distribution/asset_distribution.behat.inc
```diff
@@ -0,0 +1,302 @@
...
+    /** @var \Drupal\rdf_entity\RdfInterface $asset_distribution */
+    $asset_distribution = $this->getAssetDistributionByName($asset_distribution);
+
+    /** @var \Drupal\rdf_entity\RdfInterface $solution */
+    $solution = $this->getRdfEntityByLabelUnchanged($solution, 'solution');
+
+    // This handles the solutions added by form submit.
+    if (!in_array($asset_distribution, $this->assetDistributions)) {
+      $this->assetDistributions[$asset_distribution->id()] = $asset_distribution;
+    }
```
pfrenssen: This should not be here. This is an assert (starts with the keyword `@Then`), so this should only check if any visible results are as expected. It should not do any unrelated logic. If a test creates an asset distribution by submitting the form through the interface, they should also remove it again through the interface.

tests/features/authentication.feature
```diff
@@ -18,13 +18,10 @@ Feature: User authentication
...
     And I should not see the text "My account"
 
-  Scenario Outline: Anonymous user can access public pages
+  Scenario: Anonymous user can access public pages
     Given I am not logged in
-    Then I visit "<path>"
-
-    Examples:
-      | path                           |
-      | rdf_entity/http%3A\\drupal.org |
```
pfrenssen: Don't remove this. This is the generic authentication test which tests lists of URLs.

tests/features/authentication.feature
```diff
@@ -18,13 +18,10 @@ Feature: User authentication
...
+    When I go to the homepage of the "Überwaldean Land Eels" collection
+    Then I should see the heading "Überwaldean Land Eels"
```
pfrenssen: Isn't this already tested? Why is this here in the authentication test? This should be in the collection test suite.

tests/features/authentication.feature
```diff
@@ -44,13 +41,10 @@ Feature: User authentication
...
-    Given I am logged in as a user with the "authenticated" role
-    Then I visit "<path>"
-
-    Examples:
-      | path                           |
-      | rdf_entity/http%3A\\drupal.org |
+  Scenario: Authenticated user can access pages they are authorized to
+    Given I am logged in as a user with the "authenticated user" role
+    When I go to the homepage of the "Überwaldean Land Eels" collection
+    Then I should see the heading "Überwaldean Land Eels"
```
pfrenssen: Revert this, this doesn't belong here, this is the generic authentication test for administration pages. If you want to test collections, do this in one of the tests in the `tests/features/collection/` folder.

tests/src/Context/JoinupTrait.php
```diff
@@ -0,0 +1,164 @@
...
+   * @param string $type
+   *   The rdf entity type.
+   *
+   * @return \Drupal\rdf_entity\Entity\Rdf
+   *   The asset distribution.
+   *
+   * @throws \InvalidArgumentException
+   *   Thrown when an asset distribution with the given name does not exist.
+   */
+  protected static function getRdfEntityByLabel($title, $type) {
```
pfrenssen: A protected static method doesn't make much sense.

web/modules/custom/asset_distribution/asset_distribution.module
pfrenssen: No don't rely on the route. This functionality is related to the form that creates the distribution, not to the route that happens to display the form :)

The best way would be to prefill the reference in the form using a hidden value.

tests/features/asset_distribution/add_asset_distribution.feature
```diff
@@ -0,0 +1,49 @@
...
+    And I click "Add distribution"
+    Then I should see the heading "Add Asset distribution"
+    When I fill in "Title" with "Custom title of asset distribution"
+    And I attach the file "test.zip" to "Add a new file"
+    And I press "Save"
+    Then I should have 1 asset distribution
+    When I go to the homepage of the "Asset another random name" solution
+    Then I should see the text "Distribution"
+    And I should see the link "Custom title of asset distribution"
+    And the "Custom title of asset distribution" asset distribution is related to the "Asset another random name" solution
```
pfrenssen: Let's click on it to see if the asset distribution is accessible. It would also be nice to check this as an anonymous user, this will help catch caching problems.

web/modules/custom/asset_distribution/asset_distribution.module
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * Submit handler for rdf_entity_asset_distribution_form.
+ *
+ * When creating a new asset distribution, add the relation to the
+ * interoperability solution. Usually this relation would point the other way,
+ * but this is how the data model has been designed.
+ *
+ * @see asset_distribution_form_rdf_entity_asset_distribution_form_alter()
+ */
+function asset_distribution_form_rdf_entity_asset_distribution_form_submit(&$form, FormStateInterface $form_state) {
```
idimopoulos: Can you explain me why this is preferred in comparison to the presave or insert (better insert) hook?
I'm not saying it is wrong, it is just that my thought was that in order to be clean coding and not interfere with drupal's form submission and other possible errors during submission, it would be better to handle this functionality only after the entity is created.
Is this because the hook will be fired in every occasion and this is functionality occurring only in one specific case?

PR #37, opened 2016-04-07
ISAICP-2454

idimopoulos:
…onstraint. Maybe there is a better solution for this.


web/modules/custom/joinup_event/joinup_event.module
```diff
@@ -34,20 +35,21 @@ function joinup_event_form_node_event_edit_form_alter(&$form, FormStateInterface
...
-  /** @var \Drupal\Core\Entity\EntityInterface $entity */
-  $entity = $form_state->getFormObject()->getEntity();
-
-  $query = \Drupal::entityQuery('node')
-    ->condition('type', 'event')
-    ->condition('field_event_short_title', $form_state->getValue('field_event_short_title')[0]['value']);
-  if ($entity->id()) {
-    $query->condition('nid', $entity->id(), '<>');
-  }
-  $results = $query->execute();
```
pfrenssen: Why is this being removed?

web/modules/custom/collection/collection.module
```diff
@@ -15,3 +16,20 @@ function collection_form_block_form_alter(&$form, FormStateInterface $form_state
...
+  if ($entity_type->id() == 'rdf_entity') {
+    if (!empty($fields['label'])) {
+      $bundles = ['collection'];
+      if ($constraint = $fields['label']->getConstraint('UniqueFieldInBundle')) {
+        $bundles = array_merge($bundles, $constraint['bundles']);
+      }
+
+      $fields['label']->addConstraint('UniqueFieldInBundle', ['bundles' => $bundles]);
+    }
+  }
```
pfrenssen: This seems strange, altering the base fields will make the constraint available on ALL the bundles. It would be better to do this on a per-bundle basis. This is certainly possible since you can even assign validation constraints through the UI (e.g. make a field required).

web/modules/custom/joinup_event/joinup_event.module
pfrenssen: OK so this logic is now handled by the constraint. Nice!

web/modules/custom/joinup_event/joinup_event.module
```diff
@@ -5,49 +5,29 @@
...
-
-  if (!empty(reset($results))) {
-    $form_state->setErrorByName('field_event_short_title', t('There is already an event with this short title.'));
+function joinup_event_entity_base_field_info_alter(&$fields, EntityTypeInterface $entity_type) {
+  if ($entity_type->id() == 'node') {
+    if (!empty($fields['field_event_short_title'])) {
+      $fields['field_event_short_title']->addConstraint('UniqueField');
+    }
```
pfrenssen: Is this still needed now that the constraint is already added to the short title in `joinup_event_entity_bundle_field_info_alter()`?

web/modules/custom/rdf_entity/src/Entity/Rdf.php
```diff
@@ -248,15 +248,8 @@ public static function baseFieldDefinitions(EntityTypeInterface $entity_type) {
...
-    // @todo Only invoked after cache clear.
-    $fields = array();
-    // If ($bundle == 'admssw_softwareproject') {.
-    $fields['test'] = BaseFieldDefinition::create('string')
-      ->setLabel(t('Test bundle admssw_softwareproject'))
-      ->setDescription(t('The ID of the Contact entity.'))
-      ->setReadOnly(TRUE);
-    // }.
+    // Allow other modules to override the label field.
+    $fields['label'] = clone $base_field_definitions['label'];
```
pfrenssen: Is it really necessary to do this in the RDF Entity module? Why would the standalone contributed RDF Entity module suddenly be interested in providing the label as a bundle field? What about the `rid` and the `id`? Are other modules not allowed to override these? If all are bundle fields, why define them as base fields in the first place?

It would be better to look if it is possible to do this with a hook such as `hook_entity_bundle_field_info_alter()`.

web/modules/custom/collection/config/install/core.base_field_override.rdf_entity.collection.label.yml
```diff
@@ -0,0 +1,17 @@
...
+entity_type: rdf_entity
+bundle: collection
+label: 'Title'
+description: ''
+required: true
+translatable: true
+default_value: { }
+default_value_callback: ''
+settings: { }
+field_type: string
```
pfrenssen: Discussed this with @idimopoulos. If this base field is not overridden using this exported config, then the field will not be exposed in `hook_entity_bundle_field_info_alter()` and will not be alterable. So we need this.

PR #49, opened 2016-04-18
Isaicp 2321

sandervd:


idimopoulos:
I have checked your implementation. Everything is fine except from the fact that I believe an email should be sent after an entity is updated. Also, this saves us a couple of extra checks. 
Can your review the changes and merge the ticket?

PR #52, opened 2016-04-21
Isaicp 2240

sandervd:


web/modules/custom/collection/collection.behat.inc
```diff
@@ -58,6 +58,24 @@ class CollectionSubContext extends DrupalSubContextBase implements DrupalSubCont
   }
 
   /**
+   * Assert the order of elements.
+   *
+   * @Then I should see the following in the repeated :element element within the context of the :parentElement element:
```
idimopoulos: This is really unnecessary at this point. We only need this method for one specific case and it is really unfriendly for the user to include css tags in the behat step.
I propose to change this to @Then I should see the following collection menu items in the specified order:

If we are to need this function in the future for other cases, we can make a trait out of it or just move the code somewhere else. In my opinion it is better to hardcode the css tags in the code for now.

tests/features/collection/menu.feature
```diff
@@ -1,29 +1,55 @@
...
     And I fill in "Link" with "/user"
     Then I press the "Save" button
     Then I should see the success message "The menu link has been saved."
 
     # Check for the link on the collection
-    When I go to the homepage of the "MenuCollection1" collection
-    Then I should see the link "My user account page for the collection menu"
+    When I go to the homepage of the "Lord of the rings" collection
+    # Check order of links.
+    Then I should see the following in the repeated ".menu-item" element within the context of the ".block-og-menu ul.menu .menu-item" element:
```
idimopoulos: This statement should be simpler for the user. 
Check comment lower on the function behind it.

tests/features/collection/menu.feature
```diff
@@ -1,29 +1,55 @@
...
+      | text           |
+      | Mines of Moria |
+      | Mordor         |
+
+    # Edit menu and change the order of items.
+    Then I click the contextual link "Edit menu" in the "Primary menu" region
```
idimopoulos: Since the contextual link is hidden behind an icon, we should include a comment here for the user to be able to know how to find it.

tests/src/Context/DrupalContext.php
```diff
@@ -38,6 +39,47 @@ public function assertContentPageByTitle($type, $title) {
   }
 
   /**
+   * Resolves contextual links directly, without the need for javascript.
+   *
+   * @Then I click the contextual link :linkText in the :region region
```
idimopoulos: I would still prefer this to be named something like "Then I click the "Edit Menu" link of the collection menu" and also hardcode the region or map the regions with the menu names. The user does not know the theme region names to begin with.

PR #55, opened 2016-04-26
Isaicp 2238

sandervd:


web/modules/custom/custom_page/custom_page.module
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * @var \Drupal\og\Entity\OgMembership $og_membership
+ *  The membership that has just been inserted.
+ */
+function custom_page_og_membership_insert(OgMembership $og_membership) {
+  $account = \Drupal::currentUser();
+  $current_user = User::load($account->id());
+  $tag = 'rdf_entity:' . $current_user->getEntityTypeId() . ':' . $current_user->id();
+
+  Cache::invalidateTags([$tag]);
```
idimopoulos: If OG invalidates all membership list in every membership insert and this function exists to avoid this issue, how is it prevented here? I only see an invalidation of a specific cache tag. How does this prevent the rest of the tags being invalidated?

PR #56, opened 2016-04-27
Isaicp 2399

idimopoulos:


web/modules/custom/asset_distribution/src/Plugin/Field/FieldWidget/FileSizeCalculatorWidget.php
```diff
@@ -30,8 +30,14 @@ class FileSizeCalculatorWidget extends WidgetBase implements ContainerFactoryPlu
   /**
    * {@inheritdoc}
    */
-  public function __construct($plugin_id, $plugin_definition, FieldDefinitionInterface $field_definition, array $settings, array $third_party_settings,
```
pfrenssen: A constructor can never inherit the documentation from the parent, since it is constructing a different class.

web/modules/custom/asset_distribution/src/Plugin/Field/FieldWidget/FileSizeCalculatorWidget.php
```diff
@@ -30,8 +30,14 @@ class FileSizeCalculatorWidget extends WidgetBase implements ContainerFactoryPlu
...
-                              EntityFieldManagerInterface $entity_field_manager) {
+  public function __construct(
+      $plugin_id,
+      $plugin_definition,
+      FieldDefinitionInterface $field_definition,
+      array $settings,
+      array $third_party_settings,
+      EntityFieldManagerInterface $entity_field_manager
+  ) {
```
pfrenssen: This is weird, why split it out like this? I don't think I have seen this pattern in Drupal before. I would just leave it on one single line.

web/modules/custom/asset_distribution/src/Plugin/Field/FieldWidget/UrlFileSynchronizerWidget.php
```diff
@@ -30,8 +30,14 @@ class UrlFileSynchronizerWidget extends LinkWidget implements ContainerFactoryPl
...
-  public function __construct($plugin_id, $plugin_definition, FieldDefinitionInterface $field_definition, array $settings, array $third_party_settings,
-                              EntityFieldManagerInterface $entity_field_manager) {
+  public function __construct(
+      $plugin_id,
+      $plugin_definition,
+      FieldDefinitionInterface $field_definition,
+      array $settings,
+      array $third_party_settings,
+      EntityFieldManagerInterface $entity_field_manager
+  ) {
```
pfrenssen: Same here. This is not how function definitions are typically done in Drupal.

web/modules/custom/collection/collection.behat.inc
```diff
@@ -283,7 +283,7 @@ class CollectionSubContext extends DrupalSubContextBase implements DrupalSubCont
   }
 
   /**
-   * Deletes a collection.
+   * Deletes an asset distribution.
```
pfrenssen: Please revert this, the documentation was correct before.

tests/src/Context/JoinupTrait.php
```diff
@@ -0,0 +1,168 @@
...
+
+
+use Drupal\Component\Uuid\Php;
+use Drupal\file\Entity\File;
+use Drupal\rdf_entity\Entity\Rdf;
+
+/**
+ * Helper trait for behat tests.
+ */
+trait JoinupTrait {
```
pfrenssen: This entire file should be removed. This is originating from ISAICP-2401 but it has been removed before that ticket was finished. All this code is now split up in separate traits such as RdfEntityTrait etc.

tests/features/solution/api.feature
```diff
@@ -1,3 +1,4 @@
+@api
```
pfrenssen: Why is this added?

web/modules/custom/asset_distribution/src/Plugin/Field/FieldWidget/UrlFileSynchronizerWidget.php
```diff
@@ -30,8 +30,14 @@ class UrlFileSynchronizerWidget extends LinkWidget implements ContainerFactoryPl
   /**
    * {@inheritdoc}
    */
-  public function __construct($plugin_id, $plugin_definition, FieldDefinitionInterface $field_definition, array $settings, array $third_party_settings,
```
pfrenssen: A constructor cannot inherit documentation, this should be written out in full.

web/modules/custom/rdf_entity/src/Entity/Query/Sparql/Query.php
```diff
@@ -200,26 +199,29 @@ public function condition($property, $value = NULL, $operator = '=', $langcode =
         }
         return $this;
 
+      case '_field_exists-EXISTS':
+      case '_field_exists-NOT EXISTS':
+        $field_rdf_name = $this->getFieldRdfPropertyName($value, $field_storage_definitions);
+
+        if (!filter_var($field_rdf_name, FILTER_VALIDATE_URL) === FALSE) {
```
pfrenssen: It's better to use `\Drupal\Component\Utility\UrlHelper::isValid` to validate URLs.

web/modules/custom/rdf_entity/src/Entity/Query/Sparql/Query.php
```diff
@@ -200,26 +199,29 @@ public function condition($property, $value = NULL, $operator = '=', $langcode =
...
-      $field_storage = $field_storage_definitions[$field_name];
-      if (empty($column)) {
-        $column = $field_storage->getMainPropertyName();
-      }
-      $field_rdf_name = $field_storage->getThirdPartySetting('rdf_entity', 'mapping_' . $column, FALSE);
-      if (empty($field_rdf_name)) {
-        throw new \Exception('No 3rd party field settings for ' . $field_name);
-      }
-      if (UrlHelper::isValid($value, TRUE)) {
+      $field_rdf_name = $this->getFieldRdfPropertyName($field_name, $field_storage_definitions);
```
pfrenssen: Great idea to factor this out, makes sense!

web/modules/custom/rdf_entity/src/Entity/Query/Sparql/Query.php
```diff
@@ -200,26 +199,29 @@ public function condition($property, $value = NULL, $operator = '=', $langcode =
...
+
+      if (!filter_var($value, FILTER_VALIDATE_URL) === FALSE) {
```
pfrenssen: Use `\Drupal\Component\Utility\UrlHelper::isValid` to validate URLs.

web/modules/custom/solution/config/install/core.base_field_override.rdf_entity.solution.label.yml
```diff
@@ -14,4 +14,4 @@ translatable: true
 default_value: { }
 default_value_callback: ''
 settings: { }
-field_type: string
+field_type: string
```
pfrenssen: This is introducing a coding standards violation, please revert this change.

web/modules/custom/solution/solution.module
```diff
@@ -2,16 +2,24 @@
 
 /**
  * @file
- * Contains solution.module..
+ * Contains file solution.module.
```
pfrenssen: It would be nice to describe the purpose of this file. The filename is like the least interesting thing that can be said here :)

Typically this is something like:

> Contains hook implementations for the Interoperability Solutions module.

web/modules/custom/solution/solution.module
```diff
@@ -2,16 +2,24 @@
...
  */
 
-use Drupal\Core\Entity\EntityTypeInterface;
+use \Drupal\Core\Entity\EntityTypeInterface;
```
pfrenssen: Please revert this, it was correct before.

web/modules/custom/solution/src/Plugin/Validation/Constraint/UniqueSolutionTitleValidator.php
```diff
@@ -0,0 +1,76 @@
...
+      return;
+    }
+    $field_name = $items->getFieldDefinition()->getName();
+    /** @var \Drupal\rdf_entity\RdfInterface $entity */
+    $entity = $items->getEntity();
+
+    $entity_type_id = $entity->getEntityTypeId();
+    $id_key = $entity->getEntityType()->getKey('id');
+
+    // If the is_version_of field is not empty, check for the same title.
```
pfrenssen: Can you explain why this is important? I can see in the code that it is checked that the `is_version_of` field is empty. Why is this important? Why can't we check the title if there is something there?

web/modules/custom/solution/src/Plugin/Validation/Constraint/UniqueSolutionTitleValidator.php
```diff
@@ -0,0 +1,76 @@
...
+    if (!empty($entity->get('field_is_is_version_of')
+      ->getValue()[0]['target_id'])
+    ) {
+      /** @var \Drupal\rdf_entity\RdfInterface $parent */
+      $parent = Rdf::load($entity->get('field_is_is_version_of')
+        ->getValue()[0]['target_id']);
+
+      if ($parent->label() == $entity->label()) {
+        // The release has the same name as the solution.
+        return;
```
pfrenssen: This is confusing, the validator is supposed to guarantee that "a field is unique for a given entity type" but as soon as it is discovered it is not unique the validation ends?

web/modules/custom/solution/src/Plugin/Validation/Constraint/UniqueSolutionTitleValidator.php
```diff
@@ -0,0 +1,76 @@
...
+      }
+
+      // Check if the name is the same to other releases of the entity.
+      foreach ($parent->get('field_is_has_version')->getValue() as $release) {
+        $simbling = Rdf::load($release['target_id']);
+        if ($entity->label() == $simbling->label()) {
+          return;
```
pfrenssen: Same here, this seems to check the inverse of what is the goal of the validator.

Sidenote: it is spelled 'sibling' instead of 'simbling' ;)

web/modules/custom/solution/src/Plugin/Validation/Constraint/UniqueSolutionTitleValidator.php
```diff
@@ -0,0 +1,76 @@
...
+    $query = \Drupal::entityQuery($entity_type_id)
+      // The id could be NULL, so we cast it to 0 in that case.
+      ->condition($id_key, (int) $items->getEntity()->id(), '<>')
+      ->condition($field_name, $item->value)
+      ->condition('rid', 'solution');
+    // @todo: Discuss about it whether we need it.
+    if (empty($entity->get('field_is_is_version_of')
+      ->getValue()[0]['target_id'])
+    ) {
+      $query->notExists('field_is_is_version_of');
```
pfrenssen: Here is this mysterious `is_version_of` field again. Why is this used?

tests/features/solution/add_solution.feature
```diff
@@ -37,6 +37,8 @@ Feature: "Add solution" visibility options.
       | Description       | This is a test text              |
       | Documentation     | text.pdf                         |
     And I press "Save"
-    Then the "Collection solution test 2" collection has a solution named "Collection solution add solution"
+    # The name of the solution should exist in the block of the relative content in a collection.
+    When I go to the homepage of the "Collection solution test 2" collection
+    Then I should see the link "Collection solution add solution"
```
pfrenssen: What is shown on this page? I guess we don't have the specs / wireframes of this yet. Maybe just click it and assert that the title is showing correctly, then at least we can already catch any fatal errors that occur on that page.

web/modules/custom/solution/config/install/core.base_field_override.rdf_entity.solution.label.yml
idimopoulos: What is the issue here? I cannot see the difference

web/modules/custom/solution/src/Plugin/Validation/Constraint/UniqueSolutionTitleValidator.php
idimopoulos: I guess I was thinking to selfish here to not explain the validator in a comment.
Sorry for that. I have updated the validation file to include a class documentation that explain the general process and reason of the validation and updated the comments in the file itself to better explain the piece of validation that is taking place every time.

Please check the file again.

web/modules/custom/solution/src/Plugin/Validation/Constraint/UniqueSolutionTitleValidator.php
idimopoulos: Some stupid mistakes I'm making. changed to siblings.
Also the same as above. Comments and documentation updated.

web/modules/custom/solution/src/Plugin/Validation/Constraint/UniqueSolutionTitleValidator.php
idimopoulos: Also keep in mind that this validation file was built with the specifications that existed when the ticket was implemented. The functionality might need to change but according to the data I had at that moment, this was the constraint that had to be applied (now that I actually describe it in the comments :/ )

web/modules/custom/solution/src/Plugin/Validation/Constraint/UniqueSolutionTitleValidator.php
idimopoulos: Comments updated. But in short, the answer is that our solution and release entity bundles are actually one bundle. A solution can have many releases and a release belongs to one solution only. The fact that a release belongs to a solution is mandatory. An orphan release cannot exist.
When a release is being created it is attached to a solution with the field is_version_of. Since a solution cannot be a release as well, this field is always empty but if the entity is a release, this field is mandatory. 
So the whole distinction between the two ADMS properties are based on this field.

BTW this is going to change soon but the current branch was built in this philosophy and is going to be altered.

tests/features/solution/add_solution.feature
idimopoulos: Well I think you are right because there are no other tests that actually validate that clicking a link does not break the page of a solution so I am adding en extra two lines of testing here as a proof of this.

web/modules/custom/solution/config/install/core.base_field_override.rdf_entity.solution.label.yml
pfrenssen: There is no newline, this is the most "invisible" of the coding standards :) No worries, I'll fix it.

tests/features/solution/add_solution.feature
pfrenssen: Great, thanks!

web/modules/custom/solution/src/Plugin/Validation/Constraint/UniqueSolutionTitleValidator.php
```diff
@@ -9,6 +9,21 @@
...
+ *
+ * The solution entity is defined by having an empty field is_version_of.
+ * This is enough because releases can only be created through the solution and
+ * automatically have the is_version_of field filled. An entity that has the
+ * field is_version_of filled is automatically a release.
+ *
+ * The following checks make sure that a solution must have a unique title among
+ * solutions and a release must have a unique title against other solutions and
+ * their releases but can have the same name as their parent solution or their
+ * sibling releases.
```
pfrenssen: This is really really good documentation. It clears up all the confusion, not only for me, but for all developers dealing with this in the next 8+ years. Thanks very much!!

web/modules/custom/solution/src/Plugin/Validation/Constraint/UniqueSolutionTitleValidator.php
```diff
@@ -54,9 +67,7 @@ public function validate($items, Constraint $constraint) {
       ->condition($field_name, $item->value)
       ->condition('rid', 'solution');
     // @todo: Discuss about it whether we need it.
-    if (empty($entity->get('field_is_is_version_of')
-      ->getValue()[0]['target_id'])
-    ) {
+    if (empty($entity->get('field_is_is_version_of')->getValue()[0]['target_id'])) {
```
pfrenssen: Thanks for fixing.

web/modules/custom/asset_distribution/src/Plugin/Field/FieldWidget/UrlFileSynchronizerWidget.php
```diff
@@ -28,16 +28,9 @@ class UrlFileSynchronizerWidget extends LinkWidget implements ContainerFactoryPl
...
    */
-  public function __construct(
-      $plugin_id,
-      $plugin_definition,
-      FieldDefinitionInterface $field_definition,
-      array $settings,
-      array $third_party_settings,
-      EntityFieldManagerInterface $entity_field_manager
-  ) {
+  public function __construct($plugin_id, $plugin_definition, FieldDefinitionInterface $field_definition, array $settings, array $third_party_settings, EntityFieldManagerInterface $entity_field_manager) {
```
pfrenssen: We need to declare all the properties here. I'll take care of it.

PR #59, opened 2016-04-28
Isaicp 2473

pfrenssen:


tests/features/collection/menu.feature
```diff
@@ -28,28 +28,3 @@ Feature: Collections menu
     # Assert link is not on other collection pages
     When I go to the homepage of the "MenuCollection2" collection
     Then I should not see the link "My user account page for the collection menu"
-
```
idimopoulos: Yes, this test was merely a proof of concept that the bug exists because the problem was happening only after the first time a menu needed to be deleted.
Now that the bug is fixed, it is good that this test is removed as well. Should have included a comment for this.

PR #64, opened 2016-05-04
ISAICP-2487 update collection entity fields RDF mapping

idimopoulos:


idimopoulos:
I'm merging this ticket. Changes are confirmed and I also changed of the fields in the form.

PR #77, opened 2016-05-20
Isaicp 2518

idimopoulos:


web/profiles/joinup/joinup.profile
```diff
@@ -83,3 +84,24 @@ function joinup_form_field_config_edit_form_alter(&$form) {
...
+  foreach ($values as &$value) {
+    // Since the profile includes a filter format, we provide this as default.
+    switch ($storage->getType()) {
+      case 'text_long':
+        if ($value['format'] == 'full_html') {
+          $value['format'] = 'content_editor';
+        }
+        break;
+    }
+  }
```
pfrenssen: We don't need to loop over the `$values` if the storage type is not equal to `text_long`.

pfrenssen:
This will work for our needs today but this is a temporary workaround for an important missing feature in the RDF Entity module.

The actual problem is that we are not saving the field metadata. Having default values is not enough, since the configuration of the field is changeable per instance: for example one node might have the Full HTML text format for a particular field, while another node might have plain text. This field metadata is currently not being saved meaning we have a critical data loss problem. It's important that we address this. So please create a followup issue in Jira (and later in drupal.org after we publish the module) so we can address this and find a way to save the data.

Another thing that would be nice is to provide API documentation for the hook that was added here, so we can explain to developers how to use this.

PR #79, opened 2016-05-23
Isaicp 2362

sandervd:


web/modules/custom/collection/src/EventSubscriber/CollectionEventSubscriber.php
```diff
@@ -0,0 +1,56 @@
...
+   * {@inheritdoc}
+   */
+  public static function getSubscribedEvents() {
+    return [
+      PermissionEventInterface::EVENT_NAME => [['provideDefaultOgPermissions']],
+    ];
+  }
+
+  /**
+   * Provides default OG permissions.
```
sandervd: I'm a bit sleepy today (meeting overflow), so this might have something to do with it.
I didn't get it straight away that permissions are declared here.
Maybe something like: "Declare the permissions defined for collections." could help.
(Although that sentence doesn't seem grammatically correct. I can't come up with anything else at this point though. Maybe after a proper night of sleep.)

web/modules/custom/custom_page/src/Controller/CustomPageController.php
```diff
@@ -59,7 +56,28 @@ public function createCustomPageAccess(RdfInterface $rdf_entity) {
       return AccessResult::allowed();
     }
 
+    // Check if the user has the OG permission to create a custom page.
+    if (OgAccess::userAccessGroupContentEntityCrud('create', $rdf_entity, $this->createNewCustomPage($rdf_entity))->isAllowed()) {
```
idimopoulos: This is not something that needs fixed as we will not need it in our project, but there is an actual security gap here.
If another rdf_entity is a group and not intended for having custom pages, but the access is accidentally assigned to that user, he will be able to create a custom page there. This is because the `if ($rdf_entity->bundle == 'collection' ` above is part of the global permission check and if it does not guarantee that this page _is_ a collection page.

pfrenssen: The first check is intended for moderators, here we need to check the bundle, because the global permission is not granular per bundle.

The second check does not need an additional check on the bundle, we are passing the `$rdf_entity` to `OgAccess::userAccessGroupContentEntityCrud()` which has per-bundle permissions. It will get the bundle from the entity and will check if the user has this permission within the bundle. So if the bundle is not a collection then it will not return `allowed` (unless the user also happens to have access in that other group type).

web/modules/custom/collection/src/EventSubscriber/CollectionEventSubscriber.php
pfrenssen: OK! Fixed.

PR #80, opened 2016-05-23
Isaicp 2198

sandervd:


web/modules/custom/rdf_entity/rdf_export/src/Controller/RdfExportController.php
```diff
@@ -0,0 +1,114 @@
...
+    $graph = $sparql->query($query);
+    return $graph->serialise($format);
+  }
+
+  /**
+   * Builds a list of supported serialization formats.
+   */
+  protected function getSerializerFormats() {
+    // Many more are supported...
+    // @todo Move this to a settings form.
```
idimopoulos: This would indeed be better off in a settings form but even without a form, a config file installed including your settings would also be nice.

idimopoulos: I replaced the static array with a configuration form. 
Rest of the module works fine! No issues detected. 
Some of the export methods have external dependencies though and would be nice to be able to check them somehow.

PR #84, opened 2016-05-27
Isaicp 2536

idimopoulos:


web/modules/custom/collection/src/ContextProvider/CollectionRouteContext.php
```diff
@@ -48,6 +51,22 @@ public function getRuntimeContexts(array $unqualified_context_ids) {
...
+
+          // Check if the object is a og content entity.
+          if (Og::isGroupContent($entity_type, $bundle) && ($groups = Og::getGroupIds($route_parameter, 'rdf_entity', 'collection'))) {
+            // A content can belong to only one rdf_entity.
+            $collection = Rdf::load(reset($groups['rdf_entity']));
+            $value = $collection;
+          }
+        }
+      }
+    }
```
pfrenssen: This needs to be documented.

web/modules/custom/collection/src/ContextProvider/CollectionRouteContext.php
```diff
@@ -48,6 +51,22 @@ public function getRuntimeContexts(array $unqualified_context_ids) {
         }
       }
     }
+    elseif ($route_parameters = $this->routeMatch->getParameters()) {
+      /** @var ContentEntityInterface $route_parameter */
+      foreach ($route_parameters as $route_parameter) {
```
pfrenssen: This is strange, doesn't this mean that you would loop over all parts that the route consists of? So if there are multiple parameters in the route (e.g. in a route like /entity/123/other_entity/456) then it will check each of them and consider the route to be part of a collection if any of the entities match.

We should only check the actual route object, not the parameters.

web/modules/custom/collection/src/Plugin/Block/CollectionContentBlock.php
```diff
@@ -87,7 +87,9 @@ public static function create(ContainerInterface $container, array $configuratio
    */
   public function build() {
     if (empty($this->collection)) {
-      throw new \Exception('The "Collection content" block can only be shown on collection pages.');
+      // If this is not a collection page, return an empty array so that the
+      // block is not rendered.
+      return [];
```
pfrenssen: This seems like a step backwards, why do we need to relax our error checking? If there is no collection on the route this code should not even be running.

web/modules/custom/collection/src/Plugin/Block/JoinCollectionBlock.php
```diff
@@ -56,12 +57,18 @@ class JoinCollectionBlock extends BlockBase implements ContainerFactoryPluginInt
    * @param \Drupal\Core\Session\AccountProxyInterface $user
    *   The current user.
    */
-  public function __construct(array $configuration, $plugin_id, $plugin_definition, RouteMatchInterface $current_route_match, AccountProxyInterface $user) {
+  public function __construct(array $configuration, $plugin_id, $plugin_definition, RouteMatchInterface $current_route_match, AccountProxyInterface $user, ContextProviderInterface $collection_context) {
     parent::__construct($configuration, $plugin_id, $plugin_definition);
     $this->currentRouteMatch = $current_route_match;
     $this->user = $user;
     // Retrieve the collection from the route.
     $this->collection = $this->currentRouteMatch->getParameter('rdf_entity');
```
pfrenssen: This is wrong, we should get it from the context exclusively.

web/modules/custom/collection/src/Plugin/Block/JoinCollectionBlock.php
```diff
@@ -56,12 +57,18 @@ class JoinCollectionBlock extends BlockBase implements ContainerFactoryPluginInt
...
+
+    if (empty($this->collection)) {
+      if (!empty($collection_context->getRuntimeContexts(['collection'])['collection']->getContextValue())) {
+        $this->collection = $collection_context->getRuntimeContexts(['collection'])['collection']->getContextValue();
+      }
+    }
```
pfrenssen: We should throw an exception if the collection is not found on the context.

web/modules/custom/collection/src/ContextProvider/CollectionRouteContext.php
idimopoulos: I will, and update the ticket.

web/modules/custom/collection/src/ContextProvider/CollectionRouteContext.php
idimopoulos: Well yes but again, there is no "clear" documentation on the pages we want to show it where we could just list the routes. 
Also, currently, we do not have something like <entity_type>/parameter/<something else>/<other parameter> or something similar. 
Is there a better way that you suggest? The way I would see that, if this is any route path of an entity then the context can be available. If later on we create more complicated functionality and it breaks our tests, then we can fix this. (no?)

pfrenssen: Hmm ok this is apparently how the `ParameterBag` works, the upcasted entities on the route are accessed by their parameter name. This is different from Drupal 7, where you could specify the 'position' of the object. If you did a call to `menu_get_object()` in D7 you would only get an object back if it adhered to the route `entity_type/%id`, since it would ignore any entity IDs that were not in position "1". Now it matches on any entity found, whatever its position in the string.

If we are not sure of the position of the object on the route then unexpected things might happen, especially in the backend. For example we might be on some administration page that happens to have a collection content entity in it, for example some imaginary route like 'admin/config/subscriptions/collection/%id' - this would match, making the collection available on the route, and unexpected things will happen (for example the "Join collection" button might be shown).

So if we cannot rely on the RouteMatch giving us back only the "main entity" we need to come up with some other way to limit this to the correct routes. I'm guessing we should provide a whitelist of routes, let's start with `node/{node}`, `rdf_entity/{rdf_entity}` and `rdf_entity/{rdf_entity}/leave`. In the future we will need to add routes for the overviews and search results pages too.

web/modules/custom/collection/src/Plugin/Block/CollectionContentBlock.php
idimopoulos: The reason is the very context we are implementing in this ticket.
If the context is available, then the condition of "shown to collection pages" is becoming true. And while with the context present, you can consider this page as a collection pages, still, you would not want the block with the content of the collection to be shown to a (e.g.) news page that belong to this collection.

Also, IMHO, if you consider the usability of this block, it is meant to show some related content. Misplacing it to a page should not break the page throwing an exception because this change is regarding actually the visibility of the block. returning an empty array simple tells drupal to not render it.

web/modules/custom/collection/src/Plugin/Block/JoinCollectionBlock.php
idimopoulos: I will put it but do we need to? I mean, if the block is shown to pages fulfilling the requirement of "Show on collection pages"  which checks if the page is a collection and the context exists for collections, do we need to make this check?

web/modules/custom/collection/src/Plugin/Block/JoinCollectionBlock.php
pfrenssen: Meaning, the old code is now wrong, it should be replaced by the new code :)

web/modules/custom/collection/src/ContextProvider/CollectionRouteContext.php
idimopoulos: Comment added to the class.

web/modules/custom/collection/src/ContextProvider/CollectionRouteContext.php
idimopoulos: I added a filter for the routes and added basic routes to the array.
Maybe an alter hook can be set there though to allow other modules provide their own routes.
Do you think it is needed?

web/modules/custom/collection/src/Plugin/Block/CollectionContentBlock.php
idimopoulos: The idea behind this was to not render the block if the collection is not retrieved from the route.
I have put back the exception and moved the check to the blockAccess method.

web/modules/custom/collection/src/Plugin/Block/CollectionContentBlock.php
```diff
@@ -87,7 +92,7 @@ public static function create(ContainerInterface $container, array $configuratio
    */
   public function build() {
     if (empty($this->collection)) {
-      throw new \Exception('The "Collection content" block can only be shown on collection pages.');
+      throw new \Exception('The "Join Collection" block can only be shown on collection pages.');
```
pfrenssen: This is mentioning the wrong block now :)

web/modules/custom/collection/src/Plugin/Block/CollectionContentBlock.php
```diff
@@ -125,4 +130,11 @@ public function getCacheMaxAge() {
     return 0;
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  protected function blockAccess(AccountInterface $account) {
+    return ($this->currentRouteMatch->getRouteName() == 'entity.rdf_entity.canonical') ? AccessResult::allowed() : AccessResult::forbidden();
+  }
```
pfrenssen: Why is this needed? Doesn't the visibility plugin work correctly?

web/modules/custom/collection/src/Plugin/Block/JoinCollectionBlock.php
```diff
@@ -56,12 +57,13 @@ class JoinCollectionBlock extends BlockBase implements ContainerFactoryPluginInt
...
-  public function __construct(array $configuration, $plugin_id, $plugin_definition, RouteMatchInterface $current_route_match, AccountProxyInterface $user) {
+  public function __construct(array $configuration, $plugin_id, $plugin_definition, RouteMatchInterface $current_route_match, AccountProxyInterface $user, ContextProviderInterface $collection_context) {
     parent::__construct($configuration, $plugin_id, $plugin_definition);
     $this->currentRouteMatch = $current_route_match;
     $this->user = $user;
-    // Retrieve the collection from the route.
-    $this->collection = $this->currentRouteMatch->getParameter('rdf_entity');
+    if (!empty($collection_context->getRuntimeContexts(['collection'])['collection']->getContextValue())) {
+      $this->collection = $collection_context->getRuntimeContexts(['collection'])['collection']->getContextValue();
+    }
```
pfrenssen: The collection is required for this to work, it cannot be skipped. We are using the context for the block visibilty, so if this doesn't exists then we have a serious problem. If you want to do a check here, then at least throw an exception when the check fails.

tests/features/context/context_availability.feature
```diff
@@ -0,0 +1,36 @@
+@api
+Feature: Context availability
+  In order to check context availability
+  As a backend developer
+  I need to be able to see collection content in child entities pages.
```
pfrenssen: If you write this as a "backend developer" then this cannot be UAT'd. It would be much better if this was written from the perspective of a normal user of the website. This functionality actually intended for end users, so they can perform actions related to collections on content pages. Also "Context availability" is not something that will be very clear to our users.

Better would be:

"In order to be able to perform actions related to the collection to which certain content belongs
As an authenticated user
I should be presented with collection related buttons on collection content pages.

web/modules/custom/collection/src/Plugin/Block/CollectionContentBlock.php
idimopoulos: Yes but now that the context is expanded, I only want this block to be shown for the canonical view of the rdf entity. And since it is a temporary block, I don't want to bother updating the configuration file with paths etc. Directly in the code and block it for now.

web/modules/custom/collection/src/Plugin/Block/CollectionContentBlock.php
idimopoulos: I'm so clumsy :( sorry for that.

web/modules/custom/collection/src/Plugin/Block/JoinCollectionBlock.php
idimopoulos: The context parameter is in the condition of the block. If I throw an exception here, the installation of the profile fails because when it loads the configuration for the first time and tries to save the block, the constructor is fired (I guess) and the exception is thrown! So the installation fails.
Two things to notice here, first, if the `->getContextValue();` from this line has an issue, it will fire an exception (I removed the check) and it does not fire during the installation of the profile.
Second, we anyway include a check with an exception during at the build method of this block class. We are already covered for showing in pages with the collection context.

tests/features/context/context_availability.feature
idimopoulos: Well, I renamed it but I initially intended it as a non UAT'able ticket. This is why it was described like this. The reason is that while the block is available now for the rest of the pages, this was not part of the main scope of this ticket rather than simply like a proof of concept. The goal was to expand the context availability so this block is just one case.

But I can see your point so changed it.

web/modules/custom/collection/src/Plugin/Block/JoinCollectionBlock.php
```diff
@@ -84,12 +80,13 @@ public static function create(ContainerInterface $container, array $configuratio
    * {@inheritdoc}
    */
   public function build() {
-    if (empty($this->collection)) {
```
idimopoulos: If the context is missing, then the `->getContextValue() will be called on null and an exception will be fired. I think we should check if the value is empty since we throw our own exception at the next line based on the existence of the context here.

pfrenssen: OK I'll remove our exception.

PR #91, opened 2016-06-01
Isaicp 2267

idimopoulos:


web/modules/custom/joinup_news/joinup_news.module
```diff
@@ -22,3 +24,29 @@ function joinup_news_entity_bundle_field_info_alter(&$fields, EntityTypeInterfac
...
+ * how it was created. This function will return the parent of the entity.
+ *
+ * @param \Drupal\Core\Entity\EntityInterface $entity
+ *    The news content entity.
+ *
+ * @return \Drupal\rdf_entity\RdfInterface|null
+ *    The parent of the entity. This can be a collection or a solution.
+ *    If there is no parent found, return NULL.
+ */
+function joinup_news_get_parent(EntityInterface $entity) {
```
sandervd: Could we turn this into a protected method of the guard condition?

tests/features/joinup_news/add_news.collection.feature
```diff
@@ -0,0 +1,70 @@
...
+    # Add news belonging to a collection.
+    When I click "Add news"
+    Then I should see the heading "Add news"
+    And the following fields should be present "Headline, Kicker, Content, State"
+    And the following fields should not be present "Groups audience"
+    When I fill in the following:
+      | Headline | New avengers member                  |
+      | Kicker   | Black Widow                          |
+      | Content  | Specialized in close combat training |
+    And I select "Draft" from "State"
```
sandervd: Here I would certainly assert the negative as well:
I can't transition to...
I would do this for every state.

web/modules/custom/joinup_news/src/Guard/FulfillmentGuard.php
```diff
@@ -0,0 +1,219 @@
...
+  /**
+   * {@inheritdoc}
+   *
+   * We need to override default transitions allowed because this is also
+   * dependant on the parent's moderation, system roles and organic groups
+   * user roles. In the following method, the allowed transitions per
+   * moderation are checked and then if the transition is allowed, the user
+   * roles by the system and the organic groups are checked.
+   */
+  public function allowed(WorkflowTransition $transition, WorkflowInterface $workflow, EntityInterface $entity) {
```
sandervd: Check here if the guard applies:
Only 'nodes' of the type 'news' should be handled.

web/modules/custom/joinup_news/src/Guard/FulfillmentGuard.php
```diff
@@ -0,0 +1,219 @@
...
+    $from_state = $entity->field_news_state->first()->value;
+    $parent = joinup_news_get_parent($entity);
+
+    $is_moderated = self::MODERATED;
+    if ($parent) {
+      $is_moderated = ($parent->bundle() == 'collection') ?
```
sandervd: Better to unify the fields in the base field module now that we have it...

web/modules/custom/joinup_news/joinup_news.module
idimopoulos: Sure, moved!

tests/features/joinup_news/add_news.collection.feature
idimopoulos: I updated this to check for the validated option in the moderated collection.
Generally I do not thing we need to test every option because the state_machine module is coming with tests. This specific state is the one we override and remove with the guard.

web/modules/custom/joinup_news/src/Guard/FulfillmentGuard.php
idimopoulos: The guard is bound to a workflow group. It is declared in the services yml.
We are going to provide one workflow for each case we have so we do not need to check if it applies as this is called only for the entities that have field of type state using this workflow group.

The connection between the workflow and the entity is:
A workflow group is defined in the workflow_group file.
A gourp (or more) is defined in the workflow file.
A service for the guard is defined marked with the tag of the workflow group id.
A state field is created and in the settings, one of the groups are selected.

So as long as we do not use this workflow for another bundle or entity type, there is no need to check.

web/modules/custom/joinup_news/src/Guard/FulfillmentGuard.php
idimopoulos: These are two fields from different providers. One is og_group_ref and the other one is an entity_reference field, since collections are groups and solutions are not. I think there will be many problems it that happens.

tests/src/Traits/NodeTrait.php
```diff
@@ -1,7 +1,7 @@
 <?php
 
 namespace Drupal\joinup\Traits;
-use Drupal\node\Entity\Node;
+
```
brummbar: Double blank line, one could be removed.

web/modules/custom/joinup_news/config/install/joinup_news.settings.yml
```diff
@@ -0,0 +1,110 @@
...
+#   ]
+# ]
+# @endcode
+#
+# The array is a four dimensions array. The first level contains the
+# value of the moderation of the parent entity. The second level is the
+# target state which is tha state that the entity transits to. Each target
+# state is an array of allowed source states which in every check, is the
+# current state of the entity. Finally, the source states are arrays
+# of roles that are allowed to perform this action.
```
brummbar: Phrasing can be reworked a bit here, maybe.

> ... the value of the moderation of the parent entity ...

could be

> ... the moderation value of the parent entity ... 

Again here
> The second level is the state which is tha state that the entity transits to

could be 

> The second level is/represents the target state the entity transits to.

This one I think need to be rewritten, as it's not clear.

> Each target is an array of allowed source states which in every check, is the state of the entity.

web/modules/custom/joinup_news/config/install/joinup_news.settings.yml
```diff
@@ -0,0 +1,110 @@
...
+#
+# The reverse indexing here (source states
+# indexed by the target state) is to follow the state_machines module logic
```
brummbar: I wouldn't use the word "indexing", it's more "keyed by".

web/modules/custom/joinup_news/joinup_news.workflows.yml
```diff
@@ -0,0 +1,45 @@
+# In the following workflow we have added one 'virtual' state called new.
+#
+# This state will be used for when the entity is being created. In order for the state_machine module to ensure that
+# there is at least one option for the field, it forces the first available state and checks it's available transitions.
```
brummbar: > This state will be used ~~for~~ when the entity is being created ... and checks ~~it's~~ its available transitions

web/modules/custom/joinup_news/joinup_news.module
```diff
@@ -22,3 +31,123 @@ function joinup_news_entity_bundle_field_info_alter(&$fields, EntityTypeInterfac
...
+
+/**
+ * Implements hook_node_access().
+ *
+ * Interrupts the permissions for updating a node. The permission is not
+ * granted if there are no transitions allowed for the current state of the
+ * node.
+ */
+function joinup_news_node_access(EntityInterface $node, $op, AccountInterface $account) {
+  if ($op == 'update' && $node->bundle() == 'news') {
```
brummbar: You could check earlier if it's not a news bundle and return immediately `AccessResult::neutral()`. Personal preference.

web/modules/custom/joinup_news/joinup_news.module
```diff
@@ -22,3 +31,123 @@ function joinup_news_entity_bundle_field_info_alter(&$fields, EntityTypeInterfac
...
+    $allowed_transitions = $node->get('field_news_state')->first()->getTransitions();
+    if (empty($allowed_transitions)) {
+      return AccessResult::forbidden();
+    }
+  }
+
+  // Check if the user is a member and has the 'view any unpublished content'.
```
brummbar: Nitpicking, but:
> has the 'view any unpublished content' **permission**.

web/modules/custom/joinup_news/joinup_news.module
```diff
@@ -22,3 +31,123 @@ function joinup_news_entity_bundle_field_info_alter(&$fields, EntityTypeInterfac
...
+  elseif (in_array($state, ['validated', 'in_assessment', 'deletion_request'])) {
+    $entity->setPublished(TRUE);
+  }
+}
+
+/**
+ * Implements hook_field_widget_WIDGET_TYPE_form_alter().
+ *
+ * The state_machine forces the current value to be available in the options.
+ * However, there are cases that the user cannot keep the node in the same state
```
brummbar: > ... cases ~~that~~ **where** the user ...

web/modules/custom/joinup_news/joinup_news.module
brummbar: Was it moved?

web/modules/custom/joinup_news/config/install/joinup_news.settings.yml
```diff
@@ -0,0 +1,110 @@
...
+        - moderator
+        - rdf_entity-collection-facilitator
+    in_assessment:
+      validated:
+        - moderator
+        - rdf_entity-collection-member
+    deletion_request:
+      validated:
+        - moderator
+        - rdf_entity-collection-administrator
```
brummbar: Coding standards do not check for presence of empty line at the end of the yaml file, but it's better to have it (like core modules do).

web/modules/custom/joinup_news/config/schema/joinup_news.schema.yml
```diff
@@ -0,0 +1,22 @@
...
+          label: 'Target state'
+          sequence:
+            type: sequence
+            label: 'Source state'
+            sequence:
+              type: sequence
+              label: 'Roles'
+              sequence:
+                type: string
+                label: 'Role'
```
brummbar: No empty line at the end of file.

web/modules/custom/joinup_news/joinup_news.services.yml
```diff
@@ -0,0 +1,5 @@
+services:
+  joinup_news.fulfillment_guard:
+    class: Drupal\joinup_news\Guard\FulfillmentGuard
+    tags:
+      - { name: state_machine.guard, group: news }
```
brummbar: No empty line at the end of file.

web/modules/custom/joinup_news/joinup_news.workflow_groups.yml
```diff
@@ -0,0 +1,3 @@
+news:
+  label: News
+  entity_type: node
```
brummbar: No empty line at the end of file.

web/modules/custom/joinup_news/joinup_news.workflows.yml
```diff
@@ -0,0 +1,45 @@
...
+      from: [__new__, draft, validated, proposed, deletion_request]
+      to: validated
+    report:
+      label: Report
+      from: [validated]
+      to: in_assessment
+    request_deletion:
+      label: Request deletion
+      from: [validated]
+      to: deletion_request
```
brummbar: No empty line at the end of file.

web/modules/custom/joinup_news/src/Guard/FulfillmentGuard.php
```diff
@@ -0,0 +1,117 @@
...
+    // Check for the transitions allowed.
+    // @see: joinup_news.settings.yml
+    if (!isset($allowed_conditions[$is_moderated][$to_state][$from_state])) {
+      return FALSE;
+    }
+
+    // This Guard class's method called whenever the transitions are checked
+    // even outside the entity CRUD forms. Cases like this is e.g. when trying
+    // to edit the settings of the field.
+    // In these cases, there is no parent entity so we need to check for it.
```
brummbar: I would rephrase this to something like:

>
This Guard class method is called whenever the transitions are checked
even outside the entity CRUD forms, for example trying to edit the 
settings of the field.
In these cases, there is no parent entity available.

web/modules/custom/joinup_news/src/Guard/FulfillmentGuard.php
```diff
@@ -0,0 +1,117 @@
...
+    if ($parent) {
```
brummbar: I think it would be more readable to do a `if (!$parent) { return FALSE; }` as there would be no further processing anyway.

web/profiles/joinup/joinup.behat.inc
```diff
@@ -10,11 +10,14 @@ use Drupal\DrupalExtension\Context\DrupalSubContextInterface;
 use \Behat\Gherkin\Node\TableNode;
 use \Drupal\file\Entity\File;
 use Drupal\search_api\Entity\Server;
+use \Drupal\joinup\Traits\NodeTrait;
 
 /**
  * Behat step definitions for testing users.
  */
 class JoinupSubContext extends DrupalSubContextBase implements DrupalSubContextInterface {
+  use NodeTrait;
```
brummbar: Add en empty line above the use statement.

web/profiles/joinup/joinup.behat.inc
```diff
@@ -171,6 +174,53 @@ class JoinupSubContext extends DrupalSubContextBase implements DrupalSubContextI
   }
 
   /**
+   * Checks that the given select field has the given options.
+   *
+   * To do this we extract the text of the field which is the values of the
+   * select box concatenated. Then we check if the value is in this string.
```
brummbar: Rephrase it a bit. Something like:
> We extract the field text which contains the values expected to be found in the select box.
Then we check if all the options are available.

web/profiles/joinup/joinup.behat.inc
```diff
@@ -171,6 +174,53 @@ class JoinupSubContext extends DrupalSubContextBase implements DrupalSubContextI
...
+   *    The name of the field element.
+   * @param string $options
+   *    The options to be checked separated by comma. Only labels are accepted.
+   *
+   * @throws Exception
+   *    Throws an exception when a value does not exists.
+   *
+   * @Given (the ):select field has the :options option(s)
+   */
+  public function checkSelectOption($select, $options) {
```
brummbar: Since the parameter itself is `$options`, the name should reflect that: `checkSelectOptions`.

web/profiles/joinup/joinup.behat.inc
```diff
@@ -171,6 +174,53 @@ class JoinupSubContext extends DrupalSubContextBase implements DrupalSubContextI
...
+    $values = $this->getSession()->getPage()->findField($select)->getText();
+    foreach (explode(',', $options) as $option) {
+      if (strpos($values, trim($option)) === FALSE) {
+        throw new \Exception("The option {$option} should exist in the {$select} field but was not found.");
+      }
+    }
+  }
+
+  /**
+   * Checks that the given select field does not have the given values.
```
brummbar: > Does not contain any of the given values.

web/profiles/joinup/joinup.behat.inc
```diff
@@ -171,6 +174,53 @@ class JoinupSubContext extends DrupalSubContextBase implements DrupalSubContextI
...
+   *    The name of the field element.
+   * @param string $options
+   *    The options to be checked separated by comma.
+   *
+   * @throws Exception
+   *    Throws an exception when a value exists.
+   *
+   * @Given (the ):select field does not have the :options option(s)
+   */
+  public function checkSelectOptionInverse($select, $options) {
```
brummbar: Same as before, should be plural Options.

web/profiles/joinup/joinup.behat.inc
```diff
@@ -194,4 +244,63 @@ class JoinupSubContext extends DrupalSubContextBase implements DrupalSubContextI
...
+   * @throws \Exception
+   *   Throws an exception if the content is not published.
+   *
+   * @Then the :title :bundle (content )should not be published
+   */
+  public function nodeNotPublished($bundle, $title) {
+    /** @var \Drupal\node\NodeInterface $node */
+    $node = $this->getNodeByTitle($title, $bundle);
+    if ($node->isPublished()) {
+      throw new \Exception("$title content should be published but is not.");
```
brummbar: Wrong exception message here, should be "shouldn't be published but it is".

web/profiles/joinup/joinup.behat.inc
```diff
@@ -194,4 +244,63 @@ class JoinupSubContext extends DrupalSubContextBase implements DrupalSubContextI
...
+   * @throws \Exception
+   *   Throws an exception if the content is not published.
+   *
+   * @Then the :title :bundle (content )should be published
+   */
+  public function nodePublished($bundle, $title) {
+    /** @var \Drupal\node\NodeInterface $node */
+    $node = $this->getNodeByTitle($title, $bundle);
+    if (!$node->isPublished()) {
+      throw new \Exception("$title content should be published but is not.");
```
brummbar: "but it is not / it isn't" maybe?

web/modules/custom/joinup_news/joinup_news.module
idimopoulos: It was but then I needed it also for a hook node access and moved it back. When og is updated and the workaround is not needed any more, then this can move back into a protected method of the class.

web/modules/custom/joinup_news/config/install/joinup_news.settings.yml
```diff
@@ -16,14 +16,14 @@
 # @endcode
 #
 # The array is a four dimensions array. The first level contains the
-# value of the moderation of the parent entity. The second level is the
-# target state which is tha state that the entity transits to. Each target
+# moderation calue of the parent entity. The second level is the
```
brummbar: Typo here, "value" instead of "calue".

web/modules/custom/joinup_news/joinup_news.workflows.yml
```diff
@@ -0,0 +1,49 @@
+# The workflow of the news entity.
+#
+# In the following workflow we have added one 'virtual' state called new.
+# This state will be used when the entity is being created.
+# In order for the state_machine module to ensure that
+# there is at least one option for the field, it forces
+# the first available state and checks it's available transitions.
```
brummbar: Typo, should be "its".

web/profiles/joinup/joinup.behat.inc
idimopoulos: I think it is actually correct because 'but' connects the two sentences and the subject remains (content) the same.

idimopoulos: But anyway, I will change it :)

PR #92, opened 2016-06-02
Isaicp 2500

idimopoulos:


web/modules/custom/solution/src/Plugin/Block/SolutionContentBlock.php
```diff
@@ -51,12 +55,12 @@ class SolutionContentBlock extends BlockBase implements ContainerFactoryPluginIn
    * @param string $plugin_definition
    *   The plugin implementation definition.
    */
-  public function __construct(array $configuration, $plugin_id, $plugin_definition, CurrentRouteMatch $current_route_match, EntityManager $entity_manager
-  ) {
+  public function __construct(array $configuration, $plugin_id, $plugin_definition, CurrentRouteMatch $current_route_match, EntityManager $entity_manager, ContextProviderInterface $solution_context) {
     parent::__construct($configuration, $plugin_id, $plugin_definition);
```
sandervd: Why the change to use a context, as the block access is determined by the route?

idimopoulos: Hi Sander, 
The only reason for doing it is consistency. In the collection module, and the collection content block, we are using the context to provide the data for the block and the blockAccess() method to check visibility. 
One way or another, the entity should be there if needed and if not there, checks to prevent rendering of the block are implemented. But I thought that keeping the coding philosophy among similar pieces of code can help in the future.

PR #96, opened 2016-06-06
Isaicp 2530

idimopoulos:


tests/features/demo/demo_june.feature
```diff
@@ -0,0 +1,202 @@
...
+      | S.H.I.E.L.D. | Nick Fury | administrator, facilitator, member |
+      | x-Men        | Wolverine | facilitator, member                |
+      | Avengers     | Wolverine | member                             |
+    And news content:
+      | title                       | field_news_kicker                       | body                                                                                                                                |
+      | Phil Coulson is down        | Phil Coulson fell by the hands of Loki. | Phil Coulson tried to stop Loki from escaping and was killed by him.                                                                |
+      | Phoenix is down             | Wolverine took down Jean Gray.          | In an epic battle, Wolverine had to give the final blow to his great love, Jean Gray as she lost control to the Phoenix inside her. |
+      | S.H.I.E.L.D. is infiltrated | Winter soldier was spotted in action.   | As S.H.I.E.L.D. Hellicarrier is being taken down by the Winter soldier, we are also trying to spot the Hydra agents.                |
+    # The field_news_parent refers to a parent solution.
```
pfrenssen: OK good that you clarify it here. I'm guessing these steps are never intended to be executed by hand on an UAT environment right? Then it's fine to take some liberties with the 'business language' aspect of BDD.

tests/features/demo/demo_june.feature
```diff
@@ -0,0 +1,202 @@
...
+    When I click "S.H.I.E.L.D"
+    # Solutions belonging to the collection.
+    Then I should see the text "Avengers initiative"
+    And I should see the text "Project Tahiti"
+    And I should see the text "Hellicarrier"
+    And I should see the text "Project 'Captain America'"
+    # News belonging to the solution.
+    And I should see the text "Phil Coulson is down"
+    And I should see the text "S.H.I.E.L.D. is infiltrated"
+    # News from solutions.
```
pfrenssen: The difference between 'News belonging to the solution' and 'News from solutions' is not so clear.

tests/features/demo/demo_june.feature
```diff
@@ -0,0 +1,202 @@
...
+    # Collection overview.
+    When I click "S.H.I.E.L.D"
+    # Solutions belonging to the collection.
+    Then I should see the text "Avengers initiative"
+    And I should see the text "Project Tahiti"
+    And I should see the text "Hellicarrier"
+    And I should see the text "Project 'Captain America'"
+    # News belonging to the solution.
+    And I should see the text "Phil Coulson is down"
+    And I should see the text "S.H.I.E.L.D. is infiltrated"
```
pfrenssen: This is not correct I think, the "Phil Coulson is down" news article is not associated with a solution, it is directly associated with the "S.H.I.E.L.D" collection.

web/modules/custom/collection/collection.behat.inc
```diff
@@ -23,6 +25,8 @@ class CollectionSubContext extends DrupalSubContextBase implements DrupalSubCont
   use FileTrait;
   use RandomGeneratorTrait;
   use RdfEntityTrait;
+  use NodeTrait;
+  use OgMenuTrait;
```
pfrenssen: Minor: alphabetical ordering.

web/modules/custom/collection/src/ContextProvider/CollectionRouteContext.php
```diff
@@ -88,20 +87,30 @@ public function getRuntimeContexts(array $unqualified_context_ids) {
...
+    $context = new Context($collection_context_definition, $value);
     $context->addCacheableDependency($cacheability);
-
     $result['collection'] = $context;
 
+    // Also create a context definition to be available for the og_menu block.
+    $collection_context_definition = new ContextDefinition('entity', $this->t('Organic group provided by collection'), FALSE);
+    $context = new Context($collection_context_definition, $value);
+    $context->addCacheableDependency($cacheability);
+    $result['og'] = $context;
```
pfrenssen: This is strange, we have two identical contexts now? That is not necessary. We should either provide the 'collection' context, or the 'og' context, but not both.

I think now that we have the 'og' context which is provided by OG Menu, that we probably can get rid of our 'collection' context, and use 'og'.

web/modules/custom/joinup_news/joinup_news.routing.yml
```diff
@@ -1,5 +1,5 @@
 # Route to create a news entity that belongs to an rdf_entity.
-custom_page.rdf_entity_news.add:
+joinup_news.rdf_entity_news.add:
```
pfrenssen: :+1:

web/modules/custom/collection/src/ContextProvider/CollectionRouteContext.php
```diff
@@ -88,20 +87,30 @@ public function getRuntimeContexts(array $unqualified_context_ids) {
...
    * {@inheritdoc}
    */
   public function getAvailableContexts() {
-    $context = new Context(new ContextDefinition('entity:rdf_entity:collection', $this->t('Collection from URL')));
-    return ['collection' => $context];
+    $collection_context = new Context(new ContextDefinition('entity:rdf_entity:collection', $this->t('Collection from URL')));
+    $og_context = new Context(new ContextDefinition('og', $this->t('Collection from URL')));
+    return [
+      'collection' => $collection_context,
+      'og' => $og_context,
```
pfrenssen: Same here, let's not expose two identical contexts.

web/modules/custom/joinup_news/src/Controller/NewsController.php
```diff
@@ -51,8 +52,13 @@ public function add(RdfInterface $rdf_entity) {
...
     // and that the user has the permission to create news.
-    if (in_array($rdf_entity->bundle(), ['collection', 'solution']) && $this->currentUser()->hasPermission('create rdf entity news')) {
-      return AccessResult::allowed();
+    if (in_array($rdf_entity->bundle(), ['collection', 'solution'])) {
+      if ($this->currentUser()->hasPermission('create rdf entity news')) {
+        return AccessResult::allowed();
+      }
+      if (OgAccess::userAccess($rdf_entity, 'create rdf entity news')->isAllowed()) {
+        return AccessResult::allowed();
+      }
```
pfrenssen: All this will be outdated and will start failing very soon when the latest changes in OG are merged in :) This can be as early as this evening.

It is also not the correct approach, this is checking a global permission, while it should check the per-group permission. For the moment this is fine though, the per-group permissions are not merged in OG yet. So this is for the moment the best solution we can do.

We should keep track of this technical debt though since it is known that this will become problematic in the future. Can you please make a followup ticket that we still need to fix this and put a @todo here with a link to the ticket? Thanks!

web/modules/custom/rdf_entity/src/Entity/Query/Sparql/Query.php
```diff
@@ -195,7 +195,7 @@ public function condition($property, $value = NULL, $operator = '=', $langcode =
             $label_list = "(<" . implode(">, <", array_unique(array_keys($mapping[$this->entityTypeId]))) . ">)";
             $this->condition->condition('?entity', '?label_type', '?label');
             $this->filter->filter('?label_type IN ' . $label_list);
-            $this->filter->filter('regex(?label, "' . $value . '", "i")');
+            $this->filter->filter('?label IN ("' . $value . '")');
```
pfrenssen: Oh? Was this a problem? case sensitivity?

web/modules/custom/solution/src/Controller/SolutionController.php
```diff
@@ -58,6 +59,9 @@ public function createSolutionAccess(RdfInterface $rdf_entity) {
     if ($rdf_entity->bundle() == 'collection' && $this->currentUser()->hasPermission('propose solution rdf entity')) {
       return AccessResult::allowed();
     }
+    if (OgAccess::userAccess($rdf_entity, 'propose solution rdf entity')->isAllowed()) {
+      return AccessResult::allowed();
+    }
```
pfrenssen: This is a workaround, not the correct solution, this needs a followup ticket and a @todo with a link to the ticket.

This will also start failing soon, the entire `OgAccess` class has been refactored.

web/profiles/joinup/src/Plugin/Block/AddContentBlock.php
```diff
@@ -98,25 +98,27 @@ public function build() {
...
-        '#access' => $page_url->access(),
-      ];
+      if ($page_url->access()) {
+        $links['custom_page'] = [
+          '#type' => 'link',
+          '#title' => $this->t('Add custom page'),
+          '#url' => $page_url,
+          '#attributes' => ['class' => ['button', 'button--small']],
+        ];
+      }
```
pfrenssen: Aha! I think this fixes another ticket which is currently in sprint, ISAICP-2547. Great!

pfrenssen: Little tip: if you want you can do this smarter :)

You can do something like this after compiling the links:

```
$links = array_filter($links, function ($link) {
  return $link['#url']->access();
});
```

web/profiles/joinup/src/Plugin/Block/AddContentBlock.php
```diff
@@ -216,6 +224,7 @@ public function getCacheContexts() {
       'collection',
       'asset_release',
       'solution',
+      'url.path',
```
pfrenssen: Oh yes indeed! Good catch!

tests/features/demo/demo_june.feature
idimopoulos: True, plus the default context that the "And "type" content" does not support labels rather than only machine names. I don't know if it is worth of duplicating the behat phrase since e.g. the 'field_news_kickter' is not that bad. It is not like field_content_text_3. :D

tests/features/demo/demo_june.feature
idimopoulos: Sorry, this is a typo. The first one is actually news belonging to the collection (not a solution) and the second one is news belonging to a solution. Will be fixed.

tests/features/demo/demo_june.feature
idimopoulos: True.

web/modules/custom/collection/collection.behat.inc
idimopoulos: It would be so nice if reformatting in PhpStorm could do this automatically as it does I think with global use statements. :)

web/modules/custom/collection/src/ContextProvider/CollectionRouteContext.php
idimopoulos: Is it better? If you consider that we do not load anything two times (because we are assigning one value in two indexes) there is not much impact.
On the other hand, the og context is to be able to use the api of a contrib module (i.e. og_menu) because it's context does not cover our needs in total. On our side, we have blocks that specifically need the collection context. If we leave both, we keep these functionalities separate.
I can remove the collection context though and only provide the og.
We cannot use the context from the og_menu as it does not extend the context to the group entities.

web/modules/custom/joinup_news/src/Controller/NewsController.php
idimopoulos: Sure, I will do. I really look forward to having a stable og :)

web/modules/custom/rdf_entity/src/Entity/Query/Sparql/Query.php
idimopoulos: The actual problem is kind of a really weird problem that I did not have the time to really look into and it is regarding behat tests.
When creating an entity using a table, (like "Given collections: followed by a table of columns of field values) and in this entity there is a reference field, it seems that the reference field handler does not filter the given values. 
As a result, if a blank value is given, it does not consider it as empty, thus removing it, but tries to load from the database using that value.
The above code had the corresponding mysql query 'WHERE label LIKE '%%' so it matched all entities when there should be none. 
Strict checking with the IN method here works fine. This covers behat and similar cases that we might had run onto in the future.

web/modules/custom/solution/src/Controller/SolutionController.php
idimopoulos: Will do as well.

web/profiles/joinup/src/Plugin/Block/AddContentBlock.php
idimopoulos: Yup! There are actually two tickets that are marked as fulfilled by this ticket :)
This is indeed a nice one. I need to start using these inline functions though I must admit that since they are not that popular in terms of usage, it might not be that easy for newcomers to quickly adapt! but I like it! I will change it.

idimopoulos: For some reason, the results were problematic and I really do not have the time to do it right now.! I'm going to keep the classic way of checking.

sandervd: Created ISAICP-2575 as a follow-up for this...

PR #97, opened 2016-06-09
Isaicp 2510

sandervd:


tests/features/asset_distribution/add_asset_distribution.feature
```diff
@@ -16,9 +16,10 @@ Feature: Add asset distribution through the UI
       | title         | Asset release random name        |
       | description   | Some reusable random description |
       | is version of | Asset random name                |
-
     When I am logged in as a "moderator"
     And I go to the homepage of the "Asset release random name" asset release
+
+    Then I click "Add"
```
idimopoulos: Maybe this should be "Then I click the + sign" or 'the plus icon' or something similar. There is no 'Add' in the page.

web/profiles/joinup/config/install/image.style.profile_icon.yml
```diff
@@ -0,0 +1,13 @@
+langcode: en
+status: true
+dependencies: {  }
+name: profile_icon
+label: 'Profile icon'
+effects:
+  d5edf7b7-b996-4fae-a966-9908a9adb054:
+    uuid: d5edf7b7-b996-4fae-a966-9908a9adb054
```
idimopoulos: Is this UUID necessary? There are some config files that require it but if it is not mandatory, it should be removed.

web/profiles/joinup/templates/joinup-tiles.html.twig
```diff
@@ -0,0 +1,39 @@
+{#
+/**
+ * @file
+ * Default theme implementation to display a view of unformatted rows.
+ *
+ * Available variables:
+ * - title: The title of this group of rows. May be empty.
+ * - rows: A list of the view's row items.
+ *   - attributes: The row's HTML attributes.
+ *   - content: The row's content.
```
idimopoulos: Is this the actual entity?

web/themes/joinup/joinup_theme.theme
```diff
@@ -1,6 +1,42 @@
...
  */
+
+use Drupal\image\Entity\ImageStyle;
+use \Drupal\file\Entity\File;
+
+/**
+ * Implements hook_preprocess_page().
+ */
+function joinup_theme_preprocess_page(&$variables) {
+  // Set page title.
```
idimopoulos: Is the title really not available to the variables of the page by default? I mean, really? we need to provide this?

web/themes/joinup/joinup_theme.theme
```diff
@@ -1,6 +1,42 @@
...
+function joinup_theme_preprocess_menu(&$variables) {
+  if ($variables['menu_name'] != 'account') {
+    return;
+  }
+  // Set the account logo.
+  $account = \Drupal::currentUser()->getAccount();
+  $user = user_load($account->id(), TRUE);
+  /** @var \Drupal\image\Plugin\Field\FieldType\ImageItem $photo */
+  $photo = $user->get('field_user_photo')->first();
+  $variables['profile_icon'] = '/themes/joinup/images/user-profile-icon.png';
```
idimopoulos: It's good to make sure that there is an icon appearing, but we should also make sure that the user cannot create an account without a picture or at least set a default picture if none is supplied.

web/themes/joinup/joinup_theme.theme
```diff
@@ -1,6 +1,42 @@
...
+  if ($photo) {
+    $reference = $photo->getValue();
+    /** @var \Drupal\file\Entity\File $file */
+    $file = File::load($reference['target_id']);
```
idimopoulos: If the $reference is empty (i.e. the user does not have a picture) then this will throw a warning as the getValue() returns empty. We should either use ->target_id instead of ->getValue() which I think returns the value itself or null or we should check.

web/profiles/joinup/config/install/image.style.profile_icon.yml
sandervd: This is exported when running drush cde, so even when removing is possible, it would get re-added each time you export...

web/themes/joinup/joinup_theme.theme
sandervd: Yes, by default the title is only accessible from within a block...
As depending on the page the layout will change considerably (blue header entity edit pages vs. collection page) I'd prefer to have a simple variable that contains the page title to print.

web/themes/joinup/joinup_theme.theme
sandervd: If the user doesn't have a picture $photo == FALSE, so we default to user-profile-icon.png

web/profiles/joinup/templates/joinup-tiles.html.twig
sandervd: Switched to using DS.

PR #107, opened 2016-06-21
Isaicp 2538

idimopoulos:


web/modules/custom/demo_content/demo_content.info.yml
```diff
@@ -0,0 +1,5 @@
+name: Demo content
+type: module
+description: Demo content for Joinup project
+core: 8.x
+package: joinup
```
brummbar: Joinup with uppercase, or it will be shown in a separate group (this happens on my local).

web/modules/custom/demo_content/demo_content.install
```diff
@@ -0,0 +1,221 @@
+<?php
+
+/**
+ * @file
+ * Includes demo content for the joinup project.
```
brummbar: I would use the standard @file documentation here "Install, update and uninstall functions for the <name> module".

web/modules/custom/demo_content/demo_content.install
```diff
@@ -0,0 +1,221 @@
...
+      if (!is_file($file_path)) {
+        throw new \Exception("File '{$values['field_ar_logo']}' was not found in file path '$asset_path'.");
+      }
+
+      $uri = file_unmanaged_copy($file_path, 'public://' . $values['field_ar_logo']);
+      $file = File::create(['uri' => $uri]);
+      $values['field_ar_logo'] = [$file];
+    }
+
+    $values['rid'] = 'collection';
```
brummbar: Should we move this rid value to the collection data, like it is done for solution?

web/modules/custom/demo_content/demo_content.install
```diff
@@ -0,0 +1,221 @@
...
+  // @todo: This function includes commented lines. These lines refer to
+  // unimplemented fields and entities and are there for future convenience.
+  // When needed, remove the comments to add the values to the entity.
+  return [
+    [
+      'id' => 'http://www.w3.org/ns/person#',
+      'rid' => 'solution',
+      'label' => 'Core Person Vocabulary',
+      'field_is_creation_date' => '2012-05-07T09:35:24+02:00',
+      'field_is_description' => '<![CDATA[<h3>
```
brummbar: CDATA tags need to be removed.

web/modules/custom/demo_content/demo_content.module
```diff
@@ -0,0 +1,24 @@
...
+    // Main module help for the demo_content module.
+    case 'help.page.demo_content':
+      $output = '';
+      $output .= '<h3>' . t('About') . '</h3>';
+      $output .= '<p>' . t('Demo content for Joinup project') . '</p>';
+      return $output;
+
+    default:
+  }
+}
```
brummbar: I would drop the default help function, because it doesn't have any value.
Maybe swap also the @file comment with something more in line with the usual messages.

PR #114, opened 2016-06-30
Isaicp 2545

sandervd:


web/modules/custom/collection/collection.behat.inc
```diff
@@ -99,7 +99,11 @@ class CollectionSubContext extends DrupalSubContextBase implements DrupalSubCont
    *   Thrown when a collection with the given title does not exist.
    */
   protected function getCollectionByName($title) {
-    return $this->getRdfEntityByLabel($title, 'collection');
+    $collection = $this->getRdfEntityByLabel($title, 'collection');
+    if ($collection) {
```
idimopoulos: Why is this here? we are using a debug step "Then I delete the <collection> collection' in order to delete collections created from the UI.

idimopoulos: After talking with @sandervd, we decided that this will be reverted as we already use the debug step "Then I delete the <collection_label> collection" to clean entities created from UI.

PR #115, opened 2016-07-01
Isaicp 2592

idimopoulos:


build.xml
```diff
@@ -182,52 +182,44 @@
     <!-- Create demo users. -->
     <target name="create-demo-users" description="Create demo users.">
         <foreach list="${drupal.demo.users}" param="drupal.demo.user" target="create-demo-user" delimiter="," />
-        <!-- Create a user with only "authenticated user" role assigned. -->
+        <foreach list="${drupal.demo.authenticated_users}" param="drupal.demo.user" target="create-demo-authenticated_user" delimiter="," />
+    </target>
+
+    <!-- Create a demo user. Subtarget of "create-demo-users". -->
+    <target name="create-demo-user">
```
pfrenssen: Why all these changes? Nothing is mentioned in the ticket about changing the demo users that are created by Phing. It also doesn't seem like anything really changes, there is still one user created, but now there are two targets that do almost the same thing.

web/profiles/joinup/joinup.info.yml
```diff
@@ -39,7 +40,6 @@ dependencies:
   - rdf_entity
   - rdf_export
   - rdf_taxonomy
-  - solution
   - state_machine
```
pfrenssen: Why make this change?

web/profiles/joinup/joinup.behat.inc
```diff
@@ -68,6 +72,35 @@ class JoinupSubContext extends DrupalSubContextBase implements DrupalSubContextI
   }
 
   /**
+   * Creates and authenticates a user with the given og role(s).
+   *
+   * @Given I am logged in as a user with the :role role(s) of the :rdf_entity :rdf_entity_bundle
+   * @Given I am logged in as a/an :role of the :rdf_entity :rdf_entity_bundle
```
pfrenssen: Please document that multiple comma-separated roles can be passed.

tests/features/asset_distribution/add_asset_distribution.feature
```diff
@@ -5,22 +5,19 @@ Feature: Add asset distribution through the UI
...
+    And the following collection:
+      | title      | Asset Distribution Test |
+      | logo       | logo.png                |
+      | affiliates | Asset random name       |
     And the following asset release:
       | title         | Asset release random name        |
       | description   | Some reusable random description |
       | is version of | Asset random name                |
-    When I am logged in as a "moderator"
+    When I am logged in as a "facilitator" of the "Asset random name" solution
```
pfrenssen: It's great to have this step!

tests/features/asset_distribution/add_asset_distribution.feature
```diff
@@ -5,22 +5,19 @@ Feature: Add asset distribution through the UI
...
-      | logo  | logo.png                |
-    And the following solution:
+    Given the following solution:
       | title       | Asset random name                |
       | description | Some reusable random description |
-      | collection  | Asset Distribution Test          |
+    And the following collection:
+      | title      | Asset Distribution Test |
+      | logo       | logo.png                |
+      | affiliates | Asset random name       |
```
pfrenssen: It's confusing to me that this solution is called "Asset random name", it makes it sound like it is an asset instead of a solution.

tests/features/asset_distribution/add_asset_distribution.feature
```diff
@@ -5,22 +5,19 @@ Feature: Add asset distribution through the UI
...
       | title         | Asset release random name        |
       | description   | Some reusable random description |
       | is version of | Asset random name                |
-    When I am logged in as a "moderator"
+    When I am logged in as a "facilitator" of the "Asset random name" solution
     And I go to the homepage of the "Asset release random name" asset release
-
-    # Click the + button.
-    Then I click "Add"
     Then I should see the link "Add distribution"
```
pfrenssen: Isn't it necessary to click the "Add" link in normal use? Remember that these scenarios are intended to describe the user behaviour. The user doesn't see the link "Add distribution" on the screen at this point, because it is hidden behind the plus button.

tests/features/asset_distribution/api.feature
```diff
@@ -5,22 +5,22 @@ Feature: Asset distribution API
   I need to be able to use the Solution API
 
   Scenario: Programmatically create a distribution
+    And the following solution:
```
pfrenssen: Change this to `Given`, it's weird to start the first step with `And`.

tests/features/asset_release/api.feature
```diff
@@ -4,12 +4,6 @@ Feature: Solution API
   I need to be able to use the Solution API
 
   Scenario: Programmatically create a solution
-    Given the following collection:
-      | title             | Solution API foo |
-      | logo              | logo.png         |
-      | moderation        | yes              |
-      | closed            | yes              |
-      | elibrary creation | facilitators     |
     And the following solution:
```
pfrenssen: Another `Given` and `And` switcharoo.

tests/features/custom_page/api.feature
```diff
@@ -6,15 +6,16 @@ Feature: Custom page API
...
+      | owner             | Rufus Drumknott                             |
+      | logo              | logo.png                                    |
+      | moderation        | yes                                         |
+      | closed            | yes                                         |
+      | elibrary creation | facilitators                                |
     And custom_page content:
-      | title      | body                                     | groups audience |
-      | Dummy page | This is some dummy content like foo:bar. | Le Foie Heureux |
+      | title      | body                                     | groups audience                             |
+      | Dummy page | This is some dummy content like foo:bar. | http://joinup.eu/custom_page_api_collection |
```
pfrenssen: :( Doesn't this work any more? It's much more preferable to use the title instead of the URL.

tests/features/demo/demo_june.feature
```diff
@@ -4,18 +4,19 @@ Feature: June 2016 demo
   I should be able to interact with the website and manage content.
 
   Scenario: Manage collection and view collection scenarios.
-    Given collections:
-      | title          | description                           | logo     | moderation |
-      | S.H.I.E.L.D.   | Well, they are mostly flying around.  | logo.png | yes        |
-      | x-Men          | Based on Proffessor Xavier's mantion. | logo.png | no         |
-      | Avengers       | Based on Tony stark's tower.          | logo.png | yes        |
-      | Fantastic four | Based on Reed Richard's tower.        | logo.png | yes        |
     And solutions:
```
pfrenssen: `Given`

tests/features/demo/demo_june.feature
```diff
@@ -4,18 +4,19 @@ Feature: June 2016 demo
...
-      | Avengers initiative       | Gather the strongest into a group.         | text.pdf      | no         | S.H.I.E.L.D. |
-      | Project Tahiti            | Rejuvenate deadly wounds and erase memory. | text.pdf      | yes        | S.H.I.E.L.D. |
-      | Hellicarrier              | Provide a flying fortress as headquarters. | text.pdf      | no         | S.H.I.E.L.D. |
-      | Project 'Captain America' | Bring 'Captain america' back into action.  | text.pdf      | yes        | S.H.I.E.L.D. |
+      | title                     | description                                | documentation | moderation |
+      | Avengers initiative       | Gather the strongest into a group.         | text.pdf      | no         |
+      | Project Tahiti            | Rejuvenate deadly wounds and erase memory. | text.pdf      | yes        |
+      | Hellicarrier              | Provide a flying fortress as headquarters. | text.pdf      | no         |
+      | Project 'Captain America' | Bring 'Captain america' back into action.  | text.pdf      | yes        |
+    Given collections:
```
pfrenssen: `And`

tests/features/demo/demo_june.feature
```diff
@@ -4,18 +4,19 @@ Feature: June 2016 demo
...
+      | title          | description                           | logo     | moderation | affiliates                                                                   |
+      | S.H.I.E.L.D.   | Well, they are mostly flying around.  | logo.png | yes        | Avengers initiative, Project Tahiti, Hellicarrier, Project 'Captain America' |
+      | x-Men          | Based on Proffessor Xavier's mantion. | logo.png | no         |                                                                              |
```
pfrenssen: Typos :D Should be 'professor' and 'mansion' :D

tests/features/demo/demo_june.feature
```diff
@@ -157,6 +164,7 @@ Feature: June 2016 demo
       | Headline | New York under attack                                                                    |
       | Kicker   | S.H.I.E.L.D. to nuke New York?                                                           |
       | Content  | In a desperate attempt to stop the nuke, Nick fury shot down an airplane of S.H.I.E.L.D. |
+    And I select "Validated" from "State"
```
pfrenssen: Why is this changed in scope of this ticket? This is not about workflow. I guess this was working before?

tests/features/joinup_document/api.feature
```diff
@@ -6,13 +6,14 @@ Feature: Document API
...
+      | owner             | Joinup Derp                              |
+      | logo              | logo.png                                 |
+      | moderation        | yes                                      |
+      | closed            | yes                                      |
+      | elibrary creation | facilitators                             |
     And document content:
-      | title    | field_document_short_title | body               | collection              |
-      | JD title | Short                      | Dummy description. | http://joinup.eu/jd/jde |
+      | title    | field_document_short_title | body               | og_group_ref                             |
+      | JD title | Short                      | Dummy description. | http://joinup.eu/document_api_collection |
```
pfrenssen: This is a regression from a BDD standpoint. `og_group_ref` has no meaning for our business owners, but they do understand `collection`. Let's make sure we can still talk about collections and solutions in this step.

tests/src/Traits/OgTrait.php
```diff
@@ -0,0 +1,104 @@
...
+
+/**
+ * Contains helper methods regarding the organic groups.
+ *
+ * @package src\Traits
+ */
+trait OgTrait {
+
+  /**
+   * User with a given og role is already logged in.
```
pfrenssen: "Checks that a user with a given OG role is already logged in"

tests/src/Traits/OgTrait.php
```diff
@@ -0,0 +1,104 @@
...
+   */
+  protected function loggedInWithOgRole($roles, EntityInterface $entity) {
+    if (!$this->loggedIn() || !$this->user) {
+      return FALSE;
+    }
+    $user = \Drupal::entityTypeManager()->getStorage('user')->loadUnchanged($this->user->uid);
+    $membership = Og::getMembership($user, $entity);
+    if (empty($membership)) {
+      return FALSE;
+    }
```
pfrenssen: This is not reliable, a membership entity might be returned here in which the user has the role `non-member`. Use `Og::isMember()` instead.

tests/src/Traits/OgTrait.php
```diff
@@ -0,0 +1,104 @@
...
+   *
+   * @param array $roles
+   *   An array of roles to check.
+   * @param \Drupal\Core\Entity\EntityInterface $entity
+   *   A single role, or multiple comma-separated roles in a single string.
+   *
+   * @return bool
+   *   Returns TRUE if the current logged in user has this role (or roles).
+   */
+  protected function loggedInWithOgRole($roles, EntityInterface $entity) {
```
pfrenssen: I don't see the point of this method. A user doesn't "have an OG role" outside of the context of a particular group. It all depends on the group that is being tested. If the user has this role in group A, but not in group B then this will still return `TRUE` if the test is working in context of group B. This is a timebomb that will result in false positives in the future. This needs at minimum the group passed in.

tests/src/Traits/OgTrait.php
```diff
@@ -0,0 +1,104 @@
...
+   *    The organic group entity.
+   * @param array $roles
+   *    An array of roles to be passed to the membership. The full id should be
+   *    passed.
+   *
+   * @throws \Exception
+   *    Throws an exception when the user is anonymous or the entity is not a
+   *    group.
+   */
+  protected function subscribeUserToGroup($user, EntityInterface $entity, $roles = []) {
```
pfrenssen: Type hint the `$user` as an `AccountInterface`.

web/modules/custom/asset_distribution/src/Controller/AssetDistributionController.php
```diff
@@ -48,18 +49,14 @@ public function add(RdfInterface $rdf_entity) {
...
-    // Check that the passed in RDF entity is a solution, and that the user
-    // has the permission to create distribution entities.
-    // @todo Collection owners and facilitators should also have the right to
-    //   create distributions related to the solutions they manage.
-    // @see https://webgate.ec.europa.eu/CITnet/jira/browse/ISAICP-2450
-    if ($rdf_entity->bundle() == 'asset_release' && $this->currentUser()
-        ->hasPermission('create asset_distribution rdf entity')
-    ) {
-      return AccessResult::allowed();
+    $solution = asset_distribution_get_release_solution($rdf_entity);
```
pfrenssen: Don't call custom functions in the global space from inside OO code.

web/modules/custom/asset_distribution/src/Controller/AssetDistributionController.php
```diff
@@ -48,18 +49,14 @@ public function add(RdfInterface $rdf_entity) {
...
+    $user = \Drupal::currentUser();
```
pfrenssen: Don't call this function, it's in the global scope. Instead inject the `AccountProxy` service and get the user object from there.

web/modules/custom/asset_distribution/asset_distribution.module
```diff
@@ -70,3 +70,20 @@ function asset_distribution_rdf_load_bundle_alter($entity_values, &$bundle) {
...
+/**
+ * Returns the solution that a release belongs to.
+ *
+ * @param \Drupal\rdf_entity\RdfInterface $asset_release
+ *    The asset release rdf entity.
+ *
+ * @return \Drupal\rdf_entity\RdfInterface
+ *    The solution rdf entity that the release is version of.
+ */
+function asset_distribution_get_release_solution(RdfInterface $asset_release) {
```
pfrenssen: This should be put inside a service so it can be called from inside OO code.

web/modules/custom/asset_distribution/asset_distribution.module
```diff
@@ -70,3 +70,20 @@ function asset_distribution_rdf_load_bundle_alter($entity_values, &$bundle) {
...
+  if ($asset_release->bundle() != 'asset_release') {
+    return NULL;
+  }
+  $target_id = $asset_release->field_isr_is_version_of->first()->target_id;
+  return Rdf::load($target_id);
```
pfrenssen: When you convert this to a service, make sure to change this to `$this->entityTypeManager->getStorage('rdf')->create()`.

web/modules/custom/asset_release/asset_release.module
```diff
@@ -22,13 +22,12 @@ function asset_release_entity_bundle_field_info_alter(&$fields, EntityTypeInterf
  * Implements hook_ENTITY_TYPE_create().
  *
  * This takes care of the reverse reference.
- * When a release is created, it has it's isVersionOf prefilled
+ * When a release is created, it has it's isVersionOf pre-filled
```
pfrenssen: :+1:

web/modules/custom/asset_release/asset_release.module
```diff
@@ -22,13 +22,12 @@ function asset_release_entity_bundle_field_info_alter(&$fields, EntityTypeInterf
...
  * and is connected to its parent. But the parent needs to have the hasVersion
  * field filled as well.
  */
-function solution_rdf_entity_insert(EntityInterface $entity) {
+function asset_release_rdf_entity_insert(EntityInterface $entity) {
```
pfrenssen: Lolz :laughing:

web/modules/custom/asset_release/src/Controller/AssetReleaseController.php
```diff
@@ -73,18 +74,29 @@ public function add(RdfInterface $rdf_entity) {
...
-    // Check that the passed in RDF entity is a collection, and that the user
-    // has the permission to create asset_releases.
-    // @todo Collection owners and facilitators should also have the right to
-    //   create asset_releases for the collections they manage.
-    // @see https://webgate.ec.europa.eu/CITnet/jira/browse/ISAICP-2448
-    if ($rdf_entity->bundle() == 'solution' && $this->currentUser()
-        ->hasPermission('create asset_release rdf entity')
-    ) {
-      return AccessResult::allowed();
+    $user = \Drupal::currentUser();
```
pfrenssen: Don't access *any* of the static methods that are in the `Drupal` class from inside OO code. That class is purely intended as an interface between procedural code and OO code. It makes no sense to bridge from OO -> procedural -> OO.

You can directly inject the `AccountProxy` service, to get the current user object from it.

web/modules/custom/collection/collection.behat.inc
```diff
@@ -244,6 +244,16 @@ class CollectionSubContext extends DrupalSubContextBase implements DrupalSubCont
...
 
+    // Handle child solutions.
+    if (!empty($values['field_ar_affiliates'])) {
+      $solutions = explode(',', $values['field_ar_affiliates']);
+      $values['field_ar_affiliates'] = [];
+      foreach ($solutions as $solution) {
+        $entity = $this->getRdfEntityByLabel(trim($solution), 'solution');
+        $values['field_ar_affiliates'][] = $entity->id();
+      }
+    }
```
pfrenssen: Do we have a test for this?

web/modules/custom/demo_content/demo_content.install
```diff
@@ -19,6 +20,15 @@ function demo_content_install() {
  * Creates demo content for the joinup project.
  */
 function demo_content_create_demo_content() {
+  // Create demo users.
+  foreach (demo_content_get_user_data() as $user_data) {
+    $facilitator = user_load_by_name($user_data['name']);
```
pfrenssen: Why would this user already exist?

This module should only be enabled on clean installations, and only on the UAT environment. If there is already data present (e.g. users) then this will cause data loss.

I'd rather have this fail with a fatal error than silently overwrite existing users on who knows which environment somebody might enable this module on.

I think it would be safer to remove this entire module from the repo and move it to a dev-dependency so it will not be present on production environments, but that is out of scope for this issue.

web/modules/custom/demo_content/demo_content.module
```diff
@@ -13,15 +13,15 @@ use Drupal\og\OgMembershipInterface;
...
 function demo_content_rdf_entity_create(EntityInterface $entity) {
-  if ($entity->bundle() !== 'collection') {
+  if (!in_array($entity->bundle(), ['collection', 'solution'])) {
     return;
   }
-
-  // Set 'facilitator' user as a facilitator for the collection.
-  $facilitator = user_load_by_name('facilitator');
+  $user_name = $entity->bundle() . '_facilitator';
+  $facilitator = user_load_by_name($user_name);
```
pfrenssen: Haha, sneaky!!

On a side note, we're losing valuable documentation here. The docblock does not contain any explanation why this hook implementation is necessary, and now the only line that explains it is removed.

web/modules/custom/joinup_news/config/install/core.entity_form_display.node.news.default.yml
```diff
@@ -53,7 +52,6 @@ content:
     third_party_settings: {  }
 hidden:
   created: true
-  field_news_parent: true
```
pfrenssen: Why is this removed? This seems out of scope for this ticket.

web/modules/custom/joinup_news/joinup_news.behat.inc
```diff
@@ -56,38 +58,7 @@ class NewsSubContext extends DrupalSubContextBase implements DrupalSubContextInt
...
-    $news = \Drupal::entityTypeManager()
-      ->getStorage('node')
-      ->loadByProperties(['title' => $title, 'field_news_parent' => $rdf_entity->id()]);
-    /** @var \Drupal\node\NodeInterface $new */
-    $new = reset($news);
-
-    if (empty($new)) {
-      throw new \Exception("The news page titled '$title' was not found or the entity is not related to solution {$rdf_entity->id()}.");
-    }
+    throw new \Exception("The news page '$title' is not associated with {$rdf_entity_bundle} '{$rdf_entity->label()}'.");
```
pfrenssen: :+1:

web/modules/custom/joinup_news/src/Controller/NewsController.php
```diff
@@ -77,21 +77,16 @@ public function createNewsAccess(RdfInterface $rdf_entity) {
...
    *
    * @return \Drupal\Core\Entity\EntityInterface
    *    A node entity.
    */
   protected function createNewsEntity(RdfInterface $rdf_entity) {
-    $field = ($rdf_entity->bundle() == 'collection') ? 'og_group_ref' : 'field_news_parent';
     return $this->entityTypeManager()->getStorage('node')->create([
       'type' => 'news',
-      $field => $rdf_entity->id(),
+      'og_group_ref' => $rdf_entity->id(),
```
pfrenssen: Hah! Good to see this ugly workaround gone.

web/modules/custom/solution/solution.module
```diff
@@ -15,3 +17,38 @@ function solution_entity_bundle_field_info_alter(&$fields, EntityTypeInterface $
...
   }
 }
+
+/**
+ * Implements hook_entity_insert().
+ *
+ * The solution is connected to the collection with the affiliation field in the
+ * collection entity. The process of linking a solution to a collection is:
+ *  - Request an affiliation. This request is registered by adding the
+ *    collection to the affiliation_request field of the solution.
```
pfrenssen: Pff that doesn't seem like a nice workflow to me. That is really unclear for the end user. We should have a separate action for this.

web/modules/custom/solution/src/Controller/SolutionController.php
```diff
@@ -69,29 +42,34 @@ public function add(RdfInterface $rdf_entity) {
...
    * @param \Drupal\rdf_entity\RdfInterface $rdf_entity
-   *   The RDF entity for which the custom page is created.
+   *   The RDF entity for which the solution is created.
    *
    * @return \Drupal\Core\Access\AccessResult
    *   The access result object.
    */
   public function createSolutionAccess(RdfInterface $rdf_entity) {
-    return $this->ogAccess->userAccessEntity('create', $this->createNewSolution($rdf_entity), $this->currentUser());
+    $user = \Drupal::currentUser();
```
pfrenssen: Don't use `\Drupal::xxx()` from OO code.

web/modules/custom/solution/src/Controller/SolutionController.php
```diff
@@ -69,29 +42,34 @@ public function add(RdfInterface $rdf_entity) {
...
+    if (empty($rdf_entity) && !$user->isAnonymous()) {
+      return AccessResult::neutral();
+    }
+    $membership = Og::getMembership($user, $rdf_entity);
+    return (!empty($membership) && $membership->hasPermission('create solution rdf_entity')) ? AccessResult::allowed() : AccessResult::forbidden();
```
pfrenssen: Hehe nice to see this being used :)

build.xml
idimopoulos: The reason behind this is that before we were creating 3 users. A moderator, an administrator and an authenticated user. The first two were mapped with a role identical to their name so we had an extra target only to create an authenticated user.

Now I have added  two more users that will serve as facilitators for collections and solutions so I converted the second target into a loop as well so that I can add the 3 authenticated users.

web/profiles/joinup/joinup.info.yml
idimopoulos: As I have seen, Drupal does not respect the dependencies during the profile installation so the modules must be put there appropriately.
So for example, before this ticket, we were installing the collection, and then the joinup news. But not that the solution is an og, news's field og_group_ref, got another dependency, which is the solution entity. This immediately broke my installation due to unmet dependencies.

tests/features/asset_distribution/add_asset_distribution.feature
idimopoulos: It is an asset. The solutions are dcat:asset. But since we use the term solution, I changed it in the test.!

tests/features/asset_distribution/add_asset_distribution.feature
idimopoulos: You are right! I removed it because it was making my test fail at a point and forgot to put it back or accidentally committed it..!

tests/features/custom_page/api.feature
idimopoulos: The problem I found out is that never actually worked :/ I tried to see this in the UI and there wasn't any connection being created. And due to the fact that these are api.feature tests, we were not seeing the problem. But then other tests started to break, like the behat scenario test.

For the rdf entities, we were having our own create functionality and we were manually loading the object. But the node is being handled by Drupal context directly.

Though, Francesco had an idea with a hook and I am going to look into that.

tests/features/demo/demo_june.feature
idimopoulos: Yeah sorry... :/ We have to have the solutions created in order to affiliate them to collections and I changed this to a lot of tests! swapped them...

tests/features/demo/demo_june.feature
idimopoulos: Removed! My test was failing, thought it was for that, turns out it wasn't.

tests/features/joinup_document/api.feature
idimopoulos: It is the same comment as above. This was a false positive. If you break the test, and see this in UI, it doesn't actually connect the entity to the collection.
And in the api.features we only test "And I have 1 item of this type" so it does not provide assurance about consistency of data.

Still, I will do something using a hook.

tests/src/Traits/OgTrait.php
idimopoulos: Not yet :D this still is not updated to the latest version of Og :) but marking this because the other ticket might go first in so I will have to then update it. thx for the notice.!

tests/src/Traits/OgTrait.php
idimopoulos: Who writes these awful documentation? The $entity object is the group. Updated the doc and renamed the $entity to $group... This documentation was crime against humanity.

For the record, initially, I started to implement it without this method. I did not care that much if I had to force the recreation of a user every time. But I saw that the drupal context actually checks if the current user has the given role before creating a new one so I thought I should mimic it.

tests/src/Traits/OgTrait.php
idimopoulos: oki doki :)

web/modules/custom/asset_distribution/src/Controller/AssetDistributionController.php
idimopoulos: Why not? what is the reason of them being global then? 
The thing here is that I needed the parent of the asset distribution but since there is not any direct access through the entity itself, I thought I should make it global but as part of the module that provides the entity. So I created a global function in the asset distribution module.

Would you have it as a separate method of the AssetDistributionController? 
And what if we would need this for more controllers? Is it still to have duplicates?

web/modules/custom/asset_distribution/asset_distribution.module
idimopoulos: Ahh so that answers my comment from above :D

web/modules/custom/demo_content/demo_content.install
idimopoulos: But in order to keep all our demo users together, I am creating them through phing as demo users with only 'authenticated user' role assigned. They are there in every installation of the site.
This module is not overriding anything.

web/modules/custom/joinup_news/config/install/core.entity_form_display.node.news.default.yml
idimopoulos: Nope. Actually this was the entire ticket :) Previously, we already had implemented the notion of a news entity belonging to a solution, but since the solution was not an Og, a normal reference was needed. Now, the field itself is obsolete. The solution is normally referenced by og_group_ref field.

web/modules/custom/joinup_news/src/Controller/NewsController.php
idimopoulos: Me too :)

web/modules/custom/solution/solution.module
idimopoulos: This documentation is not intended for the end user. The end user should not have immediate interaction with these two fields described here.
The purpose of this documentation, is for the developer to know that due to the ADMS schema, the request is saved in the solution but the affiliation itself is saved in the collection.

I also think that this should be done merely by action buttons through the UI. The end user would never manage to actually go through that process manually, nor he would have permissions to do it (edit a solution to request an affiliation by filling his own collections name there)

web/modules/custom/solution/src/Controller/SolutionController.php
idimopoulos: yay :D

build.xml
pfrenssen: That's not true, these users are created in the install hook of the demo content module. Let's get rid of these changes. There is no reason to make more than one demo user for the same role.

tests/features/joinup_document/api.feature
pfrenssen: So then our test coverage was not complete, we should add a followup to fix this so it works as before.

tests/src/Traits/OgTrait.php
pfrenssen: Oh!! I totally misunderstood this, because of the documention I thought this was a Role or OgRole entity, not a group. Now it makes sense, thanks!

tests/src/Traits/OgTrait.php
pfrenssen: Actually I was wrong, I just had a look at the documentation of `Og::getMembership()`. If you call it without specifying the state, then it will default to `STATE_ACTIVE` and will only return members. This means I have introduced a bug in my latest OG ticket.

build.xml
idimopoulos: Ah, ok I kinda proved myself wrong :D I did initially implement this like this and then thought that it would be wrong. But I still kept these changes in case we would need more than one user. I will revert it. :)

web/modules/custom/asset_distribution/src/Controller/AssetDistributionController.php
pfrenssen: > Why not? what is the reason of them being global then?

If procedural code and OO code is mixed then this breaks encapsulation, one of the core features of OO programming. The only global functions that you should use in OO code are the PHP built-in functions, and even then you should use caution (for example don't do database queries using PHP built-ins).

To make the core reusable, you can put it in a service, and inject the service into the classes that need the code.

build.properties.dist
```diff
@@ -33,6 +33,7 @@ drupal.admin.email = admin@example.com
 # Comma-separated list of demo users to create. The username and password will
 # be taken from the role.
 drupal.demo.users = administrator, moderator
+drupal.demo.authenticated_users = user
```
pfrenssen: This is not needed any more. I was thinking, maybe we can add "authenticated" to `drupal.demo.users`? So we can log in as a normal authenticated user? I'm not sure if that will work though.

build.properties.dist
```diff
@@ -148,6 +149,9 @@ phpcs.prepush.destination = ${project.basedir}/.git/hooks/pre-push
 # Development modules to enable.
 drupal.modules.dev = config_update field_ui menu_ui views_ui
 
+# UAT modules to enable.
+drupal.modules.uat = demo_content
+
```
pfrenssen: Good solution! Can you change the documentation so it's clear to people that do not know what UAT means? Maybe "Modules to enable on acceptance environments".

tests/src/Traits/RdfEntityTrait.php
```diff
@@ -14,26 +14,29 @@
...
+   *   Optional RDF entity type.
    *
    * @return \Drupal\rdf_entity\Entity\Rdf
    *   The RDF entity.
    *
    * @throws \InvalidArgumentException
    *   Thrown when an RDF entity with the given name and type does not exist.
    */
-  protected function getRdfEntityByLabel($title, $type) {
+  protected function getRdfEntityByLabel($label, $type = NULL) {
```
idimopoulos: This is not true. Rdf entities are supposed to have unique titles per bundle and not across site. And of course this is not the case, but for example, a license can have the same name as a collection and a release with it's solution (at least for now).

tests/features/asset_distribution/add_asset_distribution.feature
```diff
@@ -5,20 +5,19 @@ Feature: Add asset distribution through the UI
...
+      | title      | Asset Distribution Test |
+      | logo       | logo.png                |
+      | affiliates | Solution random x name  |
     And the following asset release:
       | title         | Asset release random name        |
       | description   | Some reusable random description |
-      | is version of | Asset random name                |
-    When I am logged in as a "moderator"
+      | is version of | Solution random x name           |
+    When I am logged in as a "facilitator" of the "Solution random x name" solution
```
pfrenssen: This is great stuff, we are no longer limited to being a moderator to administer the site.

tests/src/Traits/OgTrait.php
```diff
@@ -0,0 +1,104 @@
...
+      return FALSE;
+    }
+    $user = \Drupal::entityTypeManager()->getStorage('user')->loadUnchanged($this->user->uid);
+    $membership = Og::getMembership($user, $group);
+    if (empty($membership)) {
+      return FALSE;
+    }
+    if ($roles == $membership->getRolesIds()) {
+      return FALSE;
+    }
```
pfrenssen: Something is missing here, this check will fail if the user has more roles than the ones that are passed in `$roles`. We should probably use something like `array_intersect()` to do the comparison. It's fine to do this in a followup though, I'll add a todo.

idimopoulos: Oh no please don't. This is the intended use. Don't forget that this is a test environment so checks should occur in a controlled environment. This method is used to determine if a user needs to be recreated. If a user has more roles than what is requested, it is still an issue because he might has permissions from roles that are not in the list. The corresponding method from DrupalContext checks for the exact roles and I can imagine it is for this purpose.
We do not want to lack any roles but we do not want to have more roles as well.

pfrenssen: OK but that's not what I am used to in an assertion, that limits the usefulness IMO. If this stays like this it means that this method will not be usable to check if a user has a certain role in a certain group.

If this is only intended to be used with a single test, then don't put it in a trait but use it as a local protected function in the test this belongs to.

pfrenssen: Also, why does this return FALSE instead of throwing an exception when the assertion fails? That means that on the calling end we will always have to check the return value, and throw an exception there.

tests/src/Traits/OgTrait.php
```diff
@@ -0,0 +1,104 @@
...
+   *    The organic group entity.
+   * @param array $roles
+   *    An array of roles to be passed to the membership. The full id should be
+   *    passed.
+   *
+   * @throws \Exception
+   *    Throws an exception when the user is anonymous or the entity is not a
+   *    group.
+   */
+  protected function subscribeUserToGroup($user_id, EntityInterface $entity, $roles = []) {
```
pfrenssen: It's also a good idea to change `$entity` to `$group` here.

tests/src/Traits/RdfEntityTrait.php
pfrenssen: What were you replying to? It seems Github has lost the comment thread :(

web/modules/custom/asset_distribution/asset_distribution.module
pfrenssen: This is no longer needed, it has moved to `AssetDistributionRelations`. If you want to keep it so it can be used from procedural code, then you can replace the logic with a call to the service method, like is commonly done for example in `Drupal`.

web/modules/custom/asset_distribution/src/AssetDistributionRelations.php
```diff
@@ -0,0 +1,46 @@
...
+   * @return \Drupal\rdf_entity\RdfInterface
+   *    The solution rdf entity that the release is version of.
+   */
+  public function getReleaseSolution(RdfInterface $asset_release) {
+    if ($asset_release->bundle() != 'asset_release') {
+      return NULL;
+    }
+    $target_id = $asset_release->field_isr_is_version_of->first()->target_id;
+    return $this->entityTypeManager->getStorage('rdf_entity')->load($target_id);
+  }
```
pfrenssen: Great job!! Looking good :)

web/modules/custom/asset_distribution/src/Controller/AssetDistributionController.php
```diff
@@ -48,18 +74,14 @@ public function add(RdfInterface $rdf_entity) {
...
-    // @see https://webgate.ec.europa.eu/CITnet/jira/browse/ISAICP-2450
-    if ($rdf_entity->bundle() == 'asset_release' && $this->currentUser()
-        ->hasPermission('create asset_distribution rdf entity')
-    ) {
-      return AccessResult::allowed();
+    $solution = $this->assetDistributionRelations->getReleaseSolution($rdf_entity);
+    $user = $this->currentUser();
+    if (empty($solution) && !$user->isAnonymous()) {
+      return AccessResult::neutral();
     }
```
pfrenssen: This check seems weird to me, can you explain? How I read it is that an asset distribution is a child item of a solution, so we actually depend on it. But this check is skipped if the user is anonymous, why is this significant? Shouldn't we always deny access if the `$solution` is missing, even if the user is anonymous?

Also, shouldn't we return `AccessResult::forbidden()` here? We are the main gatekeeper for this route, and if we can determine that the required `$solution` is missing, then we can safely decide that the route should not be accessible, right?

I always understand `AccessResult::neutral()` to mean "no opinion", or "do not have enough information to decide whether or not to grant access".

web/modules/custom/asset_release/src/Controller/AssetReleaseController.php
```diff
@@ -73,18 +74,29 @@ public function add(RdfInterface $rdf_entity) {
...
-    //   create asset_releases for the collections they manage.
-    // @see https://webgate.ec.europa.eu/CITnet/jira/browse/ISAICP-2448
-    if ($rdf_entity->bundle() == 'solution' && $this->currentUser()
-        ->hasPermission('create asset_release rdf entity')
-    ) {
-      return AccessResult::allowed();
+    $user = $this->currentUser();
+    if ($user->isAnonymous()) {
+      return AccessResult::neutral();
     }
```
pfrenssen: It should not be necessary to check if the user is anonymous. `Og::getMembership()` only returns memberships for members, so the `!empty($membership)` test on the next line already covers it.

web/modules/custom/demo_content/demo_content.module
```diff
@@ -11,17 +11,24 @@ use Drupal\og\OgMembershipInterface;
 
 /**
  * Implements hook_ENTITY_TYPE_create().
+ *
+ * We have two organic groups, the collection and the solution, and two users
+ * intended to be used as facilitators, the collection_facilitator and the
+ * solution_facilitator respectively.
+ *
+ * This function will assign the appropriate user as a facilitator to the entity
+ * passed.
```
pfrenssen: This is great documentation!

web/modules/custom/demo_content/demo_content.module
pfrenssen: This comment has been addressed by the addition to the docblock, thanks!

web/modules/custom/joinup_news/config/install/core.entity_form_display.node.news.default.yml
pfrenssen: Ohh! Yes that's clear, thanks for the explanation!

web/modules/custom/solution/solution.module
pfrenssen: OK, I didn't realize this was an ADMS requirement, I thought we were really going to have the user propose to add an affiliation, and have it accepted by the collection facilitator :D

web/modules/custom/solution/solution.module
```diff
@@ -15,3 +17,38 @@ function solution_entity_bundle_field_info_alter(&$fields, EntityTypeInterface $
...
+ *  - Accept the affiliation. The collection is removed from the affiliation
+ *    request of the solution and the solution is added to the affiliates field
+ *    of the collection.
+ * If we want to create a solution and initialize it with a collection affiliate
+ * we simply need to pass the affiliate_request to the new entity before it is
+ * saved. The hook below will take care of automatically linking it.
+ */
+function solution_entity_insert(EntityInterface $entity) {
+  /** @var Rdf $entity */
+  if ($entity->getEntityTypeId() != 'rdf_entity' || $entity->bundle() != 'solution' || empty($entity->field_is_affiliations_requests->first())) {
```
pfrenssen: Since this is only intended to run on entities of type `rdf_entity` and not on all entity types, you can use `hook_ENTITY_TYPE_insert()` instead of `hook_entity_insert()` and save one if condition ;)

web/modules/custom/solution/solution.module
```diff
@@ -15,3 +17,38 @@ function solution_entity_bundle_field_info_alter(&$fields, EntityTypeInterface $
...
+
+  $collection = Rdf::load($affiliate);
+  if (empty($collection) || $collection->bundle() != 'collection') {
+    return;
+  }
+  /** @var \Drupal\rdf_entity\RdfInterface $collection */
+  $collection->set('field_ar_affiliates', $entity->id());
+  $collection->save();
+
+  $entity->set('field_is_affiliations_requests', NULL);
```
pfrenssen: Yeah OK it is clear to me now from reading the code that this is just juggling the data around to satisfy the data model. Maybe just mention it in the documentation that this is intended to satisfy ADMS compliance? It might save other developers some head scratching in the future.

web/profiles/joinup/config/install/og.og_role.rdf_entity-solution-administrator.yml
```diff
@@ -0,0 +1,12 @@
+langcode: en
+status: true
+dependencies: {  }
+id: rdf_entity-solution-administrator
+label: Solutin owner
```
pfrenssen: `Solutin`, this sounds like some kind of medication :D

web/profiles/joinup/config/install/og.og_role.rdf_entity-solution-administrator.yml
```diff
@@ -0,0 +1,12 @@
...
+weight: 2
+group_type: rdf_entity
+group_bundle: solution
+group_id: null
+permissions:
+  - 'request solution deletion'
```
pfrenssen: :+1: Great!!

web/profiles/joinup/config/install/og.og_role.rdf_entity-solution-member.yml
```diff
@@ -0,0 +1,11 @@
...
+status: true
+dependencies: {  }
+id: rdf_entity-solution-member
+label: Member
+weight: 1
+group_type: rdf_entity
+group_bundle: solution
+group_id: null
+permissions: { }
+role_type: immutable
```
pfrenssen: This has changed, this `role_type` is no longer called `immutable`, but `required`. Weird to see it appear like this. I just double checked and this change happened around 2 months ago in OG. Did you work maybe from an installation that was done from an older version of OG?

Or perhaps we're just carrying this around for a long time now. OG doesn't have update hooks yet so the data is never updated when something changes.

web/profiles/joinup/config/install/og.og_role.rdf_entity-solution-non-member.yml
```diff
@@ -0,0 +1,13 @@
...
+id: rdf_entity-solution-non-member
+label: Non-member
+weight: 3
+group_type: rdf_entity
+group_bundle: solution
+group_id: null
+permissions:
+  - 'accept facilitator invitation'
+  - subscribe
+role_type: immutable
```
pfrenssen: Here too, this should be `required` instead of `immutable`.

web/profiles/joinup/joinup.behat.inc
```diff
@@ -68,6 +73,37 @@ class JoinupSubContext extends DrupalSubContextBase implements DrupalSubContextI
...
+   */
+  public function assertAuthenticatedByOgRole($role, $rdf_entity, $rdf_entity_bunlde) {
+    $entity = $this->getRdfEntityByLabel($rdf_entity, $rdf_entity_bunlde);
+    $roles = explode(',', $role);
+    $roles = array_map('trim', $roles);
+    $roles = $this->convertOgRoleNamesToIds($roles, $entity);
+
+    // Check if a user with this role is already logged in.
+    if (!$this->loggedInWithOgRole($roles, $entity)) {
```
pfrenssen: Ah OK this explains why the `loggedInWithOgRole()` returns TRUE|FALSE instead of throwing an exception.

So `loggedInWithOgRole()` is simply not meant as a standalone assertion but rather as a helper method for this particular assertion. Let's move it out of the trait then, so people don't use it in their tests.

web/profiles/joinup/joinup.behat.inc
```diff
@@ -68,6 +73,37 @@ class JoinupSubContext extends DrupalSubContextBase implements DrupalSubContextI
...
+        'name' => $random,
+        'pass' => $random,
+      );
+      $user->mail = "{$user->name}@example.com";
+      $this->userCreate($user);
+      $this->subscribeUserToGroup($user->uid, $entity, $roles);
+
+      // Login.
+      $this->login();
+    }
```
pfrenssen: This code seems to be based on `DrupalContext::assertAuthenticatedByRole()`, let's add an `@see` to the documentation so it is clear that these methods are related.

web/profiles/joinup/joinup.behat.inc
```diff
@@ -305,4 +341,82 @@ class JoinupSubContext extends DrupalSubContextBase implements DrupalSubContextI
...
+    // Check both the node type specific aliases as well as the shared aliases.
+    foreach ([$type, 'shared'] as $key) {
+      if (!empty($aliases[$key][$alias])) {
+        return $aliases[$key][$alias];
+      }
+    }
+
+    return $alias;
+  }
+
```
pfrenssen: I added this part, can you please have a look? I do like your suggestion this morning to get the human readable field labels directly from `EntityFieldManager::getFieldDefinitions()` so that our Behat test scenarios automatically become invalid whenever a field label is changed.

web/profiles/joinup/joinup.info.yml
pfrenssen: What are you serious? That seems like a major bug to me! And a regression, because in Drupal 7 the dependencies were resolved during profile install if I remember correctly.

We need a followup for this, this needs to be investigated and reported in core, so that this can be solved. We cannot be tasked by manually ordering this list according to the dependency chain, that's not something a human should be dealing with, this is a machine's job!

pfrenssen: No it's fine, I just checked it. The modules are installed in `install_profile_modules()`. This retrieves the full module list from `system_rebuild_module_data` which has the proper dependencies resolved in it. Every module gets assigned a weight (taken from the `->sort` parameter in the module data), and this is used then to sort the profile's module. The actual order of the modules in the dependency list is only used for modules that are on the same level (i.e. that have the same dependencies).

Here are the weights that are given to the modules you mention during the actual install:

* `collection`: -45
* `joinup_news`: -117
* `og`: -40
* `solution`: -112

So according to this list, `og` will be installed first, then `collection` -> `solution` -> `joinup_news`

idimopoulos: :( clumsy as hell.. Well, it seems that I had not set the dependencies for the joinup_news module properly so it did not install everything properly.. I reverted the sorting.

web/modules/custom/asset_distribution/asset_distribution.module
idimopoulos: You are right! removed.

web/modules/custom/solution/solution.module
idimopoulos: Hopefully we will not end up doing something that ugly! Unless we are trying to troll the world :D

tests/src/Traits/RdfEntityTrait.php
idimopoulos: Oh, how did this happen? You commented that since rdf_entities have unique titles, we can provide a default value of null and if not provided, search all the rdf entities. 
So my comment was that this is not true as it is unique title per bundle.

tests/src/Traits/OgTrait.php
idimopoulos: Again, this is not a test or an assert! This is merely a check that a user might be already logged in. The whole idea is a copy of RawDrupalContext::loggedInWithRole which is just a helper function. So the whole idea is when a command "Then I am logged in as a user with the "role" role" is found, the context is checking if it needs to re create a user in order to save time. This is just the check put in a different function.

idimopoulos: Check moved to joinup.behat under the only method that is using it.

web/modules/custom/asset_distribution/src/AssetDistributionRelations.php
idimopoulos: Thx :)

web/modules/custom/asset_distribution/src/Controller/AssetDistributionController.php
idimopoulos: I think I was thinking that I should block it without having to return forbidden but you are right! it should be only forbidden. This form has only one intended use.

The <code>AccessResult::neutral()</code> is the issue here.

web/modules/custom/demo_content/demo_content.module
idimopoulos: Is it good though to use 'We'? because there are really a lot of cases that I find it really hard to come up with a way to describe it adjectively and documenting with what 'we' do is so much easier to explain. I mean in terms of best practices.

web/profiles/joinup/config/install/og.og_role.rdf_entity-solution-member.yml
idimopoulos: More or less.! not an old Og version but I copied it from the collection roles. I updated both.

web/profiles/joinup/joinup.behat.inc
idimopoulos: Moved :)

web/profiles/joinup/joinup.behat.inc
idimopoulos: I have already created a ticket for this in: https://webgate.ec.europa.eu/CITnet/jira/browse/ISAICP-2620

web/modules/custom/demo_content/demo_content.module
pfrenssen: Yeah absolutely, it was completely clear to me.

web/modules/custom/asset_distribution/src/AssetDistributionRelations.php
```diff
@@ -0,0 +1,46 @@
...
+  /**
+   * Drupal\Core\Entity\EntityTypeManager definition.
+   *
+   * @var \Drupal\Core\Entity\EntityTypeManager
+   */
+  protected $entityTypeManager;
+
+  /**
+   * Initialize injected objects.
+   */
```
brummbar: Coding standard violation here, @param is missing :P

PR #116, opened 2016-07-04
Isaicp 2609

pfrenssen:
https://webgate.ec.europa.eu/CITnet/jira/browse/ISAICP-2609


build.properties.dist
```diff
@@ -42,8 +42,9 @@ drupal.demo.users = administrator, moderator
 # Paths to executables.
 composer.bin = ${project.basedir}/vendor/bin/composer
 drush.bin = ${project.basedir}/vendor/bin/drush
-phing.bin = ${project.basedir}/vendor/bin/phing
 isql.bin = /usr/bin/isql-vt
+phing.bin = ${project.basedir}/vendor/bin/phing
+solr.bin = ${solr.dir}/bin/solr
```
pfrenssen: I would change this to the symlink in `${project.basedir}/vendor/bin/solr`. This would make it consistent with the other executables.

build.xml
```diff
@@ -227,6 +227,131 @@ sudo /etc/init.d/virtuoso-opensource-6.1 restart"
...
+        <!-- Create a symlink of the solr exec file to vendor/bin directory. -->
+        <if>
+            <available file="${project.basedir}/vendor/bin/solr" type="file" property="file.Exists"/>
+            <then>
+                <echo message="Solr binary already exists in ${project.basedir}/vendor/bin directory. Skip creating symlink" />
+            </then>
+            <else>
+                <symlink
+                        target="${solr.bin}"
+                        link="${project.basedir}/vendor/bin/solr"
```
pfrenssen: This will need to be updated too then, you can set the target to `${solr.dir}/bin/solr`, that is consistent with the other targets like `start-apache-solr` etc.

PR #119, opened 2016-07-08
Isaicp 2614

idimopoulos:


composer.json
```diff
@@ -10,13 +10,13 @@
         "cweagans/composer-patches": "~1.0",
         "drupal/console": "~0.10",
         "drupal/core": "~8.1.0",
-        "drupal/facets" : "~8.1.0-alpha3",
+        "drupal/facets" : "8.1.0-alpha3",
```
pfrenssen: The leading tilde is missing. If this is omitted we will forever be stuck on this alpha version :)

tests/features/collection/collection_overview.feature
```diff
@@ -75,3 +75,6 @@ Feature: Collections Overview
     Then I should see the link "Collections"
     When I click "Collections"
     Then I should see the link "Colonies in space"
+
+    # Debug step.
+    Then I delete the "Colonies in space" collection
```
pfrenssen: This is not exactly for debugging, but for cleaning up the collection that was made manually. This is a normal part of testing. I'll update the documentation.

tests/src/Traits/OgMenuTrait.php
```diff
@@ -1,6 +1,7 @@
 <?php
 
 namespace Drupal\joinup\Traits;
+
```
pfrenssen: Yes, this is actually in the coding standards ;)

web/modules/custom/joinup_search/config/install/facets.facet.type.yml
```diff
@@ -6,7 +6,7 @@ name: Type
 url_alias: type
 weight: 0
 show_only_one_result: false
-field_identifier: search_api_aggregation_1
+field_identifier: aggregated_field
```
pfrenssen: Aha! So this was the culprit?

web/modules/custom/joinup_search/config/install/facets.facet_source.search_api_views__search__page_1.yml
```diff
@@ -0,0 +1,8 @@
+uuid: null
+langcode: en
+status: true
+dependencies: {  }
+id: search_api_views__search__page_1
+name: 'search_api_views:search:page_1'
+filter_key: null
+url_processor: query_string
```
pfrenssen: Why is this added? If this is necessary, can we give it a proper machine name? `page_1` is overly generic, it's not clear what this is used for.

On a minor remark, it is not common to include the `uuid` key and give it the value `null`. This is usually omitted. I don't think it is harmful though.

composer.json
pfrenssen: alpha 4 is even released: https://www.drupal.org/project/facets/releases/8.x-1.0-alpha4

PR #122, opened 2016-07-12
Isaicp 2593

brummbar:


build.properties.dist
```diff
@@ -119,7 +119,7 @@ phpcs.standard = ${project.basedir}/phpcs-ruleset.xml.dist
 phpcs.files = ${website.modules.dir};${website.profiles.dir};${website.themes.dir}
 
 # Paths to ignore, delimited by semicolons.
-phpcs.ignore = ${website.modules.dir}/contrib
+phpcs.ignore = ${website.modules.dir}/contrib;${website.themes.dir}/joinup/prototype/node_modules
```
idimopoulos: Ok I can imagine this is a good improvement when it comes to integrating the theme with the project!

idimopoulos: I have spoken with @pfrenssen, these libraries should not be in the theme folder. In any case, this should not be part of our project so any dev options should be added in your build.properties.local instead.

PR #127, opened 2016-07-20
Isaicp 2404

idimopoulos:


web/modules/custom/contact_information/contact_information.module
```diff
@@ -30,3 +30,17 @@ function contact_information_rdf_entity_presave(EntityInterface $entity) {
   $names = array_column($entity->get('field_ci_name')->getValue(), 'value');
   $entity->set('label', implode(', ', $names));
 }
+
+/**
+ * Implements hook_inline_entity_form_table_fields_alter().
```
idimopoulos: Can you add a short documentation on why you are doing this? I don't think this is much helpful the way it is now. Newcomers will have no idea what this is about.

PR #132, opened 2016-07-25
Isaicp 2404

idimopoulos:


web/modules/custom/contact_information/contact_information.module
```diff
@@ -22,17 +22,15 @@ function contact_information_entity_bundle_field_info_alter(&$fields, EntityType
  *
  * Concatenates the names into a single string. This entity has support for
  * multiple names and entity label does not support this. For this reason,
- * the label actually inherits all values from field_ci_name field.
+ * the label actually concatenates all values from field_ci_name field.
  */
```
brummbar: I would rather move this documentation in the auto_label function, and simplify this one.

web/modules/custom/contact_information/contact_information.module
```diff
@@ -51,3 +49,42 @@ function contact_information_inline_entity_form_table_fields_alter(&$fields, $co
     $fields['label']['label'] = t('Name');
   }
 }
+
+/**
+ * Implements hook_inline_entity_form_entity_form_alter().
+ */
+function contact_information_inline_entity_form_entity_form_alter(&$entity_form, &$form_state) {
+  // Add an entity builder callback to fill the label field with the values of
```
brummbar: Again, this should go in the function implementing it and use a simpler "to fill the label" or similar.

PR #134, opened 2016-07-25
Isaicp 2420

brummbar:


tests/features/solution/solution.edit.feature
```diff
@@ -0,0 +1,69 @@
+@api
+Feature: Solution editing.
+  In order to manage solutions
+  As a moderator or solution facilitator
```
idimopoulos: Remove the moderator! we are only testing as a facilitator. 
We are using the strict model of Og. It means that users with site-wide rights do not have them applied in groups. 
Moderators do not currently have this kind of access.

tests/features/solution/solution.edit.feature
```diff
@@ -0,0 +1,69 @@
...
+      When I fill in "Title" with "Solution A revised"
+      And I press "Save"
+      Then I should see the heading "Solution A revised"
+
+      When I go to the homepage of the "Solution B" solution
+      Then I should not see the link "Edit"
+      When I go to the "Solution B" solution edit form
+      Then I should get an access denied error
+
+    Scenario: A solution facilitator can edit only the solutions he's associated with.
```
idimopoulos: Do you think that Scenario B is necessary?
The permission overlap bug is already tested in the above 4 steps and now Og has a functional test to prevent the overlap, so what is the purpose of this scenario (also given the fact that both tests are only using permissions of a facilitator)?

web/modules/custom/collection/collection.behat.inc
```diff
@@ -379,9 +379,9 @@ class CollectionSubContext extends DrupalSubContextBase implements DrupalSubCont
...
    *   Thrown when a collection is not found.
    *
-   * @Given (the following )user memberships:
+   * @Given (the following )collection user memberships:
    */
-  public function givenUserMemberships(TableNode $membership_table) {
+  public function givenCollectionUserMemberships(TableNode $membership_table) {
     foreach ($membership_table->getColumnsHash() as $values) {
       // Load group.
       $group = $this->getCollectionByName($values['collection']);
```
idimopoulos: Nice :) Since we are having this for more than one bundle, we should move it to a trait. I mean the functionality! not the behat step.

web/modules/custom/owner/owner.behat.inc
```diff
@@ -33,37 +33,81 @@ class OwnerSubContext extends DrupalSubContextBase implements DrupalSubContextIn
...
+   *
+   * Table format:
+   * | name          |
+   * | Mightily Oats |
+   * | John Irwin    |
+   *
+   * @param TableNode $person_table
+   *   The person table.
+   *
+   * @Given (the following )persons:
```
idimopoulos: Even if the entity is called person and the plural is person entities, I wouldn't go against the dictionary just to sound BDD. I would either set people, or person entities. An entity after all is something beyond programming that declares...well..an entity. I don't think it's that bad to use it just because it's heavily used in Drupal.

web/modules/custom/owner/owner.behat.inc
```diff
@@ -33,37 +33,81 @@ class OwnerSubContext extends DrupalSubContextBase implements DrupalSubContextIn
...
+   */
+  public function givenPersons(TableNode $person_table) {
```
idimopoulos: Same as above.

web/modules/custom/solution/solution.behat.inc
```diff
@@ -333,6 +335,51 @@ class SolutionSubContext extends DrupalSubContextBase implements DrupalSubContex
...
+   *
+   * @throws \Exception
+   *   Thrown when a solution is not found.
+   *
+   * @Given (the following )solution user memberships:
+   */
+  public function givenSolutionUserMemberships(TableNode $membership_table) {
+    foreach ($membership_table->getColumnsHash() as $values) {
+      // Load group.
+      $group = $this->getSolutionByName($values['solution']);
```
idimopoulos: Again the same with the collection. This whole thing can be moved to a trait.

idimopoulos: Also, the getSolutionByName ends up in getRdfEntityByLabel which already handles the possibility of non existent item. You do not need to check if the group is not found.

tests/features/solution/solution.edit.feature
brummbar: The ticket specifically says solution owner and solution facilitator, so the two tests are meant to be different. Giving the fact that we are assigning programmatically the permissions, it seems indeed an overlap, but not from an UAT point of view, where the solution will be created manually.

tests/features/solution/solution.edit.feature
brummbar: It's a wrong copy-paste, it should say "solution owner or solution facilitator".

web/modules/custom/owner/owner.behat.inc
brummbar: Indeed, I don't know why it didn't sound strange to me.

tests/features/solution/solution.edit.feature
idimopoulos: Still, if we let the client test this twice, we will give them the impression that there is a difference between owning a collection and being a facilitator when it comes to editing a solution. I really don't want to mess the permissions.
I mean, you are already testing as a facilitator in the first test. 
Just add a comment like 
# A solution owner can edit a solution because he is also a facilitator. 
Or something like that.

PR #136, opened 2016-07-26
ISAICP-2421

idimopoulos:


web/modules/custom/asset_release/src/Controller/AssetReleaseController.php
```diff
@@ -73,10 +99,7 @@ public function add(RdfInterface $rdf_entity) {
    *   The access result object.
    */
   public function createAssetReleaseAccess(RdfInterface $rdf_entity) {
-    $user = $this->currentUser();
-    $membership = Og::getMembership($user, $rdf_entity);
-    // @todo: Remove check for empty after ISAICP-2369 is in.
-    return (!empty($membership) && $membership->hasPermission('create asset_release rdf_entity')) ? AccessResult::allowed() : AccessResult::forbidden();
+    return $this->ogAccess->userAccessEntity('create', $this->createNewAssetRelease($rdf_entity), $this->currentUser());
```
brummbar: I don't like this. You are creating a new temporary asset release with the group set to the solution to check the permissions, and then discarding it straight away, since a new fresh asset release is going to be created to be used in the form, where again the group is set (plus other values).
At this point I think we should simply use something like this:
`return $this->ogAccess->userAccess($rdf_entity, 'create asset_release rdf_entity')`
which avoids temporary entities.

web/modules/custom/asset_release/src/Controller/AssetReleaseController.php
```diff
@@ -17,6 +17,32 @@
...
 
+  /**
+   * The OG access handler.
+   *
+   * @var \Drupal\og\OgAccessInterface
+   */
+  protected $ogAccess;
+
+  /**
+   * Constructs a CustomPageController.
```
brummbar: Wrong copy-paste I guess.

web/modules/custom/asset_release/src/Controller/AssetReleaseController.php
idimopoulos: Well, the only think that this thing is actually taking care of is mapping the proper CRUD operation automatically without needing to hardcode it. Also, it's a simple creation of an entity so I don't think that it's that much of an issue.

brummbar: Permissions are written like this in Drupal. We are handling the creation of an asset release rdf entity, so it's not hardcoding from my point of view. It would be fine if you point me to something in core that uses this approach for new entities.

pfrenssen: This is how it is intended to be done in OG at the moment. OG doesn't differentiate between the create operation and operations on existing group content entities, such as 'edit' and 'delete'. 

The reason for this is that `hook_entity_create_access()` is not usable for group content in OG since it doesn't pass the group context. The standard routes that Drupal uses for creating group content (like `node/add`) don't contain any information about the group to which the group content belongs. We need to use custom routes for this (like `rdf_entity/{rdf_entity}/asset_release/add`) and hence a custom access check. In the future similar routes will be added to `og_ui` but we won't use that in this project.

Inside OG the access checking on CRUD operations for group content is handled by `OgAccess::userAccessGroupContentEntityOperations()` which accepts only a populated group content entity like @idimopoulos is doing here.

Note that you cannot use `OgAccess::userAccess()` for this. This is intended to check a user's access to operations on groups. It doesn't work with group content.

You _can_ use `OgAccess::userAccessEntity()` (as is currently done here) but it's a waste of cycles. This is simply a wrapper that is an entry point for `hook_entity_access()`. It does some checks on the entity to determine whether it is a group or group content, and then delegates to either `OgAccess::userAccess()` or `OgAccess::userAccessGroupContentEntityOperations()` .

I do like the suggestion that it should not be necessary for us as the API consumers to create this entity manually. OG should provide a convenience method `OgAccess::userAccessGroupContentEntityCreate(EntityInterface $group, string $group_content_type, string $group_content_bundle)` that can give us the desired result.

@brummbar if you want you can create an issue in https://github.com/amitaibu/og/issues to propose this (or I can do it for you).

For this particular hunk, I would suggest to keep the current approach but change `$this->ogAccess->userAccessEntity()` to `$this->ogAccess->userAccessGroupContentEntityOperations()`.

brummbar: > Note that you cannot use OgAccess::userAccess() for this. This is intended to check a user's access to operations on groups. It doesn't work with group content.

This is exactly what we are trying to do here: we are checking if inside the {rdf_entity} parameter aka solution, which is the group, we have the permission to create a release, which is the group content entity.
We don't yet have any group content, so we cannot say that we are working with group content.
This is also how it is done in og for Drupal 7, for example to render node create links.

I agree that (at least for now, depending on the og_context future implementation) we cannot use `hook_entity_create_access` because of missing context.
Still, I can see that the old `node_access()` (that naturally became an `entity_access`) has been split in an `entity_create_access`, which takes a bundle as parameter, and an `entity_access`, which takes the entity (this is a simplification of the actual implementation but we can get the picture).
This clearly defines for me the difference of permission checking during a creation and any other ~~C~~RUD operation on entities.

So that's why I still think we should check the permission on the group and not the permission on the group content.

pfrenssen: @brummbar the reason this is done like this in core is because they have no other way. The permissions are not machine readable but only human readable. You cannot look at a permission and know what kind of operation is done, and on which entity. For example if you give a permission like "edit news content" to core, it has no way to figure out that the operation is "update" and the entity type is "node".

This is solved in OG, we have actual permission classes that contain all the metadata you need (see the `GroupContentOperationPermission` class). So we don't need to revert to the inferior way that this is done in core, by hard coding a human readable string like "create news content".

But I did a great effort in ensuring that it retains compatibility with the human readable permission strings in core, so your suggestion would actually work but it is not how it is intended. The reason behind the backwards compatibility is mainly for DX reasons (people are familiar with the permission strings) and to make migration from D7 to D8 easier. Internally OG will never use `OgAccess::userAccess()` for group content CRUD operations, it will always use `OgAccess::userAccessGroupContentEntityOperations()` because it is specifically designed for this use case. Create is also a CRUD operation.

brummbar: I see that indeed `OgAccess::userAccessGroupContentEntityOperations()` is not going through `OgAccess::userAccess()`.
@pfrenssen, do you know then if the documentation of `OgAccessInterface::userAccess()` should be updated?
>All permission checks in OG should go through this function. This way we guarantee consistent behavior, and ensure that the superuser and group administrators can perform all actions.

Related to ticket: since this is the implementation, I will have to accept it, but I would have more liked the idea of using one of these `GroupContentOperationPermission` classes to check the permissions.
Something like `someclass::getGroupContentOperatioPermission($permission, $entity_bundle)` that would return a proper `GroupContentOperationPermission` class.

PR #138, opened 2016-07-27
ISAICP-2422

idimopoulos:


tests/features/asset_release/asset_release.edit.feature
```diff
@@ -0,0 +1,62 @@
...
+    When I am logged in as a "moderator"
+    And I go to the homepage of the "My awesome solution abc v1" asset release
+    Then I should not see the link "Edit" in the "Entity actions" region
+
+  Scenario: Edit a release as a solution facilitator.
+    When I am logged in as a "facilitator" of the "My awesome solution abc" solution
+    And I go to the homepage of the "My awesome solution abc v1" asset release
+    And I click "Edit"
+    Then I should see the heading "Edit Asset release My awesome solution abc v1"
+    And the following fields should be present "Name, Release number, Release notes"
```
brummbar: Either here or in the add form (being the same form) we should test all the fields that (at the moment) should be displayed: documentation, spatial coverage, keyword, status, language.

idimopoulos: You have implemented this for the add release form. I removed this check in all as there is no need as long as we are having the same form.

PR #139, opened 2016-07-27
Isaicp 2650

pfrenssen:


tests/features/custom_page/add_custom_page.feature
```diff
@@ -41,3 +46,27 @@ Feature: "Add custom page" visibility options.
     # Check that the collection content such as the 'Join collection block' is
     # available in context of the custom page.
     Then I should see the link "Leave this collection"
+
+    # I should not be able to add a custom page to a different collection
+    When I go to the homepage of the "Code Camp" collection
+    Then I should not see the link "Add custom page"
```
idimopoulos: Just a question, the comment says "I should not be able to add a custom page to a different collection".
The button of "Add custom page" does not have this meaning (we have seen that there were cases that buttons appeared when they shouldn't or vice versa).

But anyway, my question is, do we need this tested in behat since it's already covered in Og tests? I mean the permission overlapping between groups.

web/modules/custom/joinup_news/src/Controller/NewsController.php
```diff
@@ -64,14 +64,25 @@ public function add(RdfInterface $rdf_entity) {
   /**
    * Handles access to the news add form through rdf entity pages.
    *
+   * Access is granted to moderators and group members that have the permission
+   * to create news articles inside of their group, which in practice means this
+   * is granted to collection and solution facilitators.
```
idimopoulos: For the collection, under certain circumstances, members can also create news. It depends on the fields 'Closed' and 'eLibrary creation'.

web/modules/custom/joinup_news/src/Controller/NewsController.php
```diff
@@ -64,14 +64,25 @@ public function add(RdfInterface $rdf_entity) {
...
    *   The RDF entity for which the news entity is created.
    *
    * @return \Drupal\Core\Access\AccessResult
    *   The access result object.
    */
   public function createNewsAccess(RdfInterface $rdf_entity) {
-    return $this->ogAccess->userAccessGroupContentEntityOperations('create', $rdf_entity, $this->createNewsEntity($rdf_entity), $this->currentUser());
+    $user = $this->currentUser();
+    // Grant access if the user is a moderator.
+    if (in_array('moderator', $user->getRoles())) {
```
idimopoulos: I know the ticket is about restoring the create permissions to the moderator but is this the proper way to go? I mean hardcoding the moderator role in the permission handler of the custom controller?
Also, this seems like a very temporary workaround to me. The very next thing that we will have to fix, will be the 'edit' permission. And this does not have a custom controller.

Why not implement something like node_strict_access of Og but for entities? then we will have global coverage for all entities.
In any case, global permission management should not be handled in the controller. It should be generic. Otherwise we will be hunting ghosts until the end of a project and we are going to be filled with workarounds like this.

tests/features/custom_page/add_custom_page.feature
pfrenssen: This is a description of a business objective. A facilitator should only be able to add custom pages to the collection they are a facilitator of, not to other collections.

web/modules/custom/joinup_news/src/Controller/NewsController.php
pfrenssen: Good point, I didn't think of this. I'll add it to the documentation.

tests/features/custom_page/add_custom_page.feature
pfrenssen: > The button of "Add custom page" does not have this meaning (we have seen that there were cases that buttons appeared when they shouldn't or vice versa).

I don't understand this, what else would it mean? Can you show me the user scenario that describes when this button is used for a different purpose?

web/modules/custom/joinup_news/src/Controller/NewsController.php
pfrenssen: In this case this is how it should be done in the current implementation of OG. `node_access_strict` doesn't work for creation of group content since `og_entity_access()` doesn't fire for it. Instead `og_entity_access_create()` fires but this doesn't have the correct group context and cannot decide whether a user has the rights to create group content because it doesn't know the groups.

I discussed this with @idimopoulos and we have created an issue in OG to deal with it: https://github.com/amitaibu/og/issues/272

I'll add links here and create a followup.

web/modules/custom/custom_page/src/Controller/CustomPageController.php
```diff
@@ -66,14 +67,25 @@ public function add(RdfInterface $rdf_entity) {
   /**
    * Handles access to the custom page add form through collection pages.
    *
+   * Access is granted to moderators and group members that have the permission
+   * to create custom pages inside of their group, which in practice means this
+   * is granted to collection and solution facilitators.
```
idimopoulos: Here as well, collection members (not facilitators) might be able to add news. It depends on collection's settings.

idimopoulos: I think we do not have custom pages for solutions.

web/modules/custom/joinup_news/src/Controller/NewsController.php
```diff
@@ -62,7 +62,18 @@ public function add(RdfInterface $rdf_entity) {
   }
 
   /**
-   * Handles access to the news add form through rdf entity pages.
+   * Handles access to the news add form through RDF entity pages.
+   *
+   * Access is granted to moderators and group members that have the permission
+   * to create news articles inside of their group, which in practice means this
+   * is granted to collection and solution facilitators.
```
idimopoulos: It can also be collection simple members if the settings allow it.

idimopoulos: This is marked in a todo in the next line so it's ok.

web/modules/custom/custom_page/src/Controller/CustomPageController.php
idimopoulos: I thought wrong :/

PR #145, opened 2016-08-02
Isaicp 2662

idimopoulos:


tests/features/collection/api.feature
```diff
@@ -17,3 +18,31 @@ Feature: Collection API
     Given the following collection:
       | title | EU Interoperability Support Group |
     Then I should have 1 collection
+
+  Scenario: Assign OG ownership during creation of collections through UI
```
brummbar: Shouldn't we remove the OG?

tests/features/collection/api.feature
```diff
@@ -17,3 +18,31 @@ Feature: Collection API
...
+    Given the following person:
+      | name | Person collection API example |
```
brummbar: I would like to use better names. These long composed names are hard to understand, a simple real name would reflect more its meaning.

tests/features/collection/api.feature
```diff
@@ -17,3 +18,31 @@ Feature: Collection API
...
+    And users:
+      | name                | pass |
```
brummbar: Single user here, and password is not needed, usually we use mail.

> And user
  | name | Daryl Sharman |
  | mail   | daryl.sharman@example.com |
I generated a random name on the internet.

tests/features/solution/api.feature
```diff
@@ -36,4 +37,45 @@ Feature: Solution API
       | closed            | yes                         |
       | elibrary creation | facilitators                |
       | affiliates        | My first solution mandatory |
-    Then I should have 1 solution
\ No newline at end of file
+    Then I should have 1 solution
+
+  Scenario: Assign OG ownership during creation of solutions through UI
```
brummbar: Same here, as above.

tests/features/solution/api.feature
```diff
@@ -36,4 +37,45 @@ Feature: Solution API
...
+    Given the following organisation:
+      | name | Organisation solution API example |
```
brummbar: Same as above, better naming maybe? :)

tests/src/Traits/OgTrait.php
```diff
@@ -70,4 +72,33 @@ protected function convertOgRoleNamesToIds(array $roles, EntityInterface $entity
     return $roles;
   }
 
+  /**
+   * Asserts the ownership of a user to a given group.
+   *
+   * An ownership is defined as having a specific collection of roles in that
```
brummbar: I don't like the word collection. Maybe "set of roles" or "having specific roles in that..."?

tests/src/Traits/OgTrait.php
```diff
@@ -70,4 +72,33 @@ protected function convertOgRoleNamesToIds(array $roles, EntityInterface $entity
...
+   * group.
+   *
+   * @param AccountInterface $user
+   *    The user to be checked.
+   * @param RdfInterface $group
+   *    The group entity. In this project, only rdf entities are groups.
```
brummbar: Uhm ok, but using EntityInterface wouldn't have changed much, no?

tests/src/Traits/OgTrait.php
```diff
@@ -70,4 +72,33 @@ protected function convertOgRoleNamesToIds(array $roles, EntityInterface $entity
     return $roles;
   }
 
+  /**
+   * Asserts the ownership of a user to a given group.
```
brummbar: I would rephrase, something like "Asserts ownership of a group by an user".

web/modules/custom/solution/solution.behat.inc
```diff
@@ -188,6 +189,37 @@ class SolutionSubContext extends DrupalSubContextBase implements DrupalSubContex
...
+   *    The name of the user.
+   * @param string $rdf_entity
+   *    The label of the group entity.
+   *
+   * @throws \Exception
+   *    Throws an exception when the user is not found.
+   *
+   * @Given (the user ):username should be the owner of the :rdf_entity solution
+   */
+  public function assertCollectionOwnership($username, $rdf_entity) {
```
brummbar: Parameter name should be $label, $rdf_label. $rdf_entity is misleading.

tests/src/Traits/OgTrait.php
```diff
@@ -70,4 +72,33 @@ protected function convertOgRoleNamesToIds(array $roles, EntityInterface $entity
...
+   *    An array of roles to be checked. Roles must be passed as simple names
+   *    and not as full IDs. Names will be converted accordingly to IDs.
+   *
+   * @throws \Exception
+   *    Throws exception when the user is not a member or is not an owner.
+   */
+  protected function assertOgGroupOwnership(AccountInterface $user, RdfInterface $group, $roles) {
+    $membership = Og::getMembership($user, $group);
+    if (empty($membership)) {
+      throw new \Exception("User {$user->name} is no member of the {$group->label()} group.");
```
brummbar: "is not a member" and it should be "{$user->getAccountName()}" or an exception is thrown.

tests/src/Traits/OgTrait.php
idimopoulos: Well we are using this only for rdf entities and it is project specific. So I don't think it should change here. because we are also securing the proper entity type passed in.

PR #154, opened 2016-08-05
Isaicp 2671

brummbar:


web/modules/custom/rdf_entity/src/Entity/Query/Sparql/Query.php
```diff
@@ -147,13 +158,17 @@ public function condition($property, $value = NULL, $operator = '=', $langcode =
...
         }
         return $this;
 
       case $id . '-IN':
         if ($value) {
           $ids_list = "(<" . implode(">, <", $value) . ">)";
-          $this->condition->condition('?entity', 'rdf:type', '?type');
+          if (!$this->filterAdded) {
+            $this->condition->condition('?entity', 'rdf:type', '?type');
+            $this->filterAdded = TRUE;
```
idimopoulos: Wouldn't it be much better to have one method like conditionType() and then pass the data and have a protected method that has this type and if it is not empty then do not change it. We are not throwing any exception anyway so we are expecting probably the same type. Then we could even have an array of types if needed and keep them as unique values.

idimopoulos: This is now split and will be updated in ISAICP-2705.

PR #159, opened 2016-08-16
Isaicp 2602

idimopoulos:


web/modules/custom/solution/solution.module
```diff
@@ -86,3 +88,25 @@ function solution_rdf_entity_insert(EntityInterface $entity) {
...
+ */
+function solution_rdf_entity_access(EntityInterface $entity, $operation, AccountInterface $account) {
+  if ($entity->bundle() != 'solution') {
+    return AccessResult::neutral();
+  }
+
+  if ($operation == 'edit') {
+    $allowed_transitions = $entity->get('field_is_state')->first()->getTransitions();
+    if (empty($allowed_transitions)) {
+      return AccessResult::forbidden();
```
sandervd: Could we move this check to an access check on the route?
Now the user still sees the edit tab, but gets an access denied whenever he clicks the tab.

web/modules/custom/solution/solution.workflows.yml
```diff
@@ -0,0 +1,55 @@
+# The workflow of the solution entity.
+#
+# In the following workflow we have added one 'virtual' state called new.
+# This state will be used when the entity is being created.
+# In order for the state_machine module to ensure that
```
sandervd: Nice explanation! Thanks!

web/modules/custom/solution/solution.module
idimopoulos: hook_entity_access is invoked when the Route check occurs. It shouldn't show the edit link if no permission is granted. Also, we already have tests for this. If this was shown, it would cause the features/solution/solution.moderation.feature to fail.

PR #165, opened 2016-08-24
Isaicp 2579

brummbar:


tests/src/Context/DrupalContext.php
```diff
@@ -219,4 +219,20 @@ public function assertFieldsetsPresent($fieldsets) {
...
   }
 
+  /**
+   * Click on an element by css class.
+   *
+   * @Then /^I click on element "([^"]*)"$/
+   */
+  public function iClickOn($element) {
+    $page = $this->getSession()->getPage();
+    $findName = $page->find("css", $element);
```
brummbar: Use single quotes here.

web/themes/joinup/joinup_theme.theme
```diff
@@ -8,26 +8,37 @@
...
+  // Set page title.
   $title = \Drupal::service('title_resolver')->getTitle($request, $route_match->getRouteObject());
-  $variables['title'] = $title;
 
   $route_name = $route_match->getRouteName();
   $blue_header_routes = array(
     'user.login',
     'user.register',
     'user.pass',
+    'node.add',
```
brummbar: What about sorting this alphabetically? @pfrenssen suggestion eheh

web/themes/joinup/joinup_theme.theme
```diff
@@ -8,26 +8,37 @@
...
 
 /**
  * Implements hook_preprocess_page().
  */
 function joinup_theme_preprocess_page(&$variables) {
-  // Set page title.
   $request = \Drupal::request();
   $route_match = \Drupal::routeMatch();
+
+  // Set page title.
```
brummbar: Documentation here could be improved, saying what we are trying to do with the code here. For example
`// We need to show the header in a blue variant on certain routes`

web/themes/joinup/joinup_theme.theme
```diff
@@ -8,26 +8,37 @@
...
+    'rdf_entity.rdf_add',
+    'entity.node.edit_form',
+    'collection.propose_form',
+    'rdf_entity.rdf_add_page',
+    'entity.rdf_entity.edit_form',
   );
 
   if (!empty($route_name) && in_array($route_name, $blue_header_routes)) {
     $variables['is_blue_header'] = TRUE;
+    $title = $variables['page']['content']['pagetitle'];
```
brummbar: Again here, something like
`// Move the title from the content region to {...}`.

web/themes/joinup/joinup_theme.theme
```diff
@@ -43,6 +54,34 @@ function joinup_theme_preprocess_page(&$variables) {
...
+}
+
+/**
+ * Implements hook_preprocess_page_title().
+ */
+function joinup_theme_preprocess_page_title(&$variables) {
+  $route_match = \Drupal::routeMatch();
+  $route_name = $route_match->getRouteName();
+
+  // Array contains create pages.
```
brummbar: Here also improve the documentation. Something explaining the reason of what we are doing.
`// Set a flag when we are on content creation routes in order to customise some templates.`

web/themes/joinup/joinup_theme.theme
```diff
@@ -116,23 +142,70 @@ function joinup_theme_preprocess_fieldset(&$variables) {
...
+  $route_name = $route_match->getRouteName();
+  $form_routes = array(
+    'node.add',
+    'rdf_entity.rdf_add',
+    'entity.node.edit_form',
+    'collection.propose_form',
+    'entity.rdf_entity.edit_form',
+  );
+
+  // Check if container is field element - it has only one parent.
```
brummbar: Why is this needed? In which case we have more than one parent?

web/themes/joinup/joinup_theme.theme
```diff
@@ -116,23 +142,70 @@ function joinup_theme_preprocess_fieldset(&$variables) {
...
+    $variables['is_form_wrapper'] = TRUE;
+  }
+}
+
+/**
  * Implements hook_theme_suggestions_alter().
  */
 function joinup_theme_theme_suggestions_alter(array &$suggestions, array $variables) {
-  if (isset($variables['element']) && isset($variables['element']['#type'])) {
+  if (isset($variables['element']['#type'])) {
```
brummbar: You can swap the `isset()` with `!empty()` as it will check the presence of each key.

web/themes/joinup/joinup_theme.theme
```diff
@@ -116,23 +142,70 @@ function joinup_theme_preprocess_fieldset(&$variables) {
...
+    && count($variables['element']['#parents']) == 1
+    && $variables['element']['#id'] != 'edit-actions') {
+    $variables['is_form_wrapper'] = TRUE;
+  }
+}
+
+/**
  * Implements hook_theme_suggestions_alter().
  */
 function joinup_theme_theme_suggestions_alter(array &$suggestions, array $variables) {
```
brummbar: You can provide a third parameter here, `$hook`, and replace the `$variables['theme_hook_original']` usage.

web/themes/joinup/joinup_theme.theme
```diff
@@ -116,23 +142,70 @@ function joinup_theme_preprocess_fieldset(&$variables) {
...
+      'textfield',
+      'url',
+      'password',
+      'email',
+      'entity_autocomplete',
+    );
+
     if ($variables['element']['#type'] == 'form') {
-      $original_theme_hook = $variables['theme_hook_original'];
       $suggestions[] = $original_theme_hook . '__' . str_replace('-', '_', $variables['element']['#id']);
```
brummbar: There is an issue here. The html id is being used, but Drupal appends a counter to it in some cases (ajax, multiple forms in same page). The safe string to use is '#form_id', which the actually id is based on (e.g. 'rdf_entity_collection_propose_form').

web/themes/joinup/joinup_theme.theme
```diff
@@ -116,23 +142,70 @@ function joinup_theme_preprocess_fieldset(&$variables) {
...
       $suggestions[] = $original_theme_hook . '__' . str_replace('-', '_', $variables['element']['#type']);
     }
 
     $slider_ids = array(
       'edit-field-ar-elibrary-creation',
-      'edit-field-test-lista',
-      'edit-field-test-luist-normal',
     );
 
     if ($variables['theme_hook_original'] == 'select' && in_array($variables['element']['#id'], $slider_ids)) {
```
brummbar: All this code could be split off in single functions implementing `hook_theme_suggestions_HOOK_alter()`. Only the floating part would stay in here (more or less).

web/themes/joinup/js/script.js
```diff
@@ -3,22 +3,48 @@
...
+  Drupal.behaviors.silderSelect = {
+    attach: function (context, settings) {
+      $(context).find('.slider__select').once('sliderSelect').each(function () {
+         var select = $(this);
+         var selectLength = $(this).find('option').length;
 
-   $.each($('.slider__select'), function (index, value) {
-     var select = $(this);
-     var selectLength = $(this).find('option').length;
+         var slider = $("<div id='slider' class='slider__slider'></div>").insertAfter(select).slider({
```
brummbar: Wrong tab size here (3 instead of 2).

web/themes/joinup/js/script.js
```diff
@@ -3,22 +3,48 @@
...
+        $(this).addClass('button--default button--blue mdl-button mdl-js-button mdl-button--raised mdl-js-ripple-effect mdl-button--accent');
+      });
+    }
+  };
+
+  Drupal.behaviors.silderSelect = {
+    attach: function (context, settings) {
+      $(context).find('.slider__select').once('sliderSelect').each(function () {
+         var select = $(this);
+         var selectLength = $(this).find('option').length;
```
brummbar: Instead of `$(this)` use `select` variable to avoid creating again another jQuery object.
If possible and we all agree, I like more to put a dollar sign in front of variables that are jQuery objects. So it's easy to understand what is inside.

web/themes/joinup/js/script.js
```diff
@@ -3,22 +3,48 @@
...
-        max: selectLength,
-        range: "min",
-        value: select[ 0 ].selectedIndex + 1,
-        change: function (event, ui) {
-          select.find('option').removeAttr('selected');
-          $(select.find('option')[ui.value - 1]).attr('selected', 'selected');
-        }
+  Drupal.behaviors.ajaxReload = {
+    attach: function (context, settings) {
+      $(document).ajaxComplete(function (event, xhr, settings) {
```
brummbar: There is no `.once()` here so this event will be appended multiple times.

web/themes/joinup/templates/form/form-element--checkbox.html.twig
```diff
@@ -82,13 +74,13 @@
       {{ description.content }}
     </div>
   {% endif %}
-  {{ children }}
+  <label class="mdl-checkbox mdl-js-checkbox mdl-js-ripple-effect">
```
brummbar: Why in the password twig file there is a checkbox class?

web/themes/joinup/templates/form/input--checkbox.html.twig
```diff
@@ -10,4 +10,4 @@
  * @see template_preprocess_input()
  */
 #}
-<input{{ attributes.addClass('mdl-checkbox__input') }} />{{ children }}
+<input{{ attributes.addClass('mdl-checkbox__input') }} />
```
brummbar: Why {{ children }} has been left out?

web/themes/joinup/joinup_theme.theme
tom-waat: There is more than one parent for example in wysiwyg editor elements. "Text format" dropdown and "About text formats" link have also form_wrapper class, so I needed different class for form field.

web/themes/joinup/joinup_theme.theme
tom-waat: The #form_id is not accessible in variables in theme_suggestions_alter() function. But I think that we can use #id there (these elements are not duplicated or rebuilded by ajax).

web/themes/joinup/templates/form/form-element--checkbox.html.twig
tom-waat: Hmm, I do not understand. It is not password template but form-element template.

web/themes/joinup/templates/form/input--checkbox.html.twig
tom-waat: The children variable is empty for checkbox elements, so I think that it can be removed.

tom-waat:
I fixed most fixed issues mentioned in the comments above. I added comments where something was unclear.

web/themes/joinup/templates/form/form-element--checkbox.html.twig
brummbar: I think I was drunk. Ignore :)

web/themes/joinup/templates/form/input--checkbox.html.twig
brummbar: If it doesn't create an issue, I would leave it there then, just to be sure.

web/themes/joinup/joinup_theme.theme
brummbar: I see that you removed this part of the code. Anyway, the `form_id` will always be there, it's set in the form builder. And the id will be modelled after it.

web/themes/joinup/templates/form/input--checkbox.html.twig
tom-waat: Ok, I readded there {{ children }} variable.

web/themes/joinup/joinup_theme.theme
```diff
@@ -101,28 +128,90 @@ function joinup_theme_preprocess_fieldset(&$variables) {
...
+    $suggestions[] = 'select__slider';
+  }
+}
+
+/**
+ * Implements hook_theme_suggestions_HOOK_alter() for input.html.twig.
+ */
+function joinup_theme_theme_suggestions_input_alter(array &$suggestions, array $variables) {
+  if ($variables['element']['#type'] == 'submit') {
+    $button_value = Html::cleanCssIdentifier($variables['element']['#value'], array(' ' => '_'));
```
brummbar: Using the button value as theme suggestion is not really reliable, because in case of multilingual or label changes it's going to break. I don't see any templates committed for this suggestion, is it still needed?

tom-waat: It was needed for submit buttons, but since I moved them to another ticket (https://webgate.ec.europa.eu/CITnet/jira/browse/ISAICP-2670), they are no longer needed here. I also noticed that during my latest merge, I should not have removed form suggestion function, it is needed in develop branch. I rebuilded this function like other suggestion functions.

PR #170, opened 2016-08-28
Isaicp 2689

idimopoulos:


tests/features/joinup_event/add_event.solution.feature
```diff
@@ -0,0 +1,64 @@
...
+    And I go to the homepage of the "Ragged Tower" solution
+    Then I should not see the link "Add event"
+
+    When I am an anonymous user
+    And I go to the homepage of the "Ragged Tower" solution
+    Then I should not see the link "Add event"
+
+    When I am logged in as a "facilitator" of the "Ragged Tower" solution
+    And I go to the homepage of the "Ragged Tower" solution
+    Then I should see the link "Add event"
```
claudiu-cristea: For me the Behat test breaks here with (features/joinup_event/add_event.solution.feature:28). It cannot see the "Add event" link on the page.

idimopoulos: Are you running maybe the test with selenium? because it's not failing in my environment and it will fail with selenium as the link is css hidden until the user presses the + button.

claudiu-cristea: OK, it passes if I reinstall the site. The problem was because I forgot to import the configuration change from web/modules/custom/joinup_event/config/install/field.field.node.event.og_audience.yml.

Side note: However, we are not using a usual D8 structure. Ideally we should have a single config (sync) directory under VCS and not spread the config YAMLs all over the codebase. Then we point {{$config_directories['sync']}} in settings.php to that directory and, when a config change lands through VCS, we simply perform {{$ drush cim}} and import only the differences. With the actual structure, I wonder how we'll update the configuration on production?

PR #172, opened 2016-08-29
Isaicp 2691

idimopoulos:


tests/src/Traits/OgTrait.php
```diff
@@ -113,4 +114,46 @@ protected function getOgRoles(array $roles, EntityInterface $group) {
...
+   *   The bundle of the parent.
+   * @param string $content_type
+   *   The bundle of the content.
+   * @param string $content
+   *   The title of the content.
+   *
+   * @throws \Exception
+   *   Thrown when a event with the given title does not exist.
+   */
+  public function assertOgMembership($parent, $parent_type, $content_type, $content) {
```
brummbar: I would uniform the order of the arguments: now it's title, bundle, bundle, title. I suggest bundle, title, bundle, title.

tests/src/Traits/OgTrait.php
```diff
@@ -113,4 +114,46 @@ protected function getOgRoles(array $roles, EntityInterface $group) {
...
   }
 
+  /**
+   * Checks if the given content belongs to the given parent rdf entity.
+   *
+   * If there are multiple entities or parents with the same title, then
+   * only the first one is checked.
+   *
+   * @param string $parent
+   *   The name of the parent.
```
brummbar: I would write "the name of the parent rdf entity". Same for the others.

tests/src/Traits/OgTrait.php
```diff
@@ -113,4 +114,46 @@ protected function getOgRoles(array $roles, EntityInterface $group) {
...
+      ->getStorage('node')
+      ->loadByProperties(['title' => $content, 'type' => $content_type]);
+    /** @var \Drupal\node\NodeInterface $content */
+    $content = reset($results);
+
+    if (empty($content)) {
+      throw new \Exception("The $content_type titled '$content' was not found.");
+    }
+
+    /** @var \Drupal\og\Plugin\Field\FieldType\OgStandardReferenceItem $group */
```
brummbar: This approach works fine, but maybe it was better to use `Og::getGroups()` and check if the parent is there. In that way we would leverage the API instead of re-implementing it.

PR #194, opened 2016-09-13
Isaicp 2623

sandervd:


web/modules/custom/solution/src/EventSubscriber/WorkflowTransitionEventSubscriber.php
```diff
@@ -0,0 +1,82 @@
...
+use Drupal\message\Entity\MessageTemplate;
+use Drupal\message_notify\MessageNotifier;
+use Drupal\og\Entity\OgMembership;
+use Drupal\state_machine\Event\StateChangeEvent;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+/**
+ * Event subscribers for Organic Groups.
+ */
+class WorkflowTransitionEventSubscriber implements EventSubscriberInterface {
```
claudiu-cristea: I guess we'll have many other events to respond and perform actions. From an architectural point of view wouldn't it be better to consider a consistent way to respond to events instead of just coding subscribers? I'm thinking to Rules module. It would be more elegant and consistent if we can manage all reactions there. The current approach is correct, but I'm thinking that it would be better to have a single point where we can  react.

web/modules/custom/solution/src/EventSubscriber/WorkflowTransitionEventSubscriber.php
```diff
@@ -0,0 +1,82 @@
...
+    $message_template = MessageTemplate::load('workflow_transition');
+
+    $storage = \Drupal::entityManager()->getStorage('og_membership');
+
+    // Sent a message to all solution administrators.
+    $membership_query = $storage->getQuery();
+    $membership_query
+      ->condition('state', 'active')
+      ->condition('entity_id', $entity->id())
+      ->condition('roles', 'rdf_entity-solution-administrator');
```
claudiu-cristea: Hm. I'm not very happy with checking for a specific **role** in code. By doing this we lock ourself to that role. Instead, checking by a specific permission is a better approach. eg. "receive solution notification". Then, when we want to extend the target of notifications we only add grant that new role with this permission and we're done.

web/modules/custom/solution/src/EventSubscriber/WorkflowTransitionEventSubscriber.php
idimopoulos: If I'm not mistaking, @sandervd  and @pfrenssen  where against this idea of using the Rules module so I would not go with it at least for now.

web/modules/custom/solution/src/EventSubscriber/WorkflowTransitionEventSubscriber.php
idimopoulos: I will implement the solution permission. Since this is the first use case, I will set the permission within the solution module as the configuration templates regarding the messages are there as well. Later on, when we have more than one use cases, we can determine whether to leave these permissions separately or create generic ones.

PR #195, opened 2016-09-13
Isaicp 2731

brummbar:


web/themes/joinup/joinup_theme.theme
```diff
@@ -111,6 +111,11 @@ function joinup_theme_preprocess_block(&$variables) {
       $variables['is_view'] = TRUE;
     }
   }
+
+  // Get the ID of the block and add this to the main menu theme attributes.
```
brummbar: So the purpose of this code is to pass the block name to the menu being rendered. The documentation should state that, so it's easy to understand. Also the menu_alter() hook implemented below should reference this so we know the reason of the code.
Also, there's no need to mangle with #attributes probably. You can add another element prefixed by hash, it will be ignored and you can pick it up later on in the menu alter.

tom-waat:
Hi @brummbar ,
I fixed the documentation in the theme file.

PR #196, opened 2016-09-13
ISAICP-2672 adding description to the distribution license field

dublutz:

### DIFF
--- a/web/modules/custom/asset_distribution/config/install/field.field.rdf_entity.asset_distribution.field_ad_licence.yml
+++ b/web/modules/custom/asset_distribution/config/install/field.field.rdf_entity.asset_distribution.field_ad_licence.yml
@@ -9,7 +9,7 @@ field_name: field_ad_licence
 entity_type: rdf_entity
 bundle: asset_distribution
 label: License
-description: ''
+description: 'If you cannot find the desired license in the list above, please request a new one by <a href="/contact">contacting us</a>.'
 required: false
 translatable: false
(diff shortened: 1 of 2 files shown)


web/modules/custom/asset_distribution/config/install/field.field.rdf_entity.asset_distribution.field_ad_licence.yml
```diff
@@ -9,7 +9,7 @@ field_name: field_ad_licence
 entity_type: rdf_entity
 bundle: asset_distribution
 label: License
-description: ''
+description: 'If you cannot find the desired license in the list above, please request a new one by <a href="/contact">contacting us</a>.'
```
claudiu-cristea: This hardcoded path (`/contact`) will not work on sites where Drupal is installed in a sub-directory. As Joinup is distro, we cannot predict if such cases will exist or not. So, we need to provide the full-qualified. Fortunately the field API allows tokens. I think `[site:url]` is OK to be used as prefix to "contact". Also, let's click on the link, in the Behat test, and check if we're redirected to Contact page.

dublutz: @claudiu-cristea, Thank you for pointing this out. I added the site path to the link. 
For the moment we cannot follow the link to contact page in the tests because the page doesn't exist. I created a followup ticket for the contact form.

PR #197, opened 2016-09-13
ISAICP-2773: Extend RDF default content

claudiu-cristea:


composer.json
```diff
@@ -10,6 +10,7 @@
         "cweagans/composer-patches": "~1.0",
         "drupal/console": "~0.10",
         "drupal/core": "~8.1.0",
+        "drupal/default_content": "dev-8.x-1.x#c6a79ee2ffc71b0136d4988871618e05d8be4018",
```
sandervd: Why do we need an extra module for this?

web/modules/custom/demo_content/content/rdf_entity/asset_distribution/blue.json
```diff
@@ -0,0 +1,189 @@
+{
```
sandervd: - I don't see the added value that changing to JSON representation brings?
- I wouldn't put these fixtures in the modules themselves. Keeping everything centralized will assure an easy cleanup.
(This module should be removed from the repo befor go live imho)

claudiu-cristea: This is more practical, sorry. You don't need to code the content. If you need a new node or rdf entity., you just create it manually and then with `drush dce node {node}` you export it in JSON. No manual coding. This is the accepted standard way to provide default content in D8.

claudiu-cristea: > I wouldn't put these fixtures in the modules themselves. Keeping everything centralized will assure an easy cleanup. (This module should be removed from the repo befor go live imho)

The module can be left there uninstalled. If a new developer comes, he can build/install the site and get some dummy content just for testing and developing. I don't see why this would be wrong.

PR #208, opened 2016-10-06
Isaicp 2851

brummbar:


web/modules/custom/joinup_core/src/Plugin/Field/FieldWidget/StateMachineButtons.php
```diff
@@ -0,0 +1,192 @@
...
+
+    // Pass the label settings to the process callback.
+    $element['#use_transition_label'] = $this->getSetting('use_transition_label');
+
+    // Leave the field in place for validation purposes.
+    $element['#access'] = FALSE;
+
+    // Add a process callback to add the buttons in the form actions.
+    // @see \Drupal\content_moderation\Plugin\Field\FieldWidget\ModerationStateWidget::formElement()
+    $element['#process'][] = [get_called_class(), 'processActions'];
```
pfrenssen: Haha this doesn't exist yet in our code base, but it will be when we upgrade to Drupal 8.2 :)

web/modules/custom/joinup_core/src/Plugin/Field/FieldWidget/StateMachineButtons.php
```diff
@@ -0,0 +1,192 @@
...
+   *
+   * @param array $options
+   *   The current list of options.
+   * @param \Drupal\Core\Field\FieldItemListInterface $items
+   *   Array of default values for this field.
+   *
+   * @return array
+   *   The update options array.
+   *
+   * @todo function name kinda sucks.
```
pfrenssen: Well no I disagree, it was instantly clear to me what this function does.

web/modules/custom/joinup_core/src/Plugin/Field/FieldWidget/StateMachineButtons.php
```diff
@@ -0,0 +1,192 @@
...
+        // A transition that allows to keep the same state is available.
+        $loopback_transition = TRUE;
+      }
+    }
+
+    // If a transition to the current state is not available, prefix the state
+    // label for better readability.
+    if (!$loopback_transition) {
+      $options[$current_value] = $this->t('Save as @transition', ['@transition' => $options[$current_value]]);
+    }
```
pfrenssen: I like this idea, makes it easier to understand for the end user.

PR #210, opened 2016-10-07
Isaicp 2791

pfrenssen:


web/profiles/joinup/config/install/og.settings.yml
```diff
@@ -7,3 +7,6 @@ node_access_strict: true
 use_queue: null
 delete_orphans: true
 delete_orphans_plugin_id: simple
+group_resolvers:
+  - route_group
+  - route_group_content
```
pfrenssen: Here I have omitted the `user_access` plugin that will only put a group in context if the user has view access for it.

I think we don't need it probably, since all our content belongs to a single collection or solution, and the user should not even be able to see the content of collections they have no access to (e.g. collections that have been proposed but are not approved yet).

The drawback of adding this plugin is that by verifying if the user has view access it will attach the 'user' cache context to all of the contextual groups.

So my reasoning was that we do not gain any additional security by having this plugin active, but we do have a negative effect in having a more fragmented cache, so I decided not to include it.

I can be persuaded to include it though if anyone thinks this would add to the site's security. Thoughts welcome!

PR #213, opened 2016-10-11
Isaicp 2674 - Inline entity form components

brummbar:


web/themes/joinup/joinup_theme.theme
```diff
@@ -202,6 +209,104 @@ function joinup_theme_preprocess_field_multiple_value_form(&$variables) {
...
+      }
+    }
+  }
+}
+
+/**
+ * Implements hook_theme_suggestions_HOOK_alter() for input.html.twig.
+ */
+function joinup_theme_theme_suggestions_input_alter(array &$suggestions, array $variables) {
+  // Get input untranslated input value.
```
brummbar: This hook gets called very very often, because a lot of form elements are input. Better try to early return: since we are handling only buttons, I would use
```php
if (empty($variables['element']['#is_button'])) {
  return;
}
```

web/themes/joinup/joinup_theme.theme
```diff
@@ -202,6 +209,104 @@ function joinup_theme_preprocess_field_multiple_value_form(&$variables) {
...
+  $input_value = '';
```
brummbar: I don't think it's safe to use labels, even if using the untranslated string is a valid approach. I suggest to switch to checking the element key that was used in the original build array, and use a multi-dimension array for looking that up. Note that this is quite ok for inline entity form things, because they are prefixed by "ief_". Not super sure for the "add_more" one. This code replaces from here to the end of the function.
```php
  // We need to have different styling on some buttons. The key of this
  // multi-dimension array represents the suggestion name, while its values
  // are the build array key of the button being rendered.
  $input_suggestions = [
    'input__ief_add_another' => [
      // Generic "Add more" button for multi-value widgets.
      'add_more',
      'ief_add_existing',
      'ief_add',
      'ief_reference_save',
      'ief_add_save',
      'ief_edit_save',
    ],
    'input__ief_edit' => [
      'ief_entity_edit',
      'ief_entity_edit_save',
    ],
    'input__ief_remove' => [
      'ief_entity_remove',
      'ief_remove_confirm',
    ],
    'input__ief_cancel' => [
      'ief_entity_cancel',
      'ief_remove_cancel',
    ],
    'input__ief_cancel_circle' => [
      'ief_add_cancel',
      'ief_edit_cancel',
      'ief_reference_cancel',
    ],
  ];

  // The last key of #array_parents is the build array key used in the code.
  $element_key = end($variables['element']['#array_parents']);
  foreach ($input_suggestions as $suggestion => $names) {
    if (in_array($element_key, $names)) {
      $suggestions[] = $suggestion;
    }
  }
```

PR #215, opened 2016-10-11
Isaicp 2611

idimopoulos:


README.md
```diff
@@ -294,6 +294,63 @@ foreach ($this->solutions as $solution) {
 }
 ```
 
+### Joinup notification module
+
+The joinup notification module is a custom module that depends only on message,
+message_notify, og and state machine.
+
+The way it works is by saving the notifications in a settings file in the
```
brummbar: Wouldn't be better to say something:

> It automates notification handling by using a settings file in the...

README.md
```diff
@@ -294,6 +294,63 @@ foreach ($this->solutions as $solution) {
...
+  <entity_type_id> => [
+    <entity_bundle> => [
+      <role> => [
+        <message_id>
+      ]
+    ]
+  ]
+]
+```
+There can be multiple message ids and the roles are either site-wide or og.
```
brummbar: Expand the og to organic groups here, let's not be cheap on characters.

tests/features/solution/solution.notification.feature
```diff
@@ -15,12 +15,21 @@ Feature: Solution notification system
...
+      | The Time of the Child | The Time of the Child | logo.png | banner.jpg | The Red Search | Jody Buchanan       | proposed         |
+      | Some Scent            | Some Scent            | logo.png | banner.jpg | The Red Search | Jody Buchanan       | deletion_request |
     And the following solution user memberships:
       | solution              | user             | roles |
       | The Time of the Child | Benjamin Stevens | owner |
+      | Some Scent            | Benjamin Stevens | owner |
 
     # Test validation email.
     When the user "Cecelia Kim" changes the state of the "The Time of the Child" solution to "Validated"
+    Then an email should be sent to "Benjamin Stevens"
```
brummbar: This step doesn't really assure that the mail sent is ours. Maybe for now it's ok, but will be possible in the future to distinguish the e-mail sent?

web/modules/custom/joinup_notification/src/EventSubscriber/WorkflowTransitionEventSubscriber.php
```diff
@@ -0,0 +1,134 @@
...
+
+/**
+ * The event subscriber that handles the message notifications in joinup.
+ *
+ * @package Drupal\joinup_notification
+ */
+class WorkflowTransitionEventSubscriber implements EventSubscriberInterface {
+
+  /**
+   * Drupal\message_notify\MessageNotifier definition.
```
brummbar: This should say "The message notify sender service".

web/modules/custom/joinup_notification/src/EventSubscriber/WorkflowTransitionEventSubscriber.php
```diff
@@ -0,0 +1,134 @@
...
+   *
+   * @var \Drupal\message_notify\MessageNotifier
+   */
+  protected $messageNotifySender;
+
+  /**
+   * Drupal\Core\Entity\EntityManager definition.
```
brummbar: This should say "The entity manager service".

web/modules/custom/joinup_notification/joinup_notification.module
```diff
@@ -0,0 +1,93 @@
...
+    else {
+      $membership_query = \Drupal::service('entity.manager')->getStorage('og_membership')->getQuery()
+        ->condition('state', 'active')
+        ->condition('entity_id', $entity->id());
+      $memberships_ids = $membership_query->execute();
+      $memberships = OgMembership::loadMultiple($memberships_ids);
+      $memberships = array_filter($memberships, function ($membership) use ($role_id) {
+        $roles = $membership->getRoles();
+        $role_ids = array_keys($roles);
+        return in_array($role_id, $role_ids);
```
idimopoulos: After the latest changes in Og, the returned array is not indexed by role ids so I changed this part of the code but simply replaced it with an array map.

tests/features/solution/solution.notification.feature
idimopoulos: For now, we only have two distinct cases where emails are sent. One is in a transition change and the other one is in a deletion of a solution. There is no point in handling multiple emails right now as it is not testable.
Still, the way the emails are handled in the tests, is by saving all properties of the email so we can later test the content of the email or the id.

idimopoulos:
All remarks not answered were fixed.

PR #219, opened 2016-10-12
Isaicp 2857

tom-waat:
Merge ISAICP-2857 into develop.


web/themes/joinup/joinup_theme.libraries.yml
```diff
@@ -10,6 +10,14 @@ global-components:
   css:
     theme:
       css/style.min.css: {}
+  js:
```
brummbar: I couldn't find this script.js in the js folder. Also, if it not needed, avoid adding the whole set of dependencies, it was fixed from D7 to D8.

tom-waat: Hi, I removed dependencies. I also added script.js file.

PR #221, opened 2016-10-13
Isaicp 2255: Collection workflow

idimopoulos:


web/modules/custom/collection/collection.module
```diff
@@ -81,24 +104,75 @@ function collection_entity_insert(EntityInterface $entity) {
...
-  $bundle = $entity->bundle();
-  if (Og::isGroupContent($entity_type_id, $bundle)) {
-    foreach (Og::getGroupIds($entity) as $entity_type => $ids) {
-      foreach ($ids as $id) {
-        Cache::invalidateTags(['og_group_content:' . $id]);
-      }
-    }
+
+  $state = $entity->field_ar_state->value;
+  if (in_array($state, ['validated'])) {
```
idimopoulos: I will fix this that way but later on, when the merge actually happens, this will also need to be generic and the generic rule implies that more than one states can exist in that array.

web/modules/custom/collection/collection.module
```diff
@@ -81,24 +104,90 @@ function collection_entity_insert(EntityInterface $entity) {
...
 
 /**
+ * Implements hook_ENTITY_TYPE_access().
+ *
+ * Interrupts the permissions for updating a collection. The permission is not
+ * granted if there are no transitions allowed for the current state of the
+ * node.
+ */
+function collection_rdf_entity_access(EntityInterface $entity, $operation, AccountInterface $account) {
+  if ($entity->bundle() != 'collection' || $operation != 'view') {
```
claudiu-cristea: Hm. Now on $operation == 'edit' will exit here. I we should have here next "OR" chained conditions:
- `$entity->bundle() != 'collection'`
- `$account->isAnonymous()` (taken from `collection_entity_access()`)
- `\Drupal::currentUser()->hasPermission('administer rdf entity')`. Shouldn't this be in effect also for `view` op? If yes, you should move it here and remove it from `edit`.

web/modules/custom/collection/collection.module
```diff
@@ -20,6 +21,28 @@ use Drupal\rdf_entity\Entity\RdfEntityType;
 use Drupal\views\Views;
 
 /**
+ * Implements hook_entity_access().
+ */
+function collection_entity_access(EntityInterface $entity, $operation, AccountInterface $account) {
```
claudiu-cristea: Seems you you forgot to remove it :)

web/modules/custom/collection/collection.module
```diff
@@ -81,24 +104,90 @@ function collection_entity_insert(EntityInterface $entity) {
...
+      if (!$entity->isPublished() && $membership->hasPermission('view unpublished rdf entity')) {
+        return AccessResult::allowed();
+      }
+      if ($entity->isPublished() && $membership->hasPermission('view rdf entity')) {
+        return AccessResult::allowed();
+      }
+      break;
+
+    case 'edit':
+      if (\Drupal::currentUser()->hasPermission('administer rdf entity')) {
```
claudiu-cristea: See the above comment. It might need to be moved up.

web/modules/custom/collection/collection.module
idimopoulos: Sure, pretty nice :)

idimopoulos: Sorry for the first one.

PR #223, opened 2016-10-14
Isaicp 2884: Extend the behat step to verify the email sent to a user

idimopoulos:


tests/features/joinup_notification/solution.notification.feature
```diff
@@ -25,11 +25,11 @@ Feature: Solution notification system
 
     # Test validation email.
     When the user "Cecelia Kim" changes the state of the "The Time of the Child" solution to "Validated"
-    Then an email should be sent to "Benjamin Stevens"
+    Then an email should be sent to "Benjamin Stevens" described as "Message to the owner when a solution transits to 'Validated' state"
```
brummbar: What about using

> Then an email with the template "Message to the owner when a solution is deleted" should be sent to "Benjamin Stevens"

PR #224, opened 2016-10-16
Excel table migration ISAICP-2841

claudiu-cristea:


web/modules/custom/joinup_migrate/migrations/mapping_table.yml
```diff
@@ -0,0 +1,98 @@
...
+      'asset node reference': asset_node_reference
+      'Binary poll': advpoll_binary
+      'Blog': blog
+      'CAMSS Assessment': isa_camss_assessment
+      'Case': case_epractice
+      'Checksum': checksum
+      'Community': community
+      'Contact point': contact_point
+      'Context Help': contexthelp
+      'ContextHelp FAQ': contexthelp_faq
```
pfrenssen: This is a whole bunch of work that is needed for future migrations. Many of these types are not present in the current data set belonging to this ticket. It is not possible for me to validate the correctness of these at this time, but the data looks OK at first glance. If there are any errors here this will become clear when these content types are migrated in the future.

claudiu-cristea: Well, this is only reverse mapping "Node-type label" > "Node-type ID". This is because they stored in Excel the Node-type label but I wanted to end up in the MySQL table with the node-type ID. I just copied all from the source node_type table to be sure I'm not missing any.

PR #238, opened 2016-10-24
Isaicp 2930: Label on latest release

idimopoulos:


web/modules/custom/solution/solution.module
```diff
@@ -264,7 +265,22 @@ function solution_get_latest_release(RdfInterface $solution) {
...
+
+  if (empty($ids)) {
+    return NULL;
+  }
+
+  $releases = Rdf::loadMultiple($ids);
+  usort($releases, function ($release1, $release2) {
+    $ct1 = $release1->field_isr_creation_date->value;
+    $ct2 = $release2->field_isr_creation_date->value;
+    if (empty($ct1) || empty($ct2) || ($ct1 == $ct2)) {
```
pfrenssen: `==` has higher precedence than `||` so the brackets are not needed.

web/modules/custom/asset_release/asset_release.behat.inc
```diff
@@ -314,4 +314,32 @@ class AssetReleaseSubContext extends DrupalSubContextBase implements DrupalSubCo
...
+   * Checks that a release is marked as the latest one.
+   *
+   * This step is meant for the releases overview page within a solution.
+   *
+   * @param string $release
+   *    The name of the release.
+   *
+   * @Then the :release release should be marked as latest
+   *
+   * @throws \Exception
```
pfrenssen: Respect the tag order in the docblock. First `@param`, then `@return`, then `@throws`, then all others.

PR #240, opened 2016-10-25
Isaicp 2777

pfrenssen:


web/modules/custom/joinup_news/joinup_news.module
```diff
@@ -92,7 +92,14 @@ function joinup_news_node_presave(EntityInterface $entity) {
...
   }
-  elseif (in_array($state, ['validated', 'in_assessment', 'deletion_request'])) {
+  // @todo When a news article is in "Needs update" state its publication state
+  //   should not change. Otherwise a newly proposed news article that is
+  //   moved to "needs_update" by a moderator will suddenly become published.
+  //   This is absolutely not the intention that unpublished news articles
+  //   become published on rejection. This also needs a dedicated test to make
+  //   sure it does not regress.
+  // @see https://webgate.ec.europa.eu/CITnet/jira/browse/ISAICP-2761
+  elseif (in_array($state, ['validated', 'needs_update', 'deletion_request'])) {
```
idimopoulos: This is indeed an issue! Good catch! This issue was introduced in the first changes done in the news workflow. The initial workflow only has one transition towards the needs-update state and this is from the validated state.

In any case, sadly, there will be at least one more refactoring on this workflow... The goal is that the needs update state will only apply on non published content as follows:
- If there is only an unpublished version (e.g. draft) then it is obviously going to just edit the same version.
- If there is a published version, a new version will be created as unpublished (maybe with the drafty module). The unpublished version will be marked as "Needs update". There are also two cases there. If the specifications define that the published version remains published, then no changes are going to be made there and there will be a new draft version marked as "Needs update. On the other hand, if the specifications for the content type define that the content goes unpublished if reported, then the published version will be removed and the unpublished.

Is that fine as a plan? This will be handled in the upcoming tickets anyway.
The goal is that all content will be handled in the same way and the states have a clear relationship with the "Published state"

PR #246, opened 2016-10-27
Isaicp 2869

claudiu-cristea:
A list of known-is.sues and blockers can be found in web/modules/custom/joinup_migrate/migrations/collection.yml comments


build.properties.dist
```diff
@@ -227,7 +227,7 @@ solr.core.dir = ${solr.dir}/server/solr/${solr.core}
 # -------------------
 
 # Development modules to enable.
-drupal.modules.dev = admin_toolbar config_update field_ui menu_ui views_ui
```
idimopoulos: Why remove admin toolbar? it is not a part of this ticket and as far as I remember Pieter checked that it is compatible with his keyboard so, accessibility is ok.!

web/modules/custom/joinup_migrate/README.md
```diff
@@ -26,39 +26,55 @@ In order to start a migration we will need to do some preparation:
...
 
 3. Run the migration setup. Note that this should normally only be run once
    since it will write the migration database credentials to `settings.php`.
    Running this again will cause these credentials to be appended a second
    time which is useless.
 
-    ```
-    $ ./vendor/bin/phing setup-migration
-    ```
+```
```
idimopoulos: Maybe we can revert this. The reason I did not fix this here as well is that for an ordered list it will indeed indent the wrapper and will present it as a child of the ordered list item.
Lower in this readme, it was the same but outside the ordered list. This caused the wrapper to filter the ``` and present them as plain text. 
I think this feels better to be within the ordered list item rather than outside.

idimopoulos: You can see the difference between the versions
https://github.com/ec-europa/joinup-dev/blob/db07bc08e7b88e2f4a6334c9ed30236b1955b7f6/web/modules/custom/joinup_migrate/README.md
and
https://github.com/ec-europa/joinup-dev/tree/ISAICP-2869/web/modules/custom/joinup_migrate which is the current one.

web/modules/custom/joinup_migrate/README.md
```diff
@@ -26,39 +26,55 @@ In order to start a migration we will need to do some preparation:
...
-    migration.db.user = my_db_user
-    migration.db.password = my_db_pass
-    migration.db.import_path = /my/path/to/d6-joinup.sql
-    ```
+Note that `migration.db.host`, `migration.db.port`, `migration.db.user` and
+`migration.db.password` are defaulting to the main database (D8) values as the
+migration is performing select queries that are joining tables across the two
+databases. For this reason both, source database and the destination database,
+should live on the same server and should be accessible by the same use. The
+MySQL user used to connect to the Drupal 8 site should be granted with read-only
```
idimopoulos: "...should be granted with at least read-only permissions..."

PR #248, opened 2016-10-27
ISAICP-2886: Distribution form front end implementations.

brummbar:


web/themes/joinup/joinup_theme.theme
```diff
@@ -328,6 +328,7 @@ function joinup_theme_theme_suggestions_input_alter(array &$suggestions, array $
     'input__ief_remove' => [
```
brummbar: Since this is not anymore a inline entity form only template, I suggest to call this 'input__remove_button' and to rename its related template.

tom-waat: Ok, It is changed.

PR #263, opened 2016-11-04
2882: Merge 'organisation' & 'person' into 'owner'

claudiu-cristea:
[Hans comment from ISAICP-2882](https://webgate.ec.europa.eu/CITnet/jira/browse/ISAICP-2882?focusedCommentId=1937623&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-1937623) is relevant for this change.

What is in?
- Added a new vocabulary `owner_type`
- Added the new 'owner' RDF entity bundle with owner name & type.
- Fixed entity reference in collection, solution, asset_release.
- Fixed default content
- Fixed Behat tests
- Added owners synchronisation between solution and asset_release
- Hidden the asset_release owner field.
- Added admin_toolbar as dev module.
- Added ScreenShotTrait to easily debug Behat tests.


web/modules/custom/owner/owner.behat.inc
```diff
@@ -28,137 +29,70 @@ class OwnerSubContext extends DrupalSubContextBase implements DrupalSubContextIn
...
-    // Convert aliased field names.
-    $values = $this->convertFieldAliases('organisation', $organisation_table->getRowsHash());
-
-    $this->createOwner('organisation', $values);
-  }
+  public function givenOwner(TableNode $owners_table) {
+    $term_storage = \Drupal::entityTypeManager()->getStorage('taxonomy_term');
+    foreach ($owners_table->getHash() as $row) {
+      // If no type was specified, fallback to 'Private Individual(s)'.
+      $row['type'] = isset($row['type']) ? $row['type'] : 'Private Individual(s)';
```
idimopoulos: Is this what happens in the UI as well? why do we need to fallback since the field instance does not include a default?

tests/features/asset_release/add_asset_release.feature
```diff
@@ -38,7 +39,7 @@ Feature: "Add release" visibility options.
     And I click "Add release"
     Then I should see the heading "Add Release"
     And the following fields should be present "Name, Release number, Release notes, Documentation, Spatial coverage, Keyword, Status, Language"
-    And the following field widgets should be present "Contact information, Owner"
+    And the following field widgets should be present "Contact information"
```
claudiu-cristea: I removed the _Owner_ field from asset_release form, in this PR,  because we sync the owner from the parent solution.

tests/features/bootstrap/FeatureContext.php
```diff
@@ -18,6 +18,8 @@ class FeatureContext extends RawDrupalContext implements SnippetAcceptingContext
 
   use EntityTrait;
   use UtilityTrait;
+  // Uncomment to enable screen-shots during development.
+  // use \Drupal\joinup\Traits\ScreenShotTrait;
```
claudiu-cristea: Hm... Seems to me an over-engineering. But, yes, we can automate this. Let's not add now to much off-topic features. I needed because after owner change, a lot of Behat tests were broken and it's a nightmare to debug blindly. Let's open a follow up for an automation, if you think.

tests/features/owner/collection_with_owner.feature
```diff
@@ -19,26 +20,50 @@ Feature: Creation of owners through UI
...
+    And I select "Private Individual(s)" in the dropdown of the "Owner" field
+    And I fill in "Name" with "John Doe"
     And I press "Create owner"
+    Then I should see "John Doe"
+    And I should see the link "Private Individual(s)"
+
+    # Because the cardinality is 1, we need to delete first the created owner in
+    # order to test more cases.
+    When I press "Remove" at the "Owner" field
+    And I press "Remove" at the "Owner" field
```
claudiu-cristea: Right.

tests/src/Traits/ScreenShotTrait.php
```diff
@@ -0,0 +1,145 @@
...
+    $this->createScreenshot($file_name, $message, FALSE);
+  }
+
+  /**
+   * Saves a screen-shot under a generated name.
+   *
+   * @Then (I )take a screenshot
+   */
+  public function takeScreenshotUnnamed() {
+    $file_name = sys_get_temp_dir() . DIRECTORY_SEPARATOR . 'behat-screenshot';
```
claudiu-cristea: I really don't know. As I added in the trait docblock, this code was stolen. I didn't touch it except the class namespace. I changed the phpdoc. But I didn't used this method. You'll get automatic screenshots with `takeScreenshotAfterFailedStep()`. We can improve this code for sure but is somehow out-of-scope. I agree that adding this trait is also out-of-scope.

web/modules/custom/owner/owner.behat.inc
claudiu-cristea: No. It's only a Behat shorthand for _Private Individual(s)_.

PR #267, opened 2016-11-10
ISAICP-2866: The context provider has been accepted, revert back to the main repository

pfrenssen:


composer.json
```diff
@@ -19,7 +19,7 @@
         "drupal/message_notify": "~1.0",
         "drupal/migrate_drush": "~1.1",
         "drupal/migrate_spreadsheet": "1.x-dev",
-        "drupal/og": "dev-242-simplified",
+        "drupal/og": "dev-1.x#dbffd5d9123b2b5b0f0bc91efb3bc9d45da3f073",
```
claudiu-cristea: Why are we pinning? I think based on our policy we should let new builds come in on composer update and catch possible failures early.

claudiu-cristea: So, we're pinning dev builds.

PR #270, opened 2016-11-11
Isaicp 2945

sandervd:


web/modules/custom/rdf_entity/rdf_entity.module
```diff
@@ -69,22 +69,35 @@ function rdf_entity_form_field_storage_config_edit_form_alter(&$form, FormStateI
...
     );
   }
 
   $form['#entity_builders'][] = 'rdf_entity_form_alter_builder';
 }
 
 /**
+ * Retrieve nested third party settings from object.
+ */
+function rdf_entity_get_third_party_property(ConfigEntityInterface $object, $property, $column, $default = NULL) {
```
idimopoulos: Do you think that this could go to the mapping handler as 'get_third_party_property'?

web/modules/custom/rdf_entity/rdf_entity.module
```diff
@@ -69,22 +69,35 @@ function rdf_entity_form_field_storage_config_edit_form_alter(&$form, FormStateI
...
 function rdf_entity_form_alter_builder($entity_type, FieldStorageConfig $entity, array &$form, FormStateInterface $form_state) {
   $schema = $entity->getSchema();
+  $data = [];
   foreach ($schema['columns'] as $column => $column_desc) {
-    $value = $form_state->getValue($column);
-    $entity->setThirdPartySetting('rdf_entity', 'mapping_' . $column, $value);
+    $data[$column] = $form_state->getValue($column);
+
   }
+  $entity->setThirdPartySetting('rdf_entity', 'mapping_' . $column, $value);
```
idimopoulos: The $value here is undefined now.

web/modules/custom/rdf_entity/rdf_taxonomy/src/TermRdfStorage.php
```diff
@@ -208,7 +208,7 @@ public function loadTree($vid, $parent = 0, $max_depth = NULL, $load_entities =
       if (empty($this->treeChildren[$vid])) {
         /** @var \Drupal\taxonomy\Entity\Vocabulary $voc */
         $voc = entity_load('taxonomy_vocabulary', $vid);
-        $settings = $voc->getThirdPartySetting('rdf_entity', 'mapping_vid');
+        $settings = $voc->getThirdPartySetting('rdf_entity', 'mapping')['vid'];
```
idimopoulos: Again, how come the vocabulary has a direct call to the getThirdPartySettings and the rdf entity has to call 'rdf_entity_get_third_party_property' instead?
Can we use the service in both cases here?
We have config/content entity properties and field properties only.

web/modules/custom/rdf_entity/src/Entity/RdfEntitySparqlStorage.php
```diff
@@ -656,34 +659,64 @@ protected function doSave($id, EntityInterface $entity) {
     $target_graph = $this->getGraphHandler()->getTargetGraphFromEntity($entity);
     $graph_uri = $this->getBundleGraphUri($bundle, $target_graph);
 
-    $insert = '';
+    $graph = new Graph($graph_uri);
```
idimopoulos: I would prefer to have this as a static method, part of rdf Query class or the RdfSparqlStorage class that will return a new graph and maybe we can handle exceptions there in the future.

web/modules/custom/rdf_entity/src/Entity/RdfEntitySparqlStorage.php
```diff
@@ -656,34 +659,64 @@ protected function doSave($id, EntityInterface $entity) {
...
-          $pred = '<' . (string) $properties['by_field'][$field_name][$column] . '>';
-          if (!filter_var($value, FILTER_VALIDATE_URL) === FALSE) {
-            $obj = '<' . $value . '>';
+          // Skip the bundle as it is handled separately later.
+          if ($field_name == 'rid') {
+            continue;
+          }
+          // Don't add empty values.
+          if ($value === "") {
+            continue;
```
idimopoulos: Is there a specific reason for filtering out empty strings? When an entity is saved in the sql database, a field with an empty value will still be saved, right? 
Why wouldn't we save the empty string value? Because there is a difference between empty string and null.

idimopoulos: I mean, this is just a question to understand why are you filtering them out. By the way, if that is the case, you can array filter the field_item variable prior to looping over it.

web/modules/custom/rdf_entity/src/Entity/RdfEntitySparqlStorage.php
```diff
@@ -656,34 +659,64 @@ protected function doSave($id, EntityInterface $entity) {
...
-    // Save the bundle.
+    // Save the bundle as rdf:type.
     $rdf_bundle_mapping = $this->mappingHandler->getRdfBundleMappedUri($entity->getEntityType()->getBundleEntityType(), $entity->bundle());
     $rdf_bundle = $rdf_bundle_mapping[$entity->bundle()];
-    $insert .= $subj . ' ' . $this->rdfBundlePredicate . ' <' . $rdf_bundle . '>  .' . "\n";
+    $graph->addResource((string) $id, $this->rdfBundlePredicate, $rdf_bundle);
 
-    $query = <<<QUERY
+    if (!$entity->isNew()) {
+      $query = <<<QUERY
```
idimopoulos: Since you moved the insert query to a private method, it would be nice to get read of this piece here to another one.

web/modules/custom/rdf_entity/tests/modules/rdf_entity_test/config/install/field.field.rdf_entity.dummy.field_text.yml
```diff
@@ -0,0 +1,18 @@
...
+entity_type: rdf_entity
+bundle: dummy
+label: Text
+description: ''
+required: false
+translatable: false
+default_value: {  }
+default_value_callback: ''
+settings: {  }
+field_type: string_long
```
idimopoulos: new line.

PR #274, opened 2016-11-17
ISAICP-2369: Associate Collection Custom Pages with Collection Menus (2nd round)

brummbar:
Second round after new remarks in UAT session.


tests/features/custom_page/navigation_menu.feature
```diff
@@ -17,19 +17,15 @@ Feature: Navigation menu for custom pages
...
     And I should see the text "There are no pages yet. Why don't you start by creating an About page?"
     And I should see the link "Add a new page"
-
-    # Check that the menu form is initially empty and shows a help text.
-    When I click the contextual link "Edit menu" in the "Left sidebar" region
-    Then I should see the heading "Edit navigation menu of the Rainbow tables collection"
-    And the navigation menu of the "Rainbow tables" collection should have 0 items
-    And I should see the text "There are no custom pages yet."
+    # Check that the 'Edit menu' local action is not present.
+    But I should not see the contextual link "Edit menu" in the "Left sidebar" region
```
pfrenssen: It should be visible by default, it is not themed yet. Once ISAICP-2730 is fixed they will be always visible.

So I would not change this now, since this already describes the expected behaviour.

PR #278, opened 2016-11-21
ISAICP-2331: Deleting a comment

sandervd:


web/modules/custom/joinup_comment/joinup_comment.module
```diff
@@ -0,0 +1,62 @@
...
+    ];
+    $message = Message::create($msg);
+    $message->set('field_comment_text', $entity->get('field_body')->getValue());
+    $message->set('field_page', $entity->getCommentedEntityId());
+    $message->setOwner($entity->getOwner());
+    $langcode = \Drupal::languageManager()->getCurrentLanguage()->getId();
+    $message->setLanguage($langcode);
+    $message->save();
+    // Send the saved message as an e-mail.
+    \Drupal::getContainer()->get('message_notify.sender')->send($message, [], 'email');
```
brummbar: I am wondering, we have the joinup_notification module. Should we put this there too? :thinking:

web/modules/custom/joinup_notification/joinup_notification.behat.inc
```diff
@@ -55,6 +55,7 @@ class JoinupNotificationSubContext extends DrupalSubContextBase implements Drupa
       throw new \Exception("User {$recipient} does not have an email.");
     }
 
+    \Drupal::state()->resetCache();
```
brummbar: I see that this is needed. I have no idea why tho, you?

brummbar: I've seen that it works also if moved in the `beforeEmailScenario()`, just after setting the test mail collector.

sandervd: This is how I think it works:
The behat process is a long running process, it starts at the start of the test, end ends at the end of it.
The actual sending of the mails happens somewhere in between the process. If however, a static gets populated in the start, and read back after other process set the state...

web/modules/custom/joinup_comment/joinup_comment.module
```diff
@@ -0,0 +1,62 @@
...
+ * Implements hook_entity_delete().
+ */
+function joinup_comment_entity_delete(EntityInterface $entity) {
+  if (!$entity instanceof CommentInterface) {
+    return;
+  }
+
+  $view_builder = \Drupal::entityManager()->getViewBuilder($entity->getEntityTypeId());
+  $build = $view_builder->view($entity, 'full');
+  $html = \Drupal::service('renderer')->renderRoot($build);
```
sandervd: The idea is that this way I don't have to care about catching all data...
Anyway, I don't like this logger stuff.

web/modules/custom/joinup_comment/config/install/field.field.message.comment_deletion.field_page.yml
```diff
@@ -0,0 +1,29 @@
+langcode: en
+status: true
+dependencies:
+  config:
+    - field.storage.message.field_page
+    - message.template.comment_deletion
+id: message.comment_deletion.field_page
```
sandervd: - joinup_notification makes some assumptions on the structure of things
(comments have to lookup their embedding entity)
With some effort we can probably try to make this fit the comment use case as well...
- reference fields can only have one target type.
- This would introduce a dependency on joinup_notifications

PR #288, opened 2016-12-01
Isaicp 2860

sandervd:


web/modules/custom/collection/config/install/core.entity_form_display.rdf_entity.collection.default.yml
```diff
@@ -204,6 +210,11 @@ content:
     settings:
       use_transition_label: '1'
     third_party_settings: {  }
+  field_collection_content:
+    weight: 18
+    settings: {  }
+    third_party_settings: {  }
+    type: search_api_field_default
```
pfrenssen: The search field should probably not show up in the collection edit form.

web/modules/custom/collection/config/install/core.entity_view_display.rdf_entity.collection.default.yml
```diff
@@ -104,35 +102,48 @@ content:
...
       link: true
     third_party_settings: {  }
     type: entity_reference_label
+  field_ar_shared_content:
+    type: entity_reference_label
+    weight: 2
+    label: above
+    settings:
+      link: true
+    third_party_settings: {  }
```
pfrenssen: I don't think it is intentional to make this field show up. I'm not even sure what it is used for.

web/modules/custom/collection/config/install/core.entity_view_display.rdf_entity.collection.default.yml
```diff
@@ -141,13 +152,16 @@ content:
...
     type: string
-    weight: 1
+    weight: 0
     settings:
       link_to_entity: false
     third_party_settings: {  }
+  rdf_entity_field_collection_content_top:
+    weight: 1
+    settings: {  }
+    third_party_settings: {  }
```
pfrenssen: What is this? Is it maybe a field group or something? Is this intentional to make this appear here?

web/modules/custom/solution/solution.module
```diff
@@ -177,6 +226,34 @@ function solution_entity_extra_field_info() {
...
+      return;
+    }
+    // In ADMS-AP collections point to solutions.
+    // The reverse relation would have been more logical,
+    // and this is quite inconvenient, especially for the search index.
+    // To work around this, we add a computed field here.
+    $ids = \Drupal::entityQuery('rdf_entity')
+      ->condition('rid', 'collection')
+      ->condition('field_ar_affiliates', $entity->id())
+      ->execute();
```
pfrenssen: I'm a bit concerned about the performance impact here. This does an additional query every time a solution is loaded, regardless of whether this data is even needed.

Maybe make a follow up issue to turn this into an entity reference field + a `hook_entity_update()` implementation that keeps the field updates whenever a reverse reference changes?

web/modules/custom/solution/solution.module
```diff
@@ -177,6 +226,34 @@ function solution_entity_extra_field_info() {
...
+    // This will only be set when submitting a new solution form.
+    $pending_collection_id = solution_relate_collection_id();
```
pfrenssen: Hmm some unnameable devilry going on here :grimacing:

PR #289, opened 2016-12-05
Isaicp 2794

sandervd:


web/modules/custom/custom_page/custom_page.behat.inc
```diff
@@ -257,6 +258,80 @@ class CustomPageSubContext extends DrupalSubContextBase implements DrupalSubCont
...
+    if ($element) {
+      throw new \Exception(sprintf("The link '%s' was present in the navigation menu on the page %s and was not supposed to be", $link, $this->getSession()->getCurrentUrl()));
+    }
+  }
+
+  /**
+   * Returns the DOM element containing the navigation menu.
+   *
+   * @return \Behat\Mink\Element\NodeElement
+   *   The DOM element containing the navigation menu.
```
sandervd: Could this be a simple behat region?
Then we wouldn't need specific checks.

PR #285, opened 2016-11-29
Isaicp 2973

claudiu-cristea:
Contains:

- Migrations `solution`, `solution_logo`, `solution_banner`.
- Introducing `migrate_run`.
- Tagging migrations.
- Overriding cache tags for `rdf_entity`.
- Fixing `{search_api_item}`, column `item_id` width. @see https://www.drupal.org/node/2829696
- Creating a kernel base test.
- Restructuring migration classes by providing base classes and traits.
- Fix SPARQL query. Use `update()` on updates. @see https://webgate.ec.europa.eu/CITnet/jira/browse/ISAICP-3014

PR #291, opened 2016-12-05
2979: Migrate owners

claudiu-cristea:
Blocked by #285


claudiu-cristea:
This change includes ISAICP-2973 and ISAICP-2979.

PR #292, opened 2016-12-06
ISAICP-3030: Include fixes from #2789315 and #2756401

claudiu-cristea:
https://www.drupal.org/node/2789315 has a prerequisite in https://www.drupal.org/node/2810381 and, sequentially, the patches aren't apllying to 8.2.x, so I had to combine them in a local patch.


web/modules/custom/rdf_entity/src/Entity/Rdf.php
```diff
@@ -300,7 +300,8 @@ public function getWeight() {
    * {@inheritdoc}
    */
   public function isPublished() {
-    $storage = $this->entityManager()->getStorage($this->getEntityTypeId());
+    /** @var \Drupal\rdf_entity\Entity\RdfEntitySparqlStorage $storage */
+    $storage = $this->entityTypeManager()->getStorage($this->getEntityTypeId());
```
pfrenssen: :+1:

pfrenssen:
Thanks!

We'll have to keep maintaining this patch for a while though. The first beta of 8.3.x is only planned for the end of February 2017, and the final release for the end of April. It's a bit too early to hop on this branch already.

PR #294, opened 2016-12-07
Isaicp 2928

sandervd:


web/modules/custom/state_machine_revisions/src/ParamConverter/EntityRevisionConverter.php
```diff
@@ -0,0 +1,93 @@
...
+      $latest_revision = $this->revisionManager->loadLatestRevision($entity);
+
+      // If the entity type is translatable, ensure we return the proper
+      // translation object for the current context.
+      // @see parent::convert()
+      if ($latest_revision instanceof EntityInterface && $entity instanceof TranslatableInterface) {
+        $latest_revision = $this->entityManager->getTranslationFromContext($latest_revision, NULL, array('operation' => 'entity_upcast'));
+      }
+
+      if ($latest_revision->isRevisionTranslationAffected()) {
```
pfrenssen: This is very interesting, but there is no test coverage for this currently since our project is only in English. I guess this is fine for now since it is a bit pointless to write integration tests for functionality that we do not support. Still if we plan to contribute this module in the future we should probably write a proper test suite for it.

web/modules/custom/state_machine_revisions/src/EventSubscriber/WorkflowTransitionEventSubscriber.php
```diff
@@ -0,0 +1,125 @@
...
+      return;
+    }
+
+    // Verify if the new state is marked as published state.
+    $is_published_state = $this->isPublishedState($event->getToState(), $event->getWorkflow());
+    // Set revision as default when:
+    // - the entity is new;
+    // - the new state is a published one;
+    // - the current default revision is not published.
+    $entity->isDefaultRevision($entity->isNew() || $is_published_state || !$this->hasPublishedDefaultRevision($entity));
```
pfrenssen: Every time I see this method call it creeps me out. https://www.drupal.org/node/2706337

web/modules/custom/state_machine_revisions/src/RevisionManager.php
```diff
@@ -0,0 +1,102 @@
...
+
+namespace Drupal\state_machine_revisions;
+
+use Drupal\Core\Entity\ContentEntityInterface;
+use Drupal\Core\Entity\EntityTypeInterface;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\Core\TypedData\TranslatableInterface;
+
+/**
+ * Manages stuff.
```
pfrenssen: :grin:

web/profiles/joinup/joinup.profile
```diff
@@ -198,6 +198,16 @@ function joinup_inline_entity_form_reference_form_alter(&$reference_form, &$form
 }
 
 /**
+ * Implements hook_form_alter().
+ *
+ * Disable access to the revision information vertical tab.
+ * This prevents access to the revision log and the revision checkbox too.
+ */
+function joinup_workflows_form_node_form_alter(&$form, FormStateInterface $form_state, $form_id) {
+  $form['revision_information']['#access'] = FALSE;
```
pfrenssen: I was looking in adding a test for this and it seems this code is never executed. This appears to be an implementation of `hook_form_FORM_ID_alter()` but there is no module named `joinup_workflows` so this function name is never considered as a valid hook implementation.

I have checked in the interface and the revision information is still visible when e.g. creating a news article.

So yeah let's add a test, and fix it too :)

PR #296, opened 2016-12-08
Isaicp 2261 - Asset release workflow

idimopoulos:


web/modules/custom/asset_release/tests/src/Kernel/AssetReleaseWorkflowTest.php
```diff
@@ -0,0 +1,385 @@
...
+      $content = Rdf::create([
+        'rid' => 'asset_release',
+        'field_isr_is_version_of' => $parent->id(),
+      ]);
+      foreach ($test_data_arrays as $test_data) {
+        $operation = 'create';
+        $user_var = $test_data[0];
+        $expected_result = $test_data[1];
+
+        $access = $this->ogAccess->userAccessEntity('create', $content, $this->{$user_var})->isAllowed();
```
brummbar: Discussed this with @pfrenssen: use the global entity access instead of directly using og access.

web/modules/custom/asset_release/asset_release.module
```diff
@@ -91,3 +94,84 @@ function asset_release_theme($existing, $type, $theme, $path) {
...
 }
+
+/**
+ * Implements hook_ENTITY_TYPE_access().
+ *
+ * Interrupts the permissions for updating a release. The permission is not
+ * granted if there are no transitions allowed for the current state of the
+ * node.
+ */
+function asset_release_rdf_entity_access(EntityInterface $entity, $operation, AccountInterface $account) {
```
idimopoulos: Indeed! We have already discussed about this. 
The conclusion of the discussions was that since we have a lot of workflows, we cannot be sure about special needs until we have at least a lot of them implemented.
After we have these, if we are sure that we can have a generic implementation for all entities, we can convert all of them in a generic one but also we should not force it to happen as it is about very few lines of code and does not make this urgent or in need of change compared to the possible risk of special handling of an entity.

web/modules/custom/asset_release/asset_release.module
```diff
@@ -91,3 +94,84 @@ function asset_release_theme($existing, $type, $theme, $path) {
...
+  }
+  return AccessResult::neutral();
+}
+
+/**
+ * Implements hook_ENTITY_TYPE_presave().
+ *
+ * Default graph is published. Draft graph is draft.
+ */
+function asset_release_rdf_entity_presave(EntityInterface $entity) {
```
idimopoulos: If I fix this now, I will have to take care of the backwards compatibility (backport it to collections solutions and news). Is this acceptable.?

web/modules/custom/asset_release/tests/src/Kernel/AssetReleaseWorkflowTest.php
idimopoulos: Probably I missexplained the issue in the previous call. I do not think that the og is meant to handle the entity create access by hooking into that. The issue is simple:
The parent entity is not passed in the entity access controller and we have no way of knowing it.

The call stack is:
EntityAccessControllHandler::createAccess invokes hook_entity_create_access
og_entity_create_access

The only context we have here is the entity type id. We cannot rely on a context to determine the parent or the field itself as it is a new entity.
This is why we went with directly calling the ogAccess service.
The og_entity_create_access offers some help and convenience mainly for node.
@See og.module::og_entity_create_access.

tests/features/asset_release/asset_release.moderation.feature
```diff
@@ -0,0 +1,63 @@
...
+    And the following contact:
+      | name  | SheriMoore              |
+      | email | SheriMoore @example.com |
+    And users:
+      | name            | roles     |
+      | Bonnie Holloway |           |
+    And the following solutions:
+      | title     | description | logo     | banner     | owner       | contact information | state     |
+      | Dark Ship | Dark ship   | logo.png | banner.jpg | Kenny Logan | SheriMoore          | validated |
+    And the following solution user memberships:
```
idimopoulos: This is anyway connecting multiple tables so I think it's proper to use the column style only.
I have updated the term to "Given the following solution user membership" by making the 's' optional.This is anyway connecting multiple tables so I think it's proper to use the column style only.
I have updated the term to "Given the following solution user membership" by making the 's' optional.

web/modules/custom/asset_release/src/Guard/AssetReleaseFulfillmentGuard.php
```diff
@@ -0,0 +1,111 @@
...
+   * Instantiates a AssetReleaseFulfillmentGuard service.
+   *
+   * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entity_type_manager
+   *    The WorkflowUserProvider service.
+   * @param \Drupal\joinup_user\WorkflowUserProvider $workflow_user_provider
+   *    The WorkflowUserProvider service.
+   * @param \Drupal\asset_release\AssetReleaseRelations $asset_release_relations
+   *    The Asset release relation service.
+   */
+  public function __construct(EntityTypeManagerInterface $entity_type_manager, WorkflowUserProvider $workflow_user_provider, AssetReleaseRelations $asset_release_relations) {
```
idimopoulos: Same as below but that is a good suggestion indeed and already considered.
Though, can we skip it for the next workflow? the reason is that this ticket is already too much loaded.

tests/features/asset_release/asset_release.moderation.feature
brummbar: If I understood correctly, this is anyway connecting multiple tables so I think it's proper to use the column style only and you have updated the term to "Given the following solution user membership" by making the 's' optional. Right? 😆

web/modules/custom/asset_release/asset_release.module
brummbar: Let's do this in the next workflow ticket as this one is already too big.

PR #318, opened 2016-12-23
Isaicp 3056

pfrenssen:


web/profiles/joinup/config/install/views.view.collections.yml
```diff
@@ -486,8 +482,9 @@ display:
       relationships: {  }
       arguments: {  }
       display_extenders: {  }
+      show_admin_links: false
```
pfrenssen: I also turned off the contextual links while I was at it, this added an additional cache context to the render array.

web/profiles/joinup/joinup.install
```diff
@@ -24,4 +24,10 @@ function joinup_install() {
     ->getEditable('user.settings')
     ->set('register', USER_REGISTER_VISITORS)
     ->save(TRUE);
+
+  // Disable the warning about rebuilding node access. This is needlessly set
+  // in the install hook of the View Unpublished module.
+  // @todo Remove this once the bug is fixed.
+  // @see https://www.drupal.org/node/2837685
+  node_access_needs_rebuild(FALSE);
```
pfrenssen: Discovered while debugging! Finally fixed this annoying error message :)

build.properties
```diff
@@ -94,7 +94,7 @@ phpunit.sparql_url = sparql://${sparql.host}:${sparql.port}/
 # -------------------------
 
 # Apache Solr version to use.
-solr.version = 6.2.1
+solr.version = 6.3.0
```
idimopoulos: Cool that we updated, but why is the previous broken? is it because it is removed after a newer version is being released?

web/modules/custom/collection/collection.module
```diff
@@ -40,6 +42,21 @@ function collection_entity_bundle_field_info_alter(&$fields, EntityTypeInterface
 }
 
 /**
+ * Implements hook_views_pre_render().
+ */
+function collection_views_pre_view(ViewExecutable $view) {
```
idimopoulos: The view displays a list of collections but the view config file itself is located in the profile. This cache manipulation should be located there as well since it's only serving this particular page.

build.properties
pfrenssen: Yes older versions are removed from the download page. I have been looking at a way to automatically derive the latest version, but it doesn't seem to be simple, they don't have a 'latest' alias on their FTP site. So we'll have to update it manually for the moment.

web/modules/custom/collection/collection.module
pfrenssen: OK!

PR #320, opened 2017-01-03
Composer 1.3.0

claudiu-cristea:
See https://webgate.ec.europa.eu/CITnet/jira/browse/ISAICP-3079

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -139,7 +139,7 @@
         {
             "_comment": "The PhingDrushTask module does not currently have a D8 release and is not available on https://packages.drupal.org/8. See https://www.drupal.org/node/2817131.",
             "type": "vcs",
-            "url": "git@github.com:pfrenssen/phingdrushtask.git"
+            "url": "https://git.drupal.org/project/phingdrushtask.git"
         },
         {
             "type": "package",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -139,7 +139,7 @@
         {
             "_comment": "The PhingDrushTask module does not currently have a D8 release and is not available on https://packages.drupal.org/8. See https://www.drupal.org/node/2817131.",
             "type": "vcs",
-            "url": "git@github.com:pfrenssen/phingdrushtask.git"
+            "url": "https://git.drupal.org/project/phingdrushtask.git"
```
idimopoulos: Since it's now in the official repo, shouldn't we be able to directly remove it in all? I mean the repo definition! It should be automatically picked.

claudiu-cristea: Here's the problem. Our composer.json uses the Drupal 8 endpoint:

```json
    "repositories": [
        {
            "type": "composer",
            "url": "https://packages.drupal.org/8"
        },
        ...
    ],
```

but this module is **version agnostic**. That meaning the 7.x version is valid regardless of Drupal version. There's no 8.x version of the module so we want to use 7.x. There's no way to specify a 7.x version when the composer repository points to Drupal 8 endpoint. I opened this issue https://www.drupal.org/node/2840593 to see if they can support such cases. This is the reason we have to keep the repo entry.

idimopoulos: Cool! ok :)

PR #323, opened 2017-01-04
Update code to be compatible with latest stable release of Behat

pfrenssen:
We're using Behat Drupal Extension which has pinned a release candidate of Behat 3.1.0. However the latest release of Behat is 3.3.0. This PR explores the possibility of upgrading to the latest stable version.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -53,7 +53,7 @@
         "drupal/config_update": "~1.2",
         "drupal/default_content": "~1.0",
         "drupal/devel": "~1.0@alpha",
-        "drupal/drupal-extension": "dev-update-behat",
+        "drupal/drupal-extension": "dev-master#4d072006873a15a704764985c22dfb591c129c7e",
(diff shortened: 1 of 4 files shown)


pfrenssen:
Great stuff. We are ready if we apply a few minor fixes, we just need to ensure that we declare our callbacks as static methods. Since we were already doing this for the large majority of our callbacks, we can pride ourselves on having top class developer intuition.

We can actually already apply these fixes without actually updating to the latest version of Behat, then we will be ready.

pfrenssen:
I have reverted the composer dependencies. This is ready for review.

claudiu-cristea:
So, here we are only switching to static methods. But I don't see any change to composer dependencies.

pfrenssen:
No, let's wait until the upstream issue is merged first. This update will arrive naturally at some point when we update our composer dependencies.

claudiu-cristea:
I merged the changes. The dependencies to be updates in a future PR.

PR #326, opened 2017-01-09
Isaicp 3007

sandervd:


web/modules/custom/rdf_entity/modules/rdf_file/src/Plugin/Field/FieldType/RdfFileItem.php
```diff
@@ -0,0 +1,364 @@
...
+ *   label = @Translation("RDF File"),
+ *   description = @Translation("This field stores the ID of a file as an URI."),
+ *   category = @Translation("Reference"),
+ *   default_widget = "rdf_file_generic",
+ *   default_formatter = "rdf_file_default",
+ *   list_class = "\Drupal\rdf_file\Plugin\Field\FieldType\RdfFileFieldItemList",
+ *
+ * )
+ */
+class RdfFileItem extends EntityReferenceItem {
```
claudiu-cristea: I don't understand why you didn't start directly from FileItem as we already have the whole file functionality there. Then, in RdfFileItem, you simply extend `\Drupal\Core\Field\FieldItemInterface::schema()`:

```php
public static function schema(FieldStorageDefinitionInterface $field_definition) {
  $schema = parent::schema();
  $schema['fields']['uri'] = [
     'description' => 'The internal or external file URI.',
     'type' => 'varchar',
     'length' => 512,
  ];
  return $schema;
}
```

Actually a our field type is a file field having an additional column/property. As an edge implementation we can just use the standard `description` that already exists in FileItem and store there the URI.

The we extend the file widget and formatter to use this property.

web/modules/custom/rdf_entity/modules/rdf_file/src/Plugin/Field/FieldWidget/RdfFileWidget.php
```diff
@@ -0,0 +1,715 @@
...
+
+    // Essentially we use the managed_file type, extended with some
+    // enhancements.
+    $element_info = $this->elementInfo->getInfo('managed_file');
+    $element['#type'] = 'elements_need_a_type';
+    $element['#theme_wrappers'] = array('form_element');
+    $element['file-wrap']['#type'] = 'container';
+    $element['file-wrap']['select'] = [
+      '#type' => 'radios',
+      '#options' => [
```
claudiu-cristea: Let's make this more clear to the user. How about?
- Upload file
- Remote file URL

web/modules/custom/rdf_entity/modules/rdf_file/rdf_file.services.yml
```diff
@@ -0,0 +1,3 @@
+services:
+  rdf_file.handler:
```
sandervd: I'd prefer to keep this as a service to be able to swap it out:
- It has a contained purpose, that is called from a few places
- The URL is hard coded, so swapping it out might make sense.

claudiu-cristea: OK. Makes sense.

web/modules/custom/rdf_entity/modules/rdf_file/config/schema/rdf_file.schema.yml
```diff
@@ -0,0 +1,43 @@
+field.storage_settings.rdf_file:
```
idimopoulos: Simplified!

web/modules/custom/rdf_entity/modules/rdf_file/config/schema/rdf_file.schema.yml
```diff
@@ -0,0 +1,43 @@
...
+      type: integer
+      label: 'Enable Display field'
+    display_default:
+      type: integer
+      label: 'Files displayed by default'
+    uri_scheme:
+      type: string
+      label: 'Upload destination'
+
+field.field_settings.rdf_file:
```
idimopoulos: Simplified!

web/modules/custom/rdf_entity/modules/rdf_file/config/schema/rdf_file.schema.yml
```diff
@@ -0,0 +1,43 @@
...
+  type: base_rdf_file_field_field_settings
+  label: 'RDF file settings'
+  mapping:
+    description_field:
+      type: integer
+      label: 'Enable Description field'
+
+base_rdf_file_field_field_settings:
```
idimopoulos: Simplified!

web/modules/custom/rdf_entity/modules/rdf_file/config/schema/rdf_file.schema.yml
claudiu-cristea: Love it! 👍

PR #330, opened 2017-01-13
3044: Elibrary creation & Moderation

claudiu-cristea:
This issue was returned from UAT for clarifications on 'moderation' field. What this PC contains:

- Improved 'elibrary creation' to log inconsistent data.
- Field/checkbox 'moderation' is alway migrated as true (checked).
- I found also a bug after ISAICP-2835. This fix has its own commit and can be isolated for review.


web/modules/custom/rdf_entity/src/Entity/RdfEntitySparqlStorage.php
```diff
@@ -734,8 +735,9 @@ protected function doSave($id, EntityInterface $entity) {
       $id = $this->generateId();
       $entity->{$this->idKey} = (string) $id;
     }
-    elseif ($entity->isNew() && $this->idExists($id)) {
-      throw new \InvalidArgumentException("Attempting to create a new entity with the ID '$id' already taken.");
+    // The ID cannot overlap an existing 'rdf_entity' or 'taxonomy_term'.
```
sandervd: The storage layer can not/should not be aware of the entity types using it.
Also, issuing a full entity load seems overkill.
Even more important is that not all triples related to a mapped rdf:type.
(So the store might contain a lot more data then that is exposed in Drupal)

web/modules/custom/rdf_entity/rdf_entity.info.yml
```diff
@@ -6,7 +6,7 @@ package: Custom
 # These modules are required by the tests, must be available at bootstrap time
 dependencies:
   - options
-  - entity_reference
+  - taxonomy
```
sandervd: I don't feel comfortable with introducing this dependency.

web/modules/custom/rdf_entity/src/Entity/RdfEntitySparqlStorage.php
claudiu-cristea: @sandervd, as I told you...

{code:php}
protected function idExists($id) {
  $query = <<<QUERY
ASK {
  <$id> ?field ?value
}
QUERY;
  return $this->sparql->query($query)->isTrue();
}
{code}

is lying. It returns TRUE in some cases where there's no entity with that ID. I cannot find the reason. Let's open a follow-up and fix the problem there. In the meantime I would not block migration on this.

claudiu-cristea:
@sandervd, I reverted all changes related to "ID collision" and moved this fix in a follow-up.

PR #338, opened 2017-01-18
Isaicp 3099: Implement the EntityOwnerInterface for the rdf entity

idimopoulos:


web/modules/custom/demo_users/demo_users.module
```diff
@@ -33,7 +33,7 @@ function demo_users_rdf_entity_insert(EntityInterface $entity) {
 
   // Check if there is already a membership. If one of the facilitators create
   // an entity, then the membership will already exist.
-  if (!Og::getMembership($entity, $facilitator)) {
+  if (!Og::getMembership($entity, $facilitator) && ($facilitator->id() != \Drupal::currentUser()->id())) {
```
brummbar: Can you drop a line to explain this? From what I can understand, we skip creating the relationship when the facilitator is the current user because OG will handle this already, right?

web/modules/custom/rdf_entity/src/Entity/Rdf.php
```diff
@@ -208,6 +210,19 @@ public static function baseFieldDefinitions(EntityTypeInterface $entity_type) {
       ->setDescription(t('The Rdf type of this entity.'))
       ->setSetting('target_type', 'rdf_type');
 
+    $fields['uid'] = BaseFieldDefinition::create('entity_reference')
+      ->setLabel(t('Authored by'))
+      ->setDescription(t('The username of the content author.'))
```
brummbar: Nitpicking: should we change "content" to something else?

web/modules/custom/rdf_entity/src/Entity/Rdf.php
```diff
@@ -325,6 +340,55 @@ public function setUnpublished() {
   /**
    * {@inheritdoc}
    */
+  public function getOwner() {
+    return ($this->hasUidMapping() ? $this->get('uid')->entity : NULL);
```
brummbar: The wrapping parenthesis can be removed.

web/modules/custom/rdf_entity/src/Entity/Rdf.php
```diff
@@ -325,6 +340,55 @@ public function setUnpublished() {
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getOwnerId() {
+    return ($this->hasUidMapping() ? $this->getEntityKey('uid') : NULL);
```
brummbar: Same as above.

web/modules/custom/rdf_entity/src/Entity/Rdf.php
```diff
@@ -325,6 +340,55 @@ public function setUnpublished() {
...
+   * Returns whether the bundle of the entity has a mapping for the uid key.
+   *
+   * @return bool
+   *    Whether the entity bundle has a value for the uid key mapping.
+   */
+  protected function hasUidMapping() {
+    if (empty($this->bundle())) {
+      return NULL;
+    }
+    $bundle = $this->entityTypeManager()->getStorage('rdf_type')->load($this->bundle());
```
brummbar: Some kind of local static cache can be added here.

web/modules/custom/rdf_entity/tests/src/Kernel/RdfOwnerTest.php
```diff
@@ -0,0 +1,60 @@
...
+    ));
+    $owned->save();
+    $ownerless = Rdf::create(array(
+      'rid' => 'with_owner',
+      'label' => $this->randomMachineName(),
+      'uid' => NULL,
+    ));
+    $ownerless->save();
+
+    $this->assertNull($not_owned->getOwnerId(), "The entity with no mapping for uid does not have an owner.");
```
brummbar: Swap all double quotes with single quotes from here downwards.

web/modules/custom/demo_users/demo_users.module
idimopoulos: Actually, not the og membership. We are handling the relationship creation manually for rdf_entities in hooks (e.g. collection_entity_insert) since our rdf_entities do not have owners to have a membership created automatically.
Thus, when the collection facilitator demo user creates a collection, this membership creation is executed 2 times resulting in an exception.
Since this is a demo module, all exceptions should be handled here as the other hooks are the proper way to function.

web/modules/custom/rdf_entity/src/Entity/Rdf.php
brummbar: No, ignore the previous :)

idimopoulos: Cool :) Ignored.

web/modules/custom/rdf_entity/src/Entity/Rdf.php
idimopoulos: The bundle is statically cached after first load (this method can be called REALLY multiple times per request). After loading the bundle, the third party settings are in the loaded array! no request is being done directly in the config storage so it's fine I think. In the end, it is just an array seek.

idimopoulos:
QA remarks implemented.

web/modules/custom/rdf_entity/src/Entity/Rdf.php
brummbar: Oki

PR #342, opened 2017-01-24
3100: Modernize RDF Taxonomy. Implement RDF entity ID generator new plugin type

claudiu-cristea:
Known issues:

* `\Drupal\rdf_taxonomy\TermRdfStorage::alterGraph()`. I added that but without understanding exactly why. Without adding that resource the save fails. We should document it, probably or move it in the right place.
* `\Drupal\rdf_entity\Entity\RdfEntitySparqlStorage::hasGraph()` might be too expensive. Is there a straight way to get that info?
* Should "vocabulary locking" feature be moved from `rdf_taxonomy` to a custom  module?

---
This change is [Reviewable](https://reviewable.io/reviews/ec-europa/joinup-dev/342)


web/modules/custom/rdf_entity/rdf_taxonomy/src/TermRdfStorage.php
```diff
@@ -223,26 +255,20 @@ public function loadTree($vid, $parent = 0, $max_depth = NULL, $load_entities =
   FILTER (lang(?label) = 'en') .
   OPTIONAL {?tid <http://www.w3.org/2004/02/skos/core#broaderTransitive> ?parent }
 }
-ORDER BY DESC(?parent) ?tid
+ORDER BY ?label
```
claudiu-cristea: How to order case insensitively?

web/modules/custom/rdf_entity/src/Plugin/rdf_entity/Id/Fallback.php
```diff
@@ -0,0 +1,28 @@
...
+ *   label = @Translation("Default entity ID generator"),
+ * )
+ */
+class Fallback extends RdfEntityIdPluginBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function generate() {
+    return 'http://placeHolder/' . (new Php())->generate();
```
claudiu-cristea: While we are caching the instances in plugin manager, we should add a protected property `$php`, then init the value in constructor only once: `$this->php = new Php();`. Or even inject it as a service.

web/modules/custom/rdf_entity/src/RdfEntityIdPluginManager.php
```diff
@@ -0,0 +1,112 @@
...
+   * Initializes the proper plugin given a RDF entity.
+   *
+   * @param \Drupal\Core\Entity\EntityInterface $entity
+   *   The entity.
+   *
+   * @return \Drupal\rdf_entity\RdfEntityIdPluginInterface
+   *   The plugin.
+   */
+  public function getPlugin(EntityInterface $entity) {
+    $entity_type_id = $entity->getEntityTypeId();
```
claudiu-cristea: I think it would be a good idea. More practical from the user perspective. But what about entity types that don't have a UI for bundle edit? I guess we can store directly in config the 3rd party setting.

web/modules/custom/rdf_entity/rdf_draft/rdf_draft.module
```diff
@@ -60,7 +71,7 @@ function rdf_draft_entity_operation(EntityInterface $entity) {
   if ($storage  instanceof RdfEntitySparqlStorage) {
     $definitions = $storage->getGraphDefinitions();
     foreach ($definitions as $name => $definition) {
-      if ($entity->hasLinkTemplate('rdf-draft-' . $name)) {
+      if ($entity->hasLinkTemplate('rdf-draft-' . $name) && $entity->hasGraph($name)) {
```
claudiu-cristea: The templated is not set for an entity but for an entity type. Example: `rdf_entity` entity type has the template but specific entity, say a solution, of type `rdf_entity` might not have a `draft` version/graph. In that case we don't want to add the operation/link "View draft". There's no draft.

composer.json
```diff
@@ -28,7 +28,6 @@
         "drupal/search_api": "dev-1.x#ff41470b2b5399b83b86085d4b7a2c2f4e753370",
         "drupal/search_api_solr": "~1.0@beta",
         "drupal/state_machine": "~1.0-beta3",
-        "drupal/taxonomy_container": "~1.0",
```
claudiu-cristea: They are performing a strict `=== 0` there. I don't know.

web/modules/custom/rdf_entity/src/Entity/RdfEntitySparqlStorage.php
```diff
@@ -771,33 +789,92 @@ protected function doSave($id, EntityInterface $entity) {
...
           // then store it as a resource.
           if ($this->fieldIsReference($item)) {
             $graph->addResource((string) $id, (string) $properties['by_field'][$field_name][$column], $value);
           }
           // All other fields get stored as a literal.
           else {
-            // @todo Set language and field data type.
-            $graph->addLiteral((string) $id, (string) $properties['by_field'][$field_name][$column], $value);
+            $langcode = $this->resolveFieldLangcode($entity, $item);
+            // @todo Add datatype to literal.
```
claudiu-cristea: No, the Literal constructor accepts both:

```php
public function __construct($value, $lang = null, $datatype = null) {
  ...
}
```

composer.json
claudiu-cristea: I would not lose time with creating patches upstream. Anyway, that selector I think will be changed because it doesn't allow to select the parent term and, as I recall, at some point Andrea said that parent terms selections should be allowed.

web/modules/custom/rdf_entity/rdf_taxonomy/src/TermRdfStorage.php
claudiu-cristea: Fixed. Thank you!

web/modules/custom/rdf_entity/src/Plugin/rdf_entity/Id/Fallback.php
claudiu-cristea: Fixed.

web/modules/custom/rdf_entity/src/RdfEntityIdPluginManager.php
claudiu-cristea: Fixed. Thank you!

web/modules/custom/rdf_entity/src/Plugin/rdf_entity/Id/Fallback.php
claudiu-cristea: Fixed, but added also the entity type ID. Thank you!

PR #343, opened 2017-01-24
ISAICP-2262: Owner workflow

brummbar:
This change is [Reviewable](https://reviewable.io/reviews/ec-europa/joinup-dev/343)


tests/features/owner/owner.moderation.feature
```diff
@@ -0,0 +1,136 @@
+@api
+Feature: Owner moderation
+  In order to manage owners
+  As a user of the website
+  I need to be able to transit the owners from one state to another.
```
pfrenssen: Like from Texas to California?

tests/features/owner/owner.moderation.feature
```diff
@@ -0,0 +1,136 @@
...
+
+    # The owner asks for deletion.
+    When I am logged in as "Raeburn Hibbert"
+    And I go to the homepage of the "EU healthy group" owner
+    And I click "Edit" in the "Entity actions" region
+    Then the "State" select available options should be "Validated, Deletion request"
+    And the option "Validated" should be selected
+    When I select "Deletion request" from "State"
+    And I press "Save"
+    Then I should see the heading "EU healthy group"
```
pfrenssen: At this point something is wrong. Mr. Hibbert cannot edit the owner anymore now. So if he changes his mind he cannot undo the deletion request.

tests/features/owner/owner.moderation.feature
```diff
@@ -0,0 +1,136 @@
...
+
+    # The moderator cannot still delete the owner as it's referenced.
+    When I am logged in as a moderator
+    And I go to the homepage of the "EU healthy group" owner
+    And I click "Edit" in the "Entity actions" region
+    Then I should not see the link "Delete"
```
pfrenssen: Here something is not right too, the state now shows as "Validated" even though the actual state is "Deletion request". This gives the wrong information to the moderator.

I think the best solution would be to show the correct state, and also show the Delete link, but have it greyed out. We can then show a message alongside it to explain what needs to be done:

> A deletion request has been filed for this owner, but it is still assigned to the following collections:
> "The healthy food European project"
> Please assign a different owner to these collections before deleting the owner.

web/modules/custom/owner/src/Guard/OwnerFulfillmentGuard.php
```diff
@@ -0,0 +1,96 @@
...
+    $allowed_conditions = \Drupal::config('owner.settings')->get('transitions');
+
+    // Check if the user has one of the allowed system roles.
+    $authorized_roles = isset($allowed_conditions[$transition->getId()][$from_state]) ? $allowed_conditions[$transition->getId()][$from_state] : [];
+    $user = $this->workflowUserProvider->getUser();
+    if (array_intersect($authorized_roles, $user->getRoles())) {
+      return TRUE;
+    }
+
+    return FALSE;
```
pfrenssen: You can save a whole line here! :tada: 

```
return (bool) array_intersect(...)
```

web/modules/custom/rdf_entity/src/RdfAccessControlHandler.php
```diff
@@ -31,11 +34,20 @@ protected function checkAccess(EntityInterface $entity, $operation, AccountInter
         return AccessResult::allowedIfHasPermission($account, 'view rdf entity');
 
       case 'edit':
-        return AccessResult::allowedIfHasPermission($account, 'edit ' . $entity_bundle . ' rdf entity');
+        if ($account->hasPermission('edit ' . $entity_bundle . ' rdf entity')) {
+          return AccessResult::allowed();
+        }
+
+        return AccessResult::allowedIf($is_owner && $account->hasPermission('edit own ' . $entity_bundle . ' rdf entity'));
```
pfrenssen: Great and useful addition to `rdf_entity` :+1:

web/profiles/joinup/joinup.behat.inc
```diff
@@ -330,27 +332,19 @@ class JoinupSubContext extends DrupalSubContextBase implements DrupalSubContextI
...
    * @param string $select
    *    The name of the field element.
    * @param string $options
    *    The options to be checked separated by comma. Only labels are accepted.
    *
-   * @throws Exception
-   *    Throws an exception when a value does not exists.
-   *
-   * @Given (the ):select field has the :options option(s)
+   * @Given (the ):select field should contain the :options option(s)
```
pfrenssen: Since this is an assertion this should not be `@Given` but `@Then`.

web/profiles/joinup/joinup.behat.inc
```diff
@@ -364,18 +358,44 @@ class JoinupSubContext extends DrupalSubContextBase implements DrupalSubContextI
...
-      }
+  public function checkSelectDoesNotContainOptions($select, $options) {
+    $field = $this->findSelect($select);
+    $available_options = array_values($this->getSelectOptions($field));
+    $options = $this->explodeCommaSeparatedStepArgument($options);
+
+    $intersection = array_intersect($available_options, $options);
+
+    if (!empty($intersection)) {
+      throw new \Exception("The select '{$select}' should not contain the options: " . implode(', ', $intersection));
```
pfrenssen: Much better :+1:

web/profiles/joinup/joinup.behat.inc
```diff
@@ -364,18 +358,44 @@ class JoinupSubContext extends DrupalSubContextBase implements DrupalSubContextI
...
+   */
+  public function assertSelectOptions($select, $options) {
+    $field = $this->findSelect($select);
+    $available_options = $this->getSelectOptions($field);
+    sort($available_options);
+
+    $options = $this->explodeCommaSeparatedStepArgument($options);
+    sort($options);
+
+    // The PHPUnit assertion will allow to undestand easily which values are
```
pfrenssen: > undeRstand

pfrenssen:
Summary of the work I did:

* The Behat test can now also be run in browsers supporting JavaScript. I needed this so I could interact with the content half way through the test so I could see what was going on in the interface.
* Fixed two bugs that were preventing this from working in PHP 7.1: one in Phing and one in Composer.
* I initially changed the behaviour so that a facilitator can still edit an owner after requesting deletion. My motivation for this was that otherwise it is impossible for a facilitator to undo their action if they requested a deletion by accident. The only thing they can do is wait helplessly until their content is removed. I reverted this after consulting with @idimopoulos who rightfully argued that this is the exact behaviour requested by the business.
* Fixed a bug in Behat Drupal Extension that was causing random failures in the `owner.moderation.feature` scenario.
* Reworked the display of the owner entity form when it is shown inline in the collection entity form. This required to change the initial workflow state from `__new__` to `validated` when an owner entity is initially created. If the workflow state is missing from the form it doesn't get updated an the entity is saved in the database with the `__new__` state. This doesn't match with how we are using workflow in general. I figured this was a good solution since a new owner is automatically validated, it doesn't need anyone's permission. However are other ways to solve this, so feedback on this change is welcome.
* This PR was using the vanilla field widget from State Machine which displays the workflow state in a dropdown. This was not working correctly because `joinup_core_field_widget_form_alter()` completely changes the default behaviour. I explored a few possible avenues on this, but after consulting with the team we decided to hide the dropdown button entirely, and use the row of state buttons in the full entity form. I implemented this change.
* Created a new field widget that displays the current state in the full entity form. This is useful if a moderator wants to see the state an entity is in before acting on it.
* Reworked the user interaction regarding deletion requests. Originally it was possible for a facilitator to request deletion for an owner that was still in used. The moderator would then be presented with an owner with the "Deletion requested" state, but the "Delete" button would be missing on the form because the owner was still in use. This is very confusing, especially because no indication was provided on what was wrong and how to proceed. I solved it by preventing a facilitator from requesting deletion in the first place, and displaying a clear error message that lists the collections that need to be assigned a new user before the old one can be deleted. This will also prevent a moderator from having to needlessly spend effort on rejecting an invalid deletion request.
* Some small improvements, like documentation updates and small bug fixes.

composer.json
```diff
@@ -38,7 +38,7 @@
         "minimaxir/big-list-of-naughty-strings": "dev-master",
         "pear/console_table": "~1.3",
         "phing/phing": "^2.15.0",
-        "php": ">=5.5.9",
+        "php": ">=7.0",
```
claudiu-cristea: Is it possible to postpone this change? I'm running locally also the Drupal 6 site in the same time with the new D8. The legacy site is not ready for PHP7 and throws a lot of fatals making it unusable. I think we can develop for PHP7 without this. Take it as temporary.

web/modules/custom/owner/owner.workflows.yml
```diff
@@ -0,0 +1,42 @@
...
+  id: owner_workflow
+  label: "Default"
+  group: owner
+  states:
+    # The state used as initial.
+    __new__:
+      label: New
+    # State used to flag content that is being reviewed.
+    in_assessment:
+      label: In assessment
```
brummbar: It's a copy paste :) That property is never used.

web/modules/custom/owner/owner.module
```diff
@@ -5,7 +5,34 @@
...
+    return;
+  }
+
+  // The initial state of an owner is validated. The owner is created as part of
+  // proposing a collection, and in the collection propose form the workflow
+  // state is hidden from the user, so it would remain in the "__new__" state
+  // after submission.
+  // We can safely set the initial state to "validated" because a newly created
+  // owner doesn't need to be validated by a third party.
+  if ($entity->field_owner_state->value === '__new__') {
```
pfrenssen: Fine for me but how is the initial state set?

web/modules/custom/rdf_entity/src/RdfPermissions.php
```diff
@@ -41,19 +41,25 @@ public function rdfTypePermissions() {
    */
   protected function buildPermissions(RdfEntityType $type) {
     $type_id = $type->id();
-    $type_params = array('%type_name' => $type->label());
+    $type_params = ['%type_name' => $type->label()];
 
-    return array(
-      "create $type_id rdf entity" => array(
+    return [
+      "create $type_id rdf entity" => [
```
pfrenssen: There is no standard pattern for permissions in Drupal. They are human readable. As far as human readability goes, I much prefer "create owner rdf entity" over "create owner rdf_entity entity", especially because this is in the contributed RDF Entity module.

web/modules/custom/owner/owner.workflows.yml
```diff
@@ -0,0 +1,42 @@
+# The workflow of the owner entity.
+owner_workflow:
+  id: owner_workflow
+  label: "Default"
+  group: owner
+  states:
+    # The state used as initial.
+    __new__:
```
pfrenssen: Fixed.

web/modules/custom/owner/owner.workflows.yml
pfrenssen: Fixed.

web/modules/custom/owner/src/Guard/OwnerFulfillmentGuard.php
pfrenssen: Fixed.

web/modules/custom/owner/src/Guard/OwnerFulfillmentGuard.php
```diff
@@ -0,0 +1,96 @@
...
+   *    The machine name value of the state.
+   *
+   * @see https://www.drupal.org/node/2745673
+   */
+  protected function getState(RdfInterface $entity) {
+    if ($entity->isNew()) {
+      return $entity->field_owner_state->first()->value;
+    }
+    else {
+      $unchanged_entity = $this->entityTypeManager->getStorage('rdf_entity')->loadUnchanged($entity->id());
```
pfrenssen: Fixed.

web/modules/custom/owner/owner.module
```diff
@@ -57,23 +84,127 @@ function owner_inline_entity_form_entity_builder($entity_type, EntityInterface $
...
+    return AccessResult::neutral();
+  }
+
+  // By specifications, deletion should be forbidden if the owner entity is
+  // referenced by any other entity.
+  if ($operation === 'delete') {
+    return AccessResult::forbiddenIf(owner_is_owner_referenced($entity));
+  }
+
+  if ($account->hasPermission('administer rdf entity')) {
```
pfrenssen: I think @brummbar meant here that they should have access. After all this `administer rdf entity` permission is something that is granted to people that even have the power to create, update and delete RDF entity bundles.

pfrenssen: Fixed.

web/modules/custom/owner/owner.module
pfrenssen: It's set to the first state that is defined in `owner.workflows.yml`. Fixed.

composer.json
pfrenssen: OK moving back to 5.5.9 :(

I made some PHP 7 only changes, I'll go through the diff to update them.

pfrenssen:
I addressed the remarks but have not done any testing. The bot is having a go at it.

web/modules/custom/rdf_entity/src/RdfPermissions.php
idimopoulos: And should probably be changed (There is already a commit to change this in the other workflows).
I know that it really is much better to be an "rdf entity" though if we change it to use the entity type id, it is going to help us in cases where we cant to auto generate some permissions like in the comment above.
So sacrifice some minor readability instead of generating duplicated classes only to override one parameter.
And I am aware that nodes are changing this to 'content' but for example og is following the pattern below.

```
    $operations = [
      [
        'name' => "create $group_content_bundle_id $group_content_entity_type_id",
        'title' => $this->t('Create %bundle @entity', $args),
        'operation' => 'create',
      ],
```
It will really help us generating automatically the strings.

P.S. there are discussions changing the entitiy id itself in the end.

pfrenssen: This is a contributed module, we shouldn't make changes in it to benefit our project.

There are initiatives ongoing for reworking the permissions in Drupal, see for example https://www.drupal.org/node/2744683#comment-11274521

If we need machine based interaction with permissions then the best thing we can do is to abstract it in classes, similar to how I've done it in OG. Relying on some dumb fixed string format is a very poor way of dealing with this. See https://github.com/amitaibu/og/issues/222

Here's how this could look: https://github.com/Gizra/og/blob/8.x-1.x/src/Permission.php

I already wrote all of this, so adapting it to our use case is not going to be too difficult.

pfrenssen: This is an example of how OG deals with the Node permissions (from `OgEventSubscriber::provideDefaultNodePermissions()`:

```
      $permission_values = [
        [
          'name' => "create $bundle_id content",
          'title' => $this->t('%type_name: Create new content', $args),
          'operation' => 'create',
        ],
```

So for RDF entities this should be:

```
      $permission_values = [
        [
          'name' => "create $type_id rdf entity",
          'title' => $this->t('Create new %type_name RDF entity', $args),
          'operation' => 'create',
        ],
```

PR #348, opened 2017-01-27
Update Composer dependencies. Fixes PHP 7.1 compatibility.

pfrenssen:


pfrenssen:
I need the update of Phing to the latest version, or I get build errors in PHP 7.1: https://github.com/phingofficial/phing/releases/tag/2.16.0

I also had a quick gander at the composer.json, and saw that Field Group has a new release so we don't have to follow the development version any more. I also locked Inline Entity Form which was no longer locked for some reason.

claudiu-cristea:
Discussing with @pfrenssen on chat to do the Drupal core in other step, probably we'll switch directly to 8.3.0-alpha. This is good to go.

PR #350, opened 2017-01-31
Isaicp 3095

sandervd:


claudiu-cristea:
OK, let's commit this. I needed a specialized exception class only for this type of errors so that I can catch them in the migration but i will add that when I'll need it.

PR #353, opened 2017-02-03
Migrate Week 6

claudiu-cristea:
Contains:

- [ISAICP-3091](https://webgate.ec.europa.eu/CITnet/jira/browse/ISAICP-3091): Migrate D6 Project into D8 Interoperability Solution.
- [ISAICP-2840](https://webgate.ec.europa.eu/CITnet/jira/browse/ISAICP-2840): Migrate D6 Project Release to D8 Release.
- [ISAICP-3054](https://webgate.ec.europa.eu/CITnet/jira/browse/ISAICP-3054): Migrate Collection owner (admin), facilitator and members

---
This change is [Reviewable](https://reviewable.io/reviews/ec-europa/joinup-dev/353)


web/modules/custom/rdf_entity/src/Entity/Controller/RdfListBuilder.php
```diff
@@ -4,43 +4,65 @@
...
     }
 
+    if ($rid = $request->get('rid') ?: NULL) {
+      $rid = in_array($rid, array_keys($bundle_info->getBundleInfo('rdf_entity'))) ? [$rid] : NULL;
+    }
+
+    $query->condition('rid', $rid, 'IN');
+    // Special treatment for 'solution' and 'asset_release'.
+    // @see https://webgate.ec.europa.eu/CITnet/jira/browse/ISAICP-3126
+    if ($rid[0] === 'asset_release') {
```
claudiu-cristea: True, this doesn't belong here as this is a contrib module. We should find a way to alter this from `joinup_core` module or just swap the entire class with a custom one. The 2nd is perfectly doable in `joinup_core_entity_type_alter()`. After ISAICP-3126 is fixed the workaround can be dropped.

sandervd: Could we remove this for now?
Only administrators (us) have access to this, so the benefits of having this in don't seem to outweigh the added technical depth.

web/modules/custom/rdf_entity/src/Entity/Controller/RdfListBuilder.php
```diff
@@ -51,22 +73,9 @@ public function load() {
...
-      // Embed the graph selection form.
-      $form = \Drupal::formBuilder()->getForm('Drupal\rdf_entity\Form\GraphSelectForm', $options);
-      if ($form) {
-        $build['graph_form'] = $form;
-      }
-    }
-    $build['table'] = parent::render();
-    return $build;
+    return [
+      'filter_form' => \Drupal::formBuilder()->getForm(RdfListBuilderFilterForm::class),
```
sandervd: Nice!

web/modules/custom/rdf_entity/src/Entity/Controller/RdfListBuilder.php
```diff
@@ -4,43 +4,65 @@
...
+    $bundle_info = \Drupal::service('entity_type.bundle.info');
+    /** @var \Drupal\rdf_entity\Entity\Query\Sparql\Query $query */
+    $query = $rdf_storage->getQuery();
 
-    $query = $rdf_storage->getQuery()->condition('rid', NULL, 'IN');
     // If a graph type is set in the url, validate it, and use it in the query.
-    if (!empty($_GET['graph'])) {
-      $def = $rdf_storage->getGraphDefinitions();
-      if (is_string($_GET['graph']) && isset($def[$_GET['graph']])) {
+    $graph = $request->get('graph');
```
sandervd: Nice

web/modules/custom/rdf_entity/src/Entity/Controller/RdfListBuilder.php
claudiu-cristea: Yes, no problem. But that I added because very often I have to check if an entity was migrated. It's very practical to have that filter in place. I will find a workaround.

PR #358, opened 2017-02-06
3126: Sparql entity query unable to distinguish between 'solution' and 'asset_release' bundles

claudiu-cristea:
This change is [Reviewable](https://reviewable.io/reviews/ec-europa/joinup-dev/358)


idimopoulos:
This has been implemented as part of ISAICP-2503 and is covered in https://github.com/ec-europa/joinup-dev/pull/437 where we use the default tag functionality.

Closing :)

PR #359, opened 2017-02-06
ISAICP-2696: Event overview page

tom-waat:
Merge ISAICP-2696 into develop


web/themes/joinup/templates/content/node--event--full.html.twig
```diff
@@ -0,0 +1,107 @@
+{#
+/**
+ * @file
+ * Theme override to display a node.
```
brummbar: Let's start to change the @file documentation. Remove the copy pasted variables, maybe add the new ones passed to the template (if any). Otherwise just
```
@file
Theme override to display a node of bundle event.
```
or similar.

PR #349, opened 2017-01-30
Isaicp 2266 backport: Backport of the generic test base for community content to the discussion entity.

idimopoulos:

PR #365, opened 2017-02-08
Use dependency injection in the Guard classes rather than sideloading…

pfrenssen:
… services.


idimopoulos:
Nice, thanks! though a lot of these are changed in https://github.com/ec-europa/joinup-dev/pull/349 but still some of them are needed. I will update the other branches! :)

PR #368, opened 2017-02-11
ISAICP-3153

pfrenssen:


web/profiles/joinup/src/Access/JoinupCustomAccess.php
```diff
@@ -0,0 +1,61 @@
...
+        // acceptance testing. This will be removed once we have custom
+        // controllers and links in the dashboard or the plus button menu.
+        // Non-moderators never have access to the default entity create forms,
+        // they can only create RDF entities through the custom controllers via
+        // the plus button.
+        $temporary_allowed_bundles = [
+          'asset_distribution',
+          'asset_release',
+          'contact_information',
+          'owner',
```
idimopoulos: Are we meant to have these controllers? Will the user be able at some point to create contact_information or owners directly?
In any case:
 - We do have controllers for asset distributions and asset releases but since they are direct children of solutions, the 'Add release' appears only for the solution context and the 'Add distribution' appears for solution and release context.
 - The users can still add contact information and owners through the collection/solution edit and there is no specific request for being able to create them directly yet, so I would suggest to remove this entirely and have a ticket to create the controllers for them because we do not need the direct form for any test at the moment.

PR #371, opened 2017-02-13
Isaicp 2263: Insert a __new__ state to the workflow. Revert the removal of the __new__ state from the owner workflow. 

idimopoulos:
The issue is that in the rdf add form ('/rdf_entity/add/contact_information'), the state is by default the validated so the transition buttons are 'Update', 'Request changes' and 'Request deletion'.

Please check the ticket online.


idimopoulos:
Terrible mistake.... the correct PR is https://github.com/ec-europa/joinup-dev/pull/372

PR #374, opened 2017-02-14
ISAICP-2327: Disabling a discussion

brummbar:


tests/features/joinup_discussion/discussion.moderation.feature
```diff
@@ -87,3 +87,52 @@ Feature: Discussion moderation
...
+
+  Scenario: Disabling a discussion prevents additional comments to be created.
+    Given users:
+      | name          | mail                      |
+      | Vince Rome    | vince.rome@example.com    |
+      | Lance Rustici | lance.rustici@example.com |
+      | Denny Winslow | denny.winslow@example.com |
+    And the following collection:
+      | title             | Valentine's day survival kit                   |
+      | description       | How to survive the most scary day of the year. |
```
idimopoulos: First world's greatest fears revealed! :+1:

web/modules/custom/joinup_discussion/src/EventSubscriber/WorkflowTransitionEventSubscriber.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+namespace Drupal\joinup_discussion\EventSubscriber;
+
+use Drupal\comment\Plugin\Field\FieldType\CommentItemInterface;
+use Drupal\state_machine\Event\WorkflowTransitionEvent;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+/**
+ * Event subscriber to close comments on archived discussions.
```
idimopoulos: The ticket in stake talks about disabling discussions, not archiving. 
Can we stick with one terminology since the action is also called "Disable"?

web/modules/custom/joinup_discussion/src/EventSubscriber/WorkflowTransitionEventSubscriber.php
```diff
@@ -0,0 +1,33 @@
...
+   * {@inheritdoc}
+   */
+  public static function getSubscribedEvents() {
+    return [
+      'discussion.disable.pre_transition' => 'closeComments',
+    ];
+  }
+
+  /**
+   * Close discussion comments when a discussion is archived.
```
idimopoulos: Same here, 'discussion is disabled."

web/modules/custom/joinup_discussion/src/EventSubscriber/WorkflowTransitionEventSubscriber.php
```diff
@@ -0,0 +1,33 @@
...
+ * Event subscriber to close comments on archived discussions.
+ */
+class WorkflowTransitionEventSubscriber implements EventSubscriberInterface {
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function getSubscribedEvents() {
+    return [
+      'discussion.disable.pre_transition' => 'closeComments',
```
idimopoulos: We should fire it in the post_transition. This is an action on a condition. The condition is the transition, so we should do that after it's applied.

web/modules/custom/joinup_discussion/src/EventSubscriber/WorkflowTransitionEventSubscriber.php
brummbar: Action is called disable, state is called archived to uniform with the rest of the states. What are we going to use?

web/modules/custom/joinup_discussion/src/EventSubscriber/WorkflowTransitionEventSubscriber.php
brummbar: Post-transition is too late, it will force a second save on the entity. We already handle these things on pre-transitions for things like new revision and published state.

web/modules/custom/joinup_discussion/src/EventSubscriber/WorkflowTransitionEventSubscriber.php
brummbar: Same as above.

web/modules/custom/joinup_discussion/src/EventSubscriber/WorkflowTransitionEventSubscriber.php
idimopoulos: Nahh it's fine then :D

PR #376, opened 2017-02-15
ISAICP-2337: keeping the user logged in

brummbar:


brummbar:
The PHP7.1 test is red due to machine issues. 7.0 behat and 7.0/7.1 phpunit tests are green, so merging.

PR #382, opened 2017-02-20
ISAICP-2991: User detail page

brummbar:


tests/features/bootstrap/FeatureContext.php
```diff
@@ -386,6 +386,42 @@ public function assertFieldOptionSelected($option, $select) {
...
+   * @param string $field
+   *   The label of the field the radio is part of.
+   *
+   * @throws \Exception
+   *   Thrown when the field or the radio is not found, or if the radio is not
+   *   checked.
+   *
+   * @Then the radio :radio from field :field should be checked
+   */
+  public function assertFieldRadioSelected($radio, $field) {
```
brummbar: Ok for me, but this was actually changed by you in the previous QA.
Commit `79b7b542cb729041a4bdcaafa22b9d360a51e142`:
"ISAICP-2991: In usual terminology a radio button is 'selected' rather than 'checked'."
What to do?

web/modules/custom/joinup_user/joinup_user.module
```diff
@@ -210,3 +215,75 @@ function joinup_user_form_alter_validate(&$form, FormStateInterface $form_state)
...
+ * - show only the content the user is author of;
+ * - do not show entities of some bundles.
+ */
+function joinup_user_search_api_query_search_api_field_field_user_content_alter(QueryInterface &$query) {
+  /** @var \Drupal\user\Entity\User $user */
+  $user = $query->getOption('search_api_field entity');
+
+  // Do not show contact information and owner entities as they are not
+  // relevant.
+  $query->addCondition('entity_bundle', ['contact_information', 'owner'], 'NOT IN');
```
brummbar: While generally whitelist is indeed better, we don't have any plans to hide any other content here. Actually, this comes from a search, and we index everything that we want to be shown. Only in this case, we don't want to show owners and contact informations as they are shown in the related entity "About" page.

tests/features/bootstrap/FeatureContext.php
pfrenssen: No just rename the step definition to "selected" like I proposed. One does not check a radio button.

* Checking is for checkboxes
* Selecting is for radio buttons

web/modules/custom/joinup_user/joinup_user.module
pfrenssen: OK fair enough

tests/features/bootstrap/FeatureContext.php
brummbar: Ok, then it has been already done with said commit by you on previous round of QA 😄  (see latest revision [here](https://github.com/ec-europa/joinup-dev/pull/382/files#diff-a2ccc714735c13b3a710a340994f3c74R405)).

web/modules/custom/joinup_user/joinup_user.module
brummbar: I'll update the inline documentation to reflect this concept.

PR #411, opened 2017-03-10
Isaicp 2709

sandervd:


composer.json
```diff
@@ -22,7 +22,7 @@
         "drupal/migrate_report": "~1.0-alpha1",
         "drupal/migrate_run": "~1.0-alpha1",
         "drupal/migrate_spreadsheet": "~1.0-beta1",
-        "drupal/og": "dev-1.x#34a9425ffeeb267ceeba21261f4719d0ced36e58",
+        "drupal/og": "dev-ISAICP-2709 as dev-1.x",
```
idimopoulos: I checked the PR in Og. 
Are you sure this ensures us?
You are adding the 'administer group' permission to the facilitator. This means that no matter what happens, when a new menu item or a new addition to the admin UI is added upstream, the facilitator will automatically acquire new abilities which we do not handle or currently have a way to see immediately.

I suggest that we either duplicate the form or at least for now, add a small test to ensure that only one menu item called 'Members' is found in the admin form. At least that ensures us that we get changes right away (though the first option is better).

web/modules/custom/collection/collection.module
```diff
@@ -109,7 +109,7 @@ function collection_rdf_entity_access(EntityInterface $entity, $operation, Accou
 
     case 'edit':
       $allowed_transitions = $entity->field_ar_state->first()->getTransitions();
-      return AccessResult::forbiddenIf(empty($allowed_transitions));
+      return empty($allowed_transitions) ? AccessResult::forbidden() : AccessResult::allowed();
```
idimopoulos: FYI, we faced the same issue with the node workflows and we ended up having this way of checking.

sandervd: I don't get it? How is this related to assigning roles?

composer.json
sandervd: I've created a custom route to the same controller, but with the 'administer members' permission.

tests/features/collection/collection.member_administration.feature
```diff
@@ -0,0 +1,49 @@
...
+    And I go to the "Medical diagnosis" collection
+    Then I go to the "Medical diagnosis" collection edit form
+    Then I should see the heading "Access denied"
+
+    # Dr Cuddy promotes Dr House to facilitator.
+    When I am logged in as "Lisa Cuddy"
+    And I go to the "Medical diagnosis" collection
+    Then I click "Group" in the "Entity actions"
+    Then I click "Members"
+    Then I check the box "Update the member Gregory House"
```
sandervd: I've added a check that will fail if it sees the default og 'Group' tab.

web/modules/custom/collection/collection.module
```diff
@@ -254,3 +254,25 @@ function collection_search_api_query_search_api_field_field_collection_content_a
...
+
+/**
+ * Implements hook_form_alter().
+ *
+ * Remove non-collection related actions (adding/removing roles)
+ * from member overview page.
+ */
+function collection_form_alter(&$form, FormStateInterface $form_state, $form_id) {
+  // Bulk operation form ids vary on entity id.
+  if (strstr($form_id, 'views_form_og_members_overview_default') === FALSE) {
```
sandervd: Sadly, we can't. The form id is something like views_form_og_members_overview_default_rdf_entity_http://joinup....
So it contains the id of the group entity.

web/modules/custom/collection/collection.module
idimopoulos: This is not. Sorry. It is a hotfix that we had elsewhere but we did not backport for solutions. It was causing an access result failure on the site so I thought I could fix it right away since it's easy and we already test it elsewhere.

PR #417, opened 2017-03-15
Isaicp 2752: Load more prototype

idimopoulos:


web/themes/joinup/prototype/js/script.js
```diff
@@ -0,0 +1,75 @@
+/**
```
idimopoulos: We are probably not going to use this script (entirely or part of it), especially since we intend to always show the "Load more" button in the last part of the row of 3 items. If there are less than 3 items (including the Load more button) that means that there are no more items to be shown and we do not show the button.

PR #418, opened 2017-03-15
Isaicp 3146: Distribution deletion

idimopoulos:
Now includes the notification functionality.


web/modules/custom/joinup_notification/joinup_notification.module
```diff
@@ -30,18 +31,20 @@ function joinup_notification_entity_predelete(EntityInterface $entity) {
 
   $entity_type = $entity->getEntityTypeId();
   $bundle = $entity->bundle();
-  $field_definitions = array_filter(\Drupal::entityManager()->getFieldDefinitions($entity_type, $bundle), function ($field_definition) {
-    return $field_definition->getType() == 'state';
-  });
+  if ($entity_type !== 'rdf_entity' || $bundle !== 'asset_distribution') {
```
brummbar: Drop a line of documentation here to explain why only the asset_distribution bundle is excluded. I could understand it from the commit history, but let's make it easier for the next devs.

web/modules/custom/joinup_notification/src/NotificationSenderService.php
```diff
@@ -31,16 +33,36 @@ class NotificationSenderService {
   protected $messageNotifySender;
 
   /**
+   * The discussions relation manager.
```
brummbar: Isn't this the global relation manager?

web/modules/custom/joinup_notification/config/install/message.template.rdf_entity_distribution_delete.yml
```diff
@@ -0,0 +1,19 @@
+langcode: en
+status: true
+dependencies: {  }
+template: rdf_entity_distribution_delete
+label: 'Rdf entity - asset distribution - delete'
+description: 'Message to the owner when a distribution is deleted'
+text:
+  -
+    value: "<p>Joinup - Content has been deleted</p>\r\n"
```
brummbar: Change "content" with "distribution", so at least when we receive an e-mail we understand what is the e-mail for.
And sneak in also the same change on the solution delete mail 😄

PR #422, opened 2017-03-20
3230: Testing migrations

claudiu-cristea:


web/modules/custom/rdf_entity/tests/src/Kernel/RdfKernelTestBase.php
```diff
@@ -28,8 +28,8 @@
    * {@inheritdoc}
    */
   protected function setUp() {
-    $this->setUpSparql();
     parent::setUp();
+    $this->setUpSparql();
```
claudiu-cristea: Just curious. For me that didn't work.

PR #423, opened 2017-03-20
Isaicp 3046: Display a link to the latest version of the rdf_entity

idimopoulos:


web/modules/custom/joinup_core/joinup_core.module
```diff
@@ -233,3 +235,32 @@ function joinup_core_form_alter(&$form, FormStateInterface $form_state, $form_id
...
+  }
+
+  // The draft link is already set in the list of the link templates of the rdf
+  // entity as 'rdf-draft-{$graph_name}'.
+  $url = $entity->toUrl('rdf-draft-draft');
+  if (!$url->access()) {
+    return;
+  }
+
+  $link = Link::fromTextAndUrl('here', $url)->toString();
```
brummbar: `here` needs to go through `t()`.

web/modules/custom/joinup_core/joinup_core.module
```diff
@@ -233,3 +235,32 @@ function joinup_core_form_alter(&$form, FormStateInterface $form_state, $form_id
     }
   }
 }
+
+/**
+ * Implements hook_ENTITY_TYPE_view().
+ *
+ * Shows an information message to users who have access to the draft version
```
brummbar: The problem of using `hook_entity_view()` is that if we render the draft graph version somewhere else, we will get the message there too. Shouldn't we show this message only on the canonical route (as we said in the ticket)?

idimopoulos: We have discussed with @sandervd and @brummbar and decided that we can have this implemented in the full view at least for now since there are no current plans to show a full view in another page other than the canonical view of the entity.

web/modules/custom/joinup_core/joinup_core.module
idimopoulos: Fixed.

web/modules/custom/joinup_core/joinup_core.module
brummbar: I confirm 😃

PR #429, opened 2017-03-27
ISAICP-2949: Document tiles should show the file type and weight

brummbar:


kalinchernev:
No problems found in implementation or styling. More details about possible comments in the ticket.

kalinchernev:
The visualization as tested:
file_example

PR #431, opened 2017-03-27
Issue #ISAICP-3242: Fix Index on uat machine, as config_ui is not

brummbar:
enabled there.


idimopoulos:
This is a very old ticket and not valid anymore.

PR #434, opened 2017-03-28
Revert 3234 (views > tables)

claudiu-cristea:
On a new install I enabled the `joinup_migrate` module. Then, looking into the new tables `d8_contact` and `d8_owner`, I see them both empty. And this is logic because they are both based on view `d8_solution`. And that is a view based on other migrations that were not run yet.

**Solution**: Use different, lightweight, views for `solution` migration.


idimopoulos:
Accepted in QA.

PR #437, opened 2017-04-03
Isaicp 2503: Refactor the SPARQL query.

idimopoulos:


web/modules/custom/rdf_entity/src/Entity/Query/Sparql/SparqlCondition.php
```diff
@@ -4,47 +4,785 @@
...
+        $predicate = $this->fieldMappings[$field_name];
+        $langcode = $this->getLangCode($condition['field'], $condition['column'], $condition['lang']);
+        if (in_array($condition['operator'], $this->requiresTriple) && isset($this->fieldMappings[$field_name])) {
+          $this->addConditionFragment(self::ID_KEY . ' ' . $this->escapePredicate($this->fieldMappings[$field_name]) . ' ' . $this->toVar($field_name));
+        }
+      }
+
+      $condition['value'] = $this->escapeValue($condition['field'], $condition['value'], $condition['column'], $condition['lang']);
+      $condition['field'] = $field_name;
+      switch ($condition['operator']) {
```
sandervd: If we could split this up too 😄

web/modules/custom/rdf_entity/src/RdfFieldHandler.php
```diff
@@ -0,0 +1,577 @@
...
+        $bundle_definitions_ids = array_keys($bundle_definitions);
+        $bundle_mapping = $rdf_bundle_entity->getThirdPartySetting('rdf_entity', 'rdf_type');
+        if (empty($bundle_mapping)) {
+          throw new \Exception("The {$rdf_bundle_entity->label()} rdf entity does not have an rdf_type set.");
+        }
+        $this->outboundMap[$entity_type_id]['bundles'][$rdf_bundle_entity->id()] = $bundle_mapping;
+        // More than one drupal bundle can share the same mapped uri.
+        $this->inboundMap[$entity_type_id]['bundles'][$bundle_mapping][] = $rdf_bundle_entity->id();
+        foreach ($storage_definitions as $id => $storage_definition) {
+          if (!in_array($id, $bundle_definitions_ids)) {
```
sandervd: Extract method

web/modules/custom/rdf_entity/src/RdfFieldHandler.php
```diff
@@ -0,0 +1,577 @@
...
+   *   Optional. Pass the language if one exists. This should be null if the
+   *    format is not t_literal.
+   * @param string $column
+   *   The column for which to calculate the value. If null, the field's main
+   *    column will be used.
+   *
+   * @return mixed
+   *   The calculated value.
+   */
+  public function getOutboundValue($entity_type_id, $field, $value, $lang = NULL, $column = NULL) {
```
sandervd: name

web/modules/custom/rdf_entity/src/RdfFieldHandler.php
```diff
@@ -0,0 +1,577 @@
...
+   * @param string $bundle
+   *   The bundle id.
+   *
+   * @return string
+   *   The bundle mapping.
+   *
+   * @throws \Exception
+   *    Thrown when the bundle is not found.
+   */
+  public function getOutboundBundleValue($entity_type_id, $bundle) {
```
sandervd: this too (name)

web/modules/custom/rdf_entity/src/RdfFieldHandler.php
```diff
@@ -0,0 +1,577 @@
...
+   * @param string $bundle
+   *   The bundle uri.
+   *
+   * @return array
+   *   An array of bundles that match the requested bundle.
+   *
+   * @throws \Exception
+   *    Thrown when the bundle is not found.
+   */
+  public function getInboundBundleValue($entity_type_id, $bundle) {
```
sandervd: name

web/modules/custom/rdf_entity/src/RdfFieldHandler.php
```diff
@@ -0,0 +1,577 @@
...
+
+  /**
+   * Returns an array of available datatypes.
+   *
+   * @return array
+   *   An array of datatypes.
+   */
+  public static function getSupportedDatatypes() {
+    return [
+      self::RESOURCE => t('Resource'),
```
sandervd: follow up: create a plugin system

web/modules/custom/rdf_entity/src/Entity/Query/Sparql/SparqlArg.php
```diff
@@ -41,16 +109,69 @@ public static function uri($uri) {
...
+   *   The outcome of the serialization.
+   */
+  public static function serialize($value, $format, $lang = NULL) {
+    $data['value'] = $value;
+    switch ($format) {
+      case RdfFieldHandler::RESOURCE:
+        $data['type'] = 'uri';
+        break;
+
+      case RdfFieldHandler::NON_TYPE:
```
idimopoulos: NON_TYPE is also a type. It is actually a literal. It is just a non-type for us because we do not need special handling. We just need to escape dangerous string parts (we still handle it and pass it through a serializer).

web/modules/custom/demo_content/content/rdf_entity/solution/core-person-vocab.json
```diff
@@ -56,7 +56,7 @@
     },
     "id": [
         {
-            "value": "http:\/\/www.w3.org\/ns\/person#"
```
idimopoulos: Good question. That was a temporary solution for the issue we were facing with the id. I _was_ planning to revert. Good that you caught it.

web/modules/custom/rdf_entity/src/Entity/RdfEntitySparqlStorage.php
```diff
@@ -910,7 +848,8 @@ protected function getColumnSchema(FieldItemInterface $item, $column) {
    *   Response.
    */
   private function insert(Graph $graph, $graphUri) {
-    $query = "INSERT DATA INTO <$graphUri> {\n";
+    $graphUri = SparqlArg::uri($graphUri);
+    $query = "INSERT DATA INTO $graphUri {\n";
```
idimopoulos: Indeed we could!

web/modules/custom/rdf_entity/src/Entity/RdfEntitySparqlStorage.php
```diff
@@ -940,18 +879,21 @@ public function hasData() {
...
+   * @param string $type
+   *   The field type.
    * @param array $values
    *   The field values.
+   *
+   * @todo: To be removed when columns will be supported. No need to manually
+   * set this.
    */
-  private function applyFieldDefaults(FieldStorageConfig $storage, array &$values) {
+  private function applyFieldDefaults($type, array &$values) {
```
idimopoulos: Indeed we should, but maybe we can afford to do this in a follow up.

web/modules/custom/rdf_entity/src/Entity/Query/Sparql/SparqlCondition.php
```diff
@@ -4,47 +4,785 @@
...
+    '<' => ['prefix' => '', 'suffix' => ''],
+    '>' => ['prefix' => '', 'suffix' => ''],
+    '>=' => ['prefix' => '', 'suffix' => ''],
+    '<=' => ['prefix' => '', 'suffix' => ''],
+  ];
+
+  /**
+   * The operators that require a default triple to be added.
+   *
+   * In SPARQL, some of the SQL conditions might need more than one conditions
```
idimopoulos: Well, I meant SQL actually. Like, some of the SQL conditions [if you convert them to RQL] they need more than one pattern to be expressed! because we still use the
`$query->condition('field', 4, '<');` but it needs 1 pattern and 1 filter on RQL.

idimopoulos: I will change it to
`In SPARQL, some of the conditions might need a combination of a pattern and
   * a condition.`

web/modules/custom/rdf_entity/src/Entity/Query/Sparql/SparqlCondition.php
```diff
@@ -4,47 +4,785 @@
...
+   * combined. Below are the operators that need a secondary condition.
+   *
+   * @var array
+   *    An array of operators.
+   */
+  protected $requiresTriple = [
```
idimopoulos: Changed it to `requiresDefaultPattern` :D

web/modules/custom/rdf_entity/src/Entity/Query/Sparql/SparqlCondition.php
```diff
@@ -4,47 +4,785 @@
...
+   *
+   * TRUE if the condition has been changed since the last compile.
+   * FALSE if the condition has been compiled and not changed.
+   *
+   * @var bool
+   */
+  protected $changed = TRUE;
+
+  /**
+   * Whether the conditions do not have a triple.
```
idimopoulos: Changed to `Whether the default triple pattern is required in the query.`

web/modules/custom/rdf_entity/src/Entity/Query/Sparql/SparqlCondition.php
```diff
@@ -4,47 +4,785 @@
...
+   *
+   * This will be turned to false if there is at least one condition that does
+   * not involve the id.
+   *
+   * @var bool
+   */
+  protected $needsDefault = TRUE;
```
idimopoulos: Changed to `$requiresDefaultPattern`

web/modules/custom/rdf_entity/src/Entity/Query/Sparql/SparqlCondition.php
```diff
@@ -4,47 +4,785 @@
...
+
+  /**
+   * The default bundle predicate.
```
idimopoulos: Yeah we need to discuss this first..

web/modules/custom/rdf_entity/src/Entity/Query/Sparql/SparqlCondition.php
```diff
@@ -4,47 +4,785 @@
...
+   *
+   * @var array
+   */
+  protected $typePredicate = '<http://www.w3.org/1999/02/22-rdf-syntax-ns#type>';
+
+  /**
+   * An array of triples in their string version.
```
idimopoulos: The 'fragment' suffix was taken as an idea from the core condition. It is not related to the SPARQL. It describes a fragment as a piece of string that will be concatenated in the end of the compilation phase with the rest of the strings to form a string version of the query.

web/modules/custom/rdf_entity/src/Entity/Query/Sparql/SparqlCondition.php
```diff
@@ -4,47 +4,785 @@
...
+   * There is no filter per bundle in the query. That makes it safe to not check
+   * on the predicate mappings that are already in the query.
    */
   public function compile($query) {
-    foreach ($this->conditions() as $condition) {
-      $query->query .= $condition['subject'] . ' ' . $condition['predicate'] . ' ' . $condition['object'] . ".\n";
+    $entity_type = $query->getEntityType();
+    $condition_stack = array_merge($this->conditions, $this->fieldMappingConditions);
+    foreach ($condition_stack as $index => $condition) {
+      if (in_array($condition['field'], [$this->idKey, $this->bundleKey])) {
```
idimopoulos: I did not extract this (at least for now) but I did refactor it a bit.

web/modules/custom/rdf_entity/src/Entity/Query/Sparql/SparqlCondition.php
```diff
@@ -4,47 +4,785 @@
...
+    // The fieldMappingConditions are added first because they are converted
+    // into a 'VALUES' clause which increases performance.
+    $this->fieldMappingConditionsToString();
+
+    // Convert the conditions.
+    $this->conditionsToString();
+
+    // Put together everything.
+    $condition_fragments = array_merge($this->tripleFragments, $this->conditionFragments);
+    $this->stringVersion = implode(self::$conjunctionMap[$this->conjunction]['delimeter'], array_unique($condition_fragments));
```
idimopoulos: Yes boss! xD

web/modules/custom/rdf_entity/src/Entity/Query/Sparql/SparqlCondition.php
```diff
@@ -4,47 +4,785 @@
...
+      $this->addConditionFragment(self::ID_KEY . ' ' . $this->escapePredicate($field_predicate) . ' ' . $this->toVar($field_name));
+      $condition['value'] = SparqlArg::toResourceUris($condition['value']);
+      $this->addConditionFragment($this->compileValuesFilter($condition));
+    }
+  }
+
+  /**
+   * Converts the conditions into string versions.
+   */
+  private function conditionsToString() {
```
idimopoulos: Well I was thinking something like a procedure to extract the pieces of the pattern separately but that takes a bit of time to design and implement and I could not spend time. If you have a suggestion we can discuss on that.

web/modules/custom/rdf_entity/src/Entity/Query/Sparql/SparqlCondition.php
idimopoulos: The problem with this and why it is hard to split is that the sequence of converting the string is tightly attached to the code above (as a meaning). We can split it up but I think it will only make it less readable. If we are to split it, we should create a better design for the whole method.

web/modules/custom/rdf_entity/src/Entity/Query/Sparql/SparqlCondition.php
```diff
@@ -4,47 +4,785 @@
...
+   * @param bool $blank
+   *   Whether or not to be a blank note.
+   *
+   * @return string
+   *   The variable.
+   */
+  protected function toVar($key, $blank = FALSE) {
+    // Deal with field.property as dots are not allowed in var names.
+    $key = str_replace('.', '_', $key);
+    if (strpos($key, '?') === FALSE && strpos($key, '_:') === FALSE) {
```
idimopoulos: I will keep the method as it is because we will need it in the future when we implement the blank nodes. Instead, I will remove the instances where I had a variable be set as a blank node (in the compileExists).

PR #443, opened 2017-04-10
Isaicp 3246: Update dependencies.

idimopoulos:
Among the dependencies are:
- Move to latest locked commit in drupal core 8.3
- Update facets to alpha8, current latest version.
- Update search_api to latest locked version.


composer.json
```diff
@@ -14,7 +14,7 @@
         "drupal/diff": "^1.0@RC",
         "drupal/digital_size_formatter": "~1.0-alpha1",
         "drupal/ds": "^3.0",
-        "drupal/facets": "1.0-alpha7",
+        "drupal/facets": "1.0-alpha8",
```
pfrenssen: It is very unfortunate that this was hardcoded to alpha8 instead of reverting to normal version tracking with `~1.0@alpha` :(

Because of this it has been missed that there has been a new Facets release on Monday :(

Let's update to the latest version.

idimopoulos: Versions of search_api and facets module are now unlocked.

PR #447, opened 2017-04-11
Isaicp 3267

pfrenssen:


pfrenssen:
Work done in scope of this ticket:

* Implemented a `hook_entity_create()` in Joinup Core that sets the publication state for newly created entities.
* In the process of retrieving the correct moderation state the parent collection/solution is loaded to check if it is set to pre-moderation or post-moderation. This requires every community content to be linked to a collection or solution. In the actual website this already works like this but some of our very old tests created a standalone community content and then associated it with a collectolution in a second step. I replaced all of these instances with a single step definition that does the association when the entity is created. This closely matches how it happens in the UI and avoids exceptions if the parent entity cannot be loaded.
* Removed the old step definition to associate a community content with a collution. We will have no need for it any more in the future.
* Replaced all tests that are setting the publication state directly to use workflow states instead. In the UI the publication state can not be controlled directly, so now the tests also closely mimick how it works in the UI.
* We already had good generic code written to interact with workflows in `WorkflowTrait`. I have placed this in a new service {{WorkflowHelper}} with associated interface so that this code can be reused more efficiently and also in procedural hook implementations. I have refactored {{WorkflowTrait}} to leverage the new service, replaced the trait inclusion with the injected service where applicable.
* I did a best-effort search on duplicated code that was also interacting with workflows. This resulted in a number of additional focused small helper methods in the service. I have replaced all instances of duplicated code I found with calls to the service.
* Provided the possibility to use human readable workflow states in Behat tests.
* Added the possibility to reorder root level menu items in the {{@Given (the following )custom page(s) menu structure:}} step. This was needed because the {{demo_june}} tests was creating custom pages separately from a collection and adding them to the menu manually in a second step. In the current version the custom pages are associated with a collection (causing the root level menu items to be created transparently) and then defining the order of the root level items in the second step.
* Added a new {{MissingRelationException}} which is thrown when the parent moderation type is attempted to be retrieved from an entity that doesn't have a parent. This is being thrown in {{JoinupRelationManager::getParentModeration()}}. I added this because I was getting nondescript errors like "accessing property on non-object" during Behat tests.
* The workflow state field was missing from the Document content type in {{JoinupSubContext::getNodeFieldNameFromAlias()}}.
* Various small fixes in the code I was working on: replaced a call to the deprecated `EntityManager` with the correct one, injected services that were being sideloaded, small documentation fixes, ...

web/modules/custom/joinup_news/joinup_news.module
```diff
@@ -53,14 +52,7 @@ function joinup_news_workflow_selector(EntityInterface $entity) {
...
   $relation_manager = \Drupal::service('joinup_core.relations_manager');
-  try {
-    $moderation = $relation_manager->getParentModeration($entity);
-  }
-  catch (MissingRelationException $e) {
-    // Throw a more helpful exception if the news article doesn't have a parent
-    // collection or solution.
-    throw new \Exception('Cannot determine the moderation workflow type on a news article that doesn\'t belong to a collection or solution.');
-  }
+  $moderation = $relation_manager->getParentModeration($entity);
```
pfrenssen: Why remove this? I added this exception specifically because I was getting a nondescript error before (like 'accessing propery on non-object') which cost me time to figure out.

This problem can occur very easily, if you just create any community content in a Behat test without associating it with a collection or solution. Now at least we get a clear exception instead of having to hunt deep into a stack trace.

web/modules/custom/joinup_notification/joinup_notification.module
```diff
@@ -46,7 +46,7 @@ function joinup_notification_entity_predelete(EntityInterface $entity) {
   // The asset distribution rdf entity does not have a workflow so it is
   // excluded from the following check.
   if ($workflow_helper->hasEntityStateField($entity)) {
-    $state_field = \Drupal::service('joinup_core.workflow.helper')->getEntityStateField($entity);
+    $state_field = $this->workflow_helper->getEntityStateField($entity);
```
pfrenssen: This is not going to work, you cannot access class members in procedural code.

web/modules/custom/joinup_news/joinup_news.module
idimopoulos: Oh I see.. it reaches up to MembershipManager::getGroups where it tries to access a method.. Cool. I will revert this one..

pfrenssen:
This looks good now, ready to merge when it comes back from the bot.

PR #450, opened 2017-04-18
Isaicp 3285

pfrenssen:


tests/features/collection/og.feature
```diff
@@ -83,27 +82,3 @@ Feature: Organic Groups integration
     Then I should see the success message "You are no longer a member of Folk Dance and Song Society."
     And I should see the "Join this collection" button
     And the "Folk Dance and Song Society" collection should have 0 members
-
-  @terms
-  Scenario: Edit a Collection
-    Given the following owner:
-      | name                 | type                    |
```
pfrenssen: I moved this test to `tests/features/collection/collection.edit.feature` so that it is in the same place as all the other edit tests. I don't know how this found its way into this completely unrelated test.

pfrenssen:
This is ready for QA if tests come back green.

When testing this the contextual link to edit a collection was not showing up for the collection facilitator. When I was debugging this I noticed that `WorkflowUserProvider` was passing on the anonymous user instead of the facilitator to the workflow guard class. This was denying access since anonymous users do not have the right to make state transitions.

In fact the entire `WorkflowUserProvider` class in unneeded, it duplicates the functionality from the `AccountProxy` service that is included with Drupal core. I have removed it and replaced it with the core service.

pfrenssen:
OK now that this has *magically* turned green again, let's continue this in #459 which is a rebased version of this branch without all the mess trying to figure our why the bot was having a hissy fit.

PR #459, opened 2017-04-20
Isaicp 3285 rebased IS IT GREEN ALREADY????

pfrenssen:
This is the continuation of #450 which had a mysterious failure happening consistently on PHP 7.1.0 only. After a range of exploratory commits and merges to figure out what went wrong, hours of lost work, and sleepless nights, the bug has seemingly vanished without a trace, never seen again by human eyes.

This is a rebased version of #450 that doesn't have the myriad of messy merges and commits to prod the testbot into compliance. It is nice and neat.

# BUT IT NEEDS TO BE GREEN


pfrenssen:
For the person doing QA, please check the comments on #450 for the motivation behind some of the changes made in this PR.

brummbar:
In tests/src/Traits/ContextualLinksTrait.php
```php
\Drupal::currentUser()->setAccount($account);
```

There should be no need to do this. The `checkNamedRoute()` method call has a `$account` parameter.
If a test is failing, it points out that we are not correctly doing access checks in our `hook_ENTITY_TYPE_access()` hooks.
Taking a look at `collection_rdf_entity_access()`:
```php
$allowed_transitions = $entity->field_ar_state->first()->getTransitions();
```
The transitions are retrieved always for the current user. So this code doesn't work when we ask permissions for an user that is not the currently logged in.
Example: we are logged in as a highly privileged user, that can do multiple transitions. Some piece of code checks if another user has permissions to edit a content, in order to take some action (send a mail, for example). The entity access hook will return the wrong value.

Changing the global `$user` should never be taken lightly. That's the reason the old `WorkflowUserProvider` was introduced (as far as I know). So, if we refactor the code to user the `AccountProxy` class (aka `current_user` service) we should take extra care, and we should take into account the passed `$account` in the entity access hooks.
Core has the class `\Drupal\Core\Session\AccountSwitcher`, that allows "safe switching of user accounts by ensuring that session data for one user is not leaked in to others". We also have the `\Drupal\joinup_core\WorkflowHelper` service, which allows to automatically retrieve allowed transitions for a certain entity and user.

Expecially in this class, we have:
```php
if ($user !== NULL) {
  \Drupal::currentUser()->setAccount($user);
}
```
This will leave the wrong user account in place on the active user, and makes the class unusable expect for testing purposes.

If we are going for this `AccountProxy` refactor, we should remove all direct accesses to the `getTransitions()` in the entity access hooks, and replace them with
`\Drupal\joinup_core\WorkflowHelper::getAvailableStates()`. In this function, we should inject the dependency of the `AccountSwitcher` service. Then, before doing the retrieval of the states, we call:
```php
$this->accountSwitcher->switchTo($user);
```
and after we do:
```php
$this->accountSwitcher->switchBack();
```

pfrenssen:
OK you're right, having this code in the service is a bad idea since it can then be called from the Drupal side. This is only intended to be run in scope of doing API calls on the bootstrapped Drupal instance in the Behat test runner, never in scope of a Drupal request.

I will move this code back to the Behat test trait.

pfrenssen:
> Changing the global $user should never be taken lightly. That's the reason the old `WorkflowUserProvider` was introduced (as far as I know).

`WorkflowUserProvider` was introduced as a way to be able to change the current user in the workflow tests, see for example `NodeWorkflowTestBase`. The `::setUser()` method was only called in tests. It was not realized by the people who worked on those tests that Drupal core already provides a way to change the current user in tests without having to keep the state in a self written object. The tests in Drupal core also rely heavily on this feature.

pfrenssen:
I have moved the code that changes the current user back to the trait, and added a safeguard that will throw an exception if the code is run outside of the scope of a Behat test. This will ensure that if a developer mistakenly imports this Behat trait in Drupal code that they will be alerted that this is a bad idea.

pfrenssen:
The last commit failed due to a random failure. This is tracked in https://webgate.ec.europa.eu/CITnet/jira/browse/ISAICP-3330

PR #466, opened 2017-04-25
ISAICP-3316: Do not throw an exception when a parent is not found. Th…

idimopoulos:
…is is breaking the admin area. Revoke access if no parent is found.


pfrenssen:
The exception was useful to get good feedback when trying to create community content without a parent, but this solution is too simple to pass up. Approved!

PR #468, opened 2017-04-26
3292: 'Organisation Type' in Event.

claudiu-cristea:


web/modules/custom/joinup_event/config/install/field.storage.node.field_organisation_type.yml
```diff
@@ -3,14 +3,36 @@ status: true
...
+  allowed_values:
+    -
+      value: academia
+      label: Academia
+    -
+      value: non_profit
+      label: 'Non-profit'
+    -
+      value: other
+      label: Other
```
idimopoulos: Shouldn't that be the last option?

idimopoulos: It's in the definition of the ticket. ok.

PR #471, opened 2017-04-27
Revert ISAICP-3213

claudiu-cristea:


tests/features/solution/add_solution.feature
```diff
@@ -82,9 +82,6 @@ Feature: "Add solution" visibility options.
     Then I should see the heading "Add Solution"
     And the following fields should be present "Title, Description, Documentation, Logo, Banner"
     And the following fields should not be present "Groups audience, Other groups, Current workflow state"
-    # Regression test for the default selected option for the status field.
-    # @see: https://webgate.ec.europa.eu/CITnet/jira/browse/ISAICP-3213
-    And the radio button "N/A" from field "Status" should be selected
```
pfrenssen: This should not be removed, this is a regression test for a bug that is fixed. We need this to ensure that the radio button is selected.

claudiu-cristea: This should be fixed in Core

pfrenssen:
I was mistaken about this issue. I thought this was an older fix that was accepted in a previous sprint. Turns out it was just accepted yesterday. Since this is still in the current sprint I think it is fine to revert this and reopen the issue to find a better solution.

PR #474, opened 2017-05-03
Isaicp 3285 v3

pfrenssen:


web/modules/custom/asset_release/tests/src/Functional/AssetReleaseWorkflowTest.php
```diff
@@ -191,8 +190,7 @@ public function testWorkflow() {
         $content->save();
 
         // Override the user to be checked for the allowed transitions.
-        $this->setCurrentUser($this->{$user_var});
-        $actual_transitions = $content->field_isr_state->first()->getTransitions();
+        $actual_transitions = $this->workflowHelper->getAvailableTransitions($content, $this->{$user_var});
```
pfrenssen: Minor remark: it is not needed to wrap this in curly braces, `$this->$user_var` is legal PHP.

web/modules/custom/joinup_core/src/WorkflowHelperInterface.php
```diff
@@ -11,6 +12,51 @@
...
+   *
+   * @param \Drupal\Core\Entity\FieldableEntityInterface $entity
+   *   The entity with the states.
+   * @param \Drupal\Core\Session\AccountInterface|null $user
+   *   The account interface object. Can be left empty.
+   *
+   * @return array
+   *   An array of transition state labels.
+   */
+  public function getAvailableStatesLabels(FieldableEntityInterface $entity, AccountInterface $user = NULL);
```
pfrenssen: Now that we are touching this, we can rename these `$user` to the more accepted `$account` in all these instances.

web/modules/custom/joinup_core/src/WorkflowHelper.php
```diff
@@ -5,14 +5,90 @@
...
+   * {@inheritdoc}
+   */
+  public function getAvailableTransitions(FieldableEntityInterface $entity, AccountInterface $user = NULL) {
+    // Set the current user so that states available are retrieved for the
+    // specific account.
+    $account_switched = FALSE;
+    if ($user !== NULL && $user->id() !== $this->currentUser->id()) {
+      $this->accountSwitcher->switchTo($user);
+      $account_switched = TRUE;
+    }
```
pfrenssen: The proper solution would be to be able to pass the `$account` to the State Machine when retrieving the transitions. This switching of the current user is rather tricky. Let's put a `@todo` here that explains this is a workaround, with the instructions that it should be removed once the upstream issue is fixed: https://www.drupal.org/node/2776969

pfrenssen:
Thanks!

PR #475, opened 2017-05-04
Isaicp 3194: Proper error display for event datetime error messages.

idimopoulos:


tests/features/joinup_event/add_event.solution.feature
```diff
@@ -62,6 +62,14 @@ Feature: "Add event" visibility options.
       | Location    | Rue Belliard, 28                      |
     And I fill in "Start date" with the date "2018-08-29"
     And I fill in "Start date" with the time "23:59:00"
+
+    # Regression test.
+    # @see: https://webgate.ec.europa.eu/CITnet/jira/browse/ISAICP-3194
+    And I fill in "End date" with the date "2018-08-30"
+    And I press "Save as draft"
+    Then I should see the error message "Both date and time fields should be filled in the datetime field"
```
pfrenssen: There are some problems with this error message:

* This error is not 100% clear. The way it is worded now it can be interpreted as 'Both "date and time fields" should be filled in'. The user might be confused since the start date and end date are in fact filled in. What is actually meant is that both the date field and the time field should be filled in. I propose to change this to the following which is not ambiguous: "The date and time should both be entered in the X field.".
* Error messages should end in a period.
* Calling the field the "datetime" field is confusing, this field is not called that way in the interface. It should be the "End date" field.

PR #490, opened 2017-05-11
Minor update in documentation

kalinchernev:
Copy&paste didn't work as it was not visible that these are 2 commands.


pfrenssen:
Thanks!

PR #497, opened 2017-05-11
3350: Allow multiple values for File URL

claudiu-cristea:
- The code of https://www.drupal.org/project/file_url should be reviewed too.
- Known Issues: see https://www.drupal.org/project/issues/file_url


web/modules/custom/rdf_entity/tests/src/Traits/RdfEntityUtilityTrait.php
```diff
@@ -0,0 +1,63 @@
+<?php
+
+namespace Drupal\Tests\rdf_entity\Traits;
+
+/**
+ * Provide some utility methods to be used in tests.
+ */
+trait RdfEntityUtilityTrait {
```
idimopoulos: I think this should better be named simply as "EntityUtilityTrait".
It is already namespaced and is used to load an entity of any entity type by label.
It is confusing as a class name.

PR #521, opened 2017-05-19
Isaicp 2922: Provide a custom 404 page.

idimopoulos:


claudiu-cristea:
As general remak. I wonder why we are writing so much custom code when we have nice contrib modules that are doing the same. There's [Page Manager](https://www.drupal.org/project/page_manager) doing this job without needing to create custom routes and controllers. At the end it's just config that we need to maintain. This is true also for other components of the site where we write a lot of, hard maintainable, custom code, like sending notifications -- that is pure Rules module job

PR #526, opened 2017-05-23
ISAICP-2892: Fix local action buttons

tom-waat:
Merge ISAICP-2892 into develop

### DIFF
--- a/web/themes/joinup/prototype/sass/components/_button.sass
+++ b/web/themes/joinup/prototype/sass/components/_button.sass
@@ -49,6 +49,9 @@
   z-index: 10
   position: relative
 
+.button__container
+  margin: 10px 0
+
 // Modifiers
 // --------------------------------------------------
 .button--blue
--- a/web/themes/joinup/templates/block/block--local-actions-block.html.twig
+++ b/web/themes/joinup/templates/block/block--local-actions-block.html.twig
@@ -7,6 +7,6 @@
 #}
 {% block content %}
   {% if content %}
-    <nav class="tabs tabs--vertical mdl-tabs__tab-bar">{{ content }}</nav>
+    <nav class="button__container center">{{ content }}</nav>
   {% endif %}
 {% endblock %}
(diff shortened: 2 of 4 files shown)


web/themes/joinup/prototype/sass/components/_button.sass
```diff
@@ -49,6 +49,9 @@
   z-index: 10
   position: relative
 
+.button__container
```
kalinchernev: Feels a bit hacky to be honest. I don't mind it. But if I were you, i'd make a simple mixin or helper to be able to re-use this rule for layouts and similar with a parameter of the spacing.

For example, there's already a quite similar scenario:
```
// Dependiencies
// --------------------------------------------------
.page__title-wrapper
  .details
    margin: 10px 0
    +breakpoint(mobile)
      margin: 0
```

web/themes/joinup/templates/block/block--local-actions-block.html.twig
```diff
@@ -7,6 +7,6 @@
 #}
 {% block content %}
   {% if content %}
-    <nav class="tabs tabs--vertical mdl-tabs__tab-bar">{{ content }}</nav>
```
kalinchernev: There's a question in the ticket: `Fixing this in the theme will be simple, but maybe we will need to do some digging to find the exact reason why this mistake was initially made.`
You might want to reply to that it seems there's no specific business reasons: this was introduced quite early in general styling of elements: https://webgate.ec.europa.eu/CITnet/jira/browse/ISAICP-2565
It seems a safe fix to me.

web/themes/joinup/templates/navigation/menu-local-action.html.twig
```diff
@@ -10,11 +10,4 @@
...
 #}
-{%
-  set classes = [
-    'tab',
-    'tab--horizontal',
-    'mdl-tabs__tab',
-  ]
-%}
-<li{{ attributes.addClass(classes) }}>{{ link }}</li>
+<span{{ attributes }}>{{ link }}</span>
```
kalinchernev: Made a quick double-check on local action links when joinup theme is admin theme.
For this specific task, no problems found in content, people and structure-related management, so it's ok.

content joinup

PR #527, opened 2017-05-23
Isaicp 2374: Authenticated user's homepage

idimopoulos:


tests/features/user/homepage.feature
```diff
@@ -0,0 +1,29 @@
...
+      | title                     | body                      | policy domain     | collection        | state     |
+      | The Danger of the Bridges | The Danger of the Bridges | Finance in EU     | The Sacred Future | validated |
+      | Girl in the Dreams        | Girl in the Dreams        | Supplier exchange | Boy of Courage    | validated |
+    And the following collection user memberships:
+      | collection        | user         | roles |
+      | The Sacred Future | Henry Austin |       |
+
+    When I am logged in as "Henry Austin"
+    And I am on the homepage
+    Then I should see the "The Danger of the Bridges" tile
```
brummbar: Can you put here two lines to explain why I should see one and not the other?

web/modules/custom/joinup_search/config/install/search_api.index.collections.yml
```diff
@@ -95,6 +95,13 @@ field_settings:
       fields:
         - 'entity:node/field_keywords'
         - 'entity:rdf_entity/field_keywords'
+  aggregated_field:
```
brummbar: I think this empty aggregated field can be removed.

web/profiles/joinup/joinup.profile
```diff
@@ -306,3 +306,16 @@ function joinup_install_tasks_alter(&$tasks, $install_state) {
...
 }
+
+/**
+ * Implements hook_preprocess_block().
+ *
+ * Prepares the front page block to be displayed as a content listing.
+ *
+ * @see: joinup_theme_preprocess_block.
+ */
+function joinup_preprocess_block(&$variables) {
```
brummbar: Move this to the theme.

web/profiles/joinup/src/Plugin/Block/FrontpageBlock.php
```diff
@@ -0,0 +1,188 @@
...
+
+/**
+ * Provides a 'FrontpageBlock' block.
+ *
+ * @Block(
+ *  id = "frontpage_block",
+ *  admin_label = @Translation("Frontpage block"),
+ * )
+ */
+class FrontpageBlock extends BlockBase implements ContainerFactoryPluginInterface {
```
brummbar: I would use a more explicative name for this block. The scope of this block is to show recommended content to the user, and it doesn't have any technical limitation regarding being in the homepage. So, I would call it something like `RecommendedContentBlock`.

web/profiles/joinup/src/Plugin/Block/FrontpageBlock.php
```diff
@@ -0,0 +1,188 @@
...
+   *
+   * @param array $groups
+   *   The user's memberships.
+   *
+   * @return array
+   *   An array of rows to render.
+   */
+  protected function getContentFromMemberships(array $groups) {
+    $rdf_entities = isset($groups['rdf_entity']) ? $groups['rdf_entity'] : [];
+    $cids = array_map(function ($rdf_entity) {
```
brummbar: As discussed yesterday, for some special cases we will be hitting the query size limit. Take only the first 100 groups to avoid breaking for our test roles, that have hundreds of groups. Do not set any message (see user profile).

PR #534, opened 2017-05-31
Issue #ISAICP-3443: Disable attachment test until a proper fix for the

sandervd:
performance issue is in place.


idimopoulos:
The migrations are done. This is not valid anymore.

PR #538, opened 2017-06-01
ISAICP-3352

tom-waat:
Merge into develop

### DIFF
--- a/web/themes/joinup/js/form.js
+++ b/web/themes/joinup/js/form.js
@@ -45,6 +45,9 @@
       $(context).find('form').once('ajaxReload').each(function () {
         $(document).ajaxComplete(function (event, xhr, settings) {
           componentHandler.upgradeAllRegistered();
+          $('.mdl-js-checkbox').each(function (index, element) {
+            element.MaterialCheckbox.updateClasses_();
+          })
         });
       });
     }
--- a/web/themes/joinup/prototype/sass/_mixins.sass
+++ b/web/themes/joinup/prototype/sass/_mixins.sass
@@ -72,7 +72,3 @@
     -ms-transform: translate(0, -50%)
     transform: translate(0, -50%)
 
-// Margin for form elements
-=form-margin
-  margin-bottom: 25px
-
--- a/web/themes/joinup/prototype/sass/_mixins.sass
+++ b/web/themes/joinup/prototype/sass/_mixins.sass
@@ -72,3 +72,7 @@
     -ms-transform: translate(0, -50%)
     transform: translate(0, -50%)
 
+// Margin for form elements
+=form-margin
+  margin-bottom: 25px
+
--- a/web/themes/joinup/prototype/sass/components/_form-table.sass
+++ b/web/themes/joinup/prototype/sass/components/_form-table.sass
@@ -23,14 +23,3 @@ thead th.form-table__header
 .form-table__cell--icon-container
   width: 40px
   padding: 0
-
-#password-policy-status
-  @extend .form-wrapper
-  table
-    th:first-child, td:first-child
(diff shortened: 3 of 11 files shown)


web/themes/joinup/prototype/sass/components/_form-table.sass
```diff
@@ -23,3 +23,14 @@ thead th.form-table__header
 .form-table__cell--icon-container
   width: 40px
   padding: 0
+
+#password-policy-status
```
kalinchernev: Please re-consider using ids, try to use classes.
If you cannot use classes, please move this into drupal hacks.

web/themes/joinup/prototype/sass/components/_form-table.sass
```diff
@@ -9,7 +9,7 @@
   margin-left: 25px
 
 .form-table__cell
-  padding: 0 15px
+  padding: 15px
```
kalinchernev: It's always better to be specific, than loose, it'd recommend leaving this specific.

web/themes/joinup/prototype/sass/components/_form-table.sass
```diff
@@ -23,3 +23,14 @@ thead th.form-table__header
 .form-table__cell--icon-container
   width: 40px
   padding: 0
+
+#password-policy-status
+  @extend .form-wrapper
```
kalinchernev: Please try to find a way to use a mixin. [More details](https://csswizardry.com/2014/11/when-to-use-extend-when-to-use-a-mixin/).

web/themes/joinup/js/form.js
```diff
@@ -45,6 +45,9 @@
       $(context).find('form').once('ajaxReload').each(function () {
         $(document).ajaxComplete(function (event, xhr, settings) {
           componentHandler.upgradeAllRegistered();
+          $('.mdl-js-checkbox').each(function (index, element) {
```
kalinchernev: Please remove, it's not related to this pull request.

tom-waat: It is accidentally added commit from ISAICP-3237, could we leave it as it is to avoid merge conflicts?

web/themes/joinup/prototype/sass/_mixins.sass
```diff
@@ -72,3 +72,7 @@
     -ms-transform: translate(0, -50%)
     transform: translate(0, -50%)
 
+// Margin for form elements
+=form-margin
```
kalinchernev: Please improve a bit:
- you can make the size a parameter
- the naming can be more generic

You can get inspiration from: https://github.com/digitaledgeit/sass-spacing

tom-waat: But in this case we need only one margin value for form elements, we don't need flexible mixin. We only need to be sure that all form elements have the same margin, so maybe I could create variable like $form-margin: 25px, and use margin-bottom: $form-margin . Would it be ok?

web/themes/joinup/js/form.js
kalinchernev: I think you can revert a commit. It'd be easier than rebase.
This one is related to ISAICP-3237 which is in progress still.

web/themes/joinup/prototype/sass/_mixins.sass
kalinchernev: Yes, it makes more sense to have a variable and use it in the form component.
My main concern here was that `form-margin` mixin is too specific and not obviously re-usable. A variable is better, please make as you propose.

tom-waat: Ok, I added the variable.

PR #544, opened 2017-06-02
ISAICP-3396

tom-waat:
Merge into develop


web/themes/joinup/templates/field/field--with-label.html.twig
```diff
@@ -0,0 +1,54 @@
...
+      {% for item in items %}
+        {{ item.content }}
+      {% endfor %}
+  {% else %}
+    {% for item in items %}
+      {{ item.content }}
+    {% endfor %}
+  {% endif %}
+{% else %}
+    <div{{ title_attributes.addClass('form__label') }}>{{ label }}</div>
```
kalinchernev: Is it that the template is added for 1 class of `form__label`?

tom-waat: Yes. It is indeed similar to field--no-wrappers, but won't affect other fields which use field--no-wrappers. Because of that I created new template. I cannot also use preprocess function there, because template is added in manage display interface.

web/themes/joinup/joinup_theme.theme
```diff
@@ -940,3 +940,40 @@ function joinup_theme_form_comment_form_alter(&$form, FormStateInterface $form_s
...
+function joinup_theme_preprocess_table__file_formatter_table(&$variables) {
+  if (!empty($variables['rows'])) {
+    foreach ($variables['rows'] as $index => $row) {
+      /** @var \Drupal\file\Entity\File $file */
+      $file = $variables['rows'][$index]['cells'][0]['content']['#file'];
+      $file_icon = pathinfo($file->getFilename(), PATHINFO_EXTENSION);
+
+      // List of available icons.
+      $available_icons = [
+        'doc',
```
kalinchernev: These are ok. Strangely though, the prototype (http://localhost:3000/section/2.8) didn't show the icons, so I had to manually check each one.

PR #548, opened 2017-06-06
ISAICP-3438: Fix download button on documents tile

tom-waat:
Merge ISAICP-3438 into develop

### DIFF
--- a/web/themes/joinup/templates/content/node--document--listing-item.html.twig
+++ b/web/themes/joinup/templates/content/node--document--listing-item.html.twig
@@ -88,10 +88,12 @@
         <div class="value gray small truncated">{{ file_name }}</div>
       {% endif %}
     </div>
+    {%  if extension %}
     <div class="listing__field">
       <div class="label gray">{{ 'Type'|t }}:</div>
       <div class="value bold">{{ extension|upper }}</div>
     </div>
+    {%  endif %}
     {% if file_size %}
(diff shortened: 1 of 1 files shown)


web/themes/joinup/templates/content/node--document--listing-item.html.twig
```diff
@@ -107,7 +107,7 @@
     {% endif %}
     <div class="listing__field listing__stats mdl-card__actions listing__action-link__wrapper action-link__wrapper">
       <a href="{{ url }}" class="listing__action-link action-link">{{ 'Details'|t }}</a>
-      <a href="{{ content.field_file|render|striptags|trim }}" class="listing__action-link action-link">{{ 'Download'|t }}</a>
+      <a href="{{ content.field_file.0|render|striptags|trim }}" class="listing__action-link action-link">{{ 'Download'|t }}</a>
```
kalinchernev: Hi @tom-waat the fix is handling the case for 1 file.
One suggestion: use a small preprocess as [this one](https://github.com/ec-europa/joinup-dev/blob/e08f22e5dae4f7e5b39d480686d33e1d11e3863b/web/themes/joinup/joinup_theme.theme#L919) to pass a the var in a bit better way. The current fix is ok for the moment, but is a bit cheese and can cause confusion later if more files are to be displayed.

PR #553, opened 2017-06-07
Isaicp 3149

sandervd:


web/modules/custom/adms_validator/src/Form/AdmsValidatorForm.php
```diff
@@ -0,0 +1,224 @@
...
+      $form_state->setError($form['adms_file'], 'Please upload a valid RDF file.');
+      return;
+    }
+    if (!$this->storeInGraph($file)) {
+      $form_state->setError($form['adms_file'], 'The provided file is not a valid RDF file.');
+      return;
+    }
+    // Delete the uploaded file from disk.
+    $file->delete();
+    $form_state->addBuildInfo('validation_errors', $this->getValidationErrors());
```
brummbar: We can use the `FormState::set()` to keep data in the storage or even a property in the form class, like the entity form does.

web/modules/custom/adms_validator/src/Form/AdmsValidatorForm.php
```diff
@@ -0,0 +1,224 @@
...
+    $form['info'] = [
+      '#markup' => $this->t('This validator uses the <a href="https://github.com/SEMICeu/adms-ap_validator">SEMIC ADMS-AP ruleset</a>.'),
+    ];
+    $form['submit'] = [
+      '#type' => 'submit',
+      '#value' => $this->t('Upload'),
+      '#button_type' => 'primary',
+    ];
+    $info = $form_state->getBuildInfo();
+    if (!empty($info['validation_errors'])) {
```
brummbar: Let's add some messages to the user when there are no validation errors, as we discussed.

PR #554, opened 2017-06-07
ISAICP-3449: Style members collection page

tom-waat:
Merge ISAICP-3449 into develop

### DIFF
--- a/web/themes/joinup/joinup_theme.theme
+++ b/web/themes/joinup/joinup_theme.theme
@@ -535,6 +535,13 @@ function joinup_theme_preprocess_image(&$variables) {
 }
 
 /**
+ * Implements hook_preprocess_user().
+ */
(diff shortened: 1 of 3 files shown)


web/themes/joinup/joinup_theme.theme
```diff
@@ -193,6 +192,7 @@ function joinup_theme_preprocess_block(&$variables) {
       case 'view.collections.page_1':
       case 'view.solutions.page_1':
       case 'view.content_overview.page_1':
+      case 'entity.rdf_entity.member_overview':
```
kalinchernev: Small problem with height of image when the tile is used on the user profile:
members joinup

Without the height property:
members joinup 1

tom-waat: Ok, it is fixed for all tiles. Currently image is never cropped, only scaled.

PR #561, opened 2017-06-09
ISAICP-3033: MDL style for the throbber

piotrsmykaj:


kalinchernev:
Hi @piotrsmykaj  @tom-waat,
@brummbar noticed that the ajax rule should be more generic to elements which are not only forms.
Here's a patch with possible suggestion.
[patch.txt](https://github.com/ec-europa/joinup-dev/files/1068055/patch.txt)
The dependency is taken our as Drupal render system adds it when necessary automatically.

PR #563, opened 2017-06-09
ISAICP-3358: Reorder and style mulifile field

tom-waat:
Merge ISAICP-3358 into develop


kalinchernev:
The order is ok, I placed a comment on your question about the design specifics.

PR #564, opened 2017-06-12
Isaicp 3169: Split published and unpublished index

idimopoulos:


web/modules/custom/joinup_core/src/Plugin/Block/UserUnpublishedBlock.php
```diff
@@ -0,0 +1,283 @@
...
+   */
+  protected function getResultEntities(ResultSetInterface $result) {
+    $results = [];
+    /* @var $item \Drupal\search_api\Item\ItemInterface */
+    foreach ($result->getResultItems() as $item) {
+      try {
+        $entity = $item->getOriginalObject()->getValue();
+        $entity_id = $item->getOriginalObject()->getValue()->id();
+        if ($entity instanceof NodeInterface) {
+          $entity = $this->getLatestContentRevision($entity_id);
```
brummbar: Use the `RevisionManager` service.

web/modules/custom/joinup_core/src/Plugin/Block/UserUnpublishedBlock.php
```diff
@@ -0,0 +1,283 @@
...
+        }
+        elseif ($entity instanceof RdfInterface) {
+          $entity = $this->getDraftRdf($entity_id);
+        }
+        else {
+          throw new \Exception("Only nodes and Rdf entities should be loaded.");
```
brummbar: Single quotes.

web/modules/custom/joinup_core/src/Plugin/Block/UserUnpublishedBlock.php
```diff
@@ -0,0 +1,283 @@
...
+  /**
+   * Loads the latest revision on a node entity.
+   *
+   * @param int $entity_id
+   *   The content id.
+   *
+   * @return \Drupal\node\NodeInterface
+   *   The loaded node.
+   */
+  protected function getLatestContentRevision($entity_id) {
```
brummbar: Same as above, this code is duplicated from RevisionManager.

web/modules/custom/joinup_core/src/Plugin/Block/UserUnpublishedBlock.php
```diff
@@ -0,0 +1,283 @@
...
+   */
+  public function getCacheContexts() {
+    return Cache::mergeContexts(parent::getCacheContexts(), ['og_role']);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getCacheTags() {
+    return Cache::mergeTags(parent::getCacheTags(), ['rdf_entity_list', 'node_list']);
```
brummbar: Same as above.

web/modules/custom/joinup_search/config/install/search_api.index.published.yml
```diff
@@ -60,7 +60,7 @@ field_settings:
         - 'entity:rdf_entity/og_audience'
         - 'entity:rdf_entity/field_isr_is_version_of'
   entity_author:
-    label: Author
+    label: 'Aggregated field'
```
brummbar: This is a regression, it should say author. I think this is caused by an unlucky merge.

web/modules/custom/joinup_search/src/Plugin/search_api/processor/JoinupEntityLatestRevision.php
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * Ensures that the item to be indexed is the latest version.
+ *
+ * @SearchApiProcessor(
+ *   id = "joinup_entity_latest_revision",
+ *   label = @Translation("Joinup entity latest revision"),
+ *   description = @Translation("Ensures that the version of the entity is the latest revision."),
+ *   stages = {
+ *     "alter_items" = -20,
```
brummbar: Cool.

web/modules/custom/joinup_search/src/Plugin/search_api/processor/JoinupEntityLatestRevision.php
```diff
@@ -0,0 +1,111 @@
...
+      $container->get('entity_type.manager'),
+      $container->get('state_machine_revisions.revision_manager')
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function supportsIndex(IndexInterface $index) {
+    $supported_entity_types = ['node', 'rdf_entity'];
```
brummbar: Cool.

web/modules/custom/joinup_search/src/Plugin/search_api/processor/JoinupEntityLatestRevision.php
```diff
@@ -0,0 +1,111 @@
...
+            $original_object->setValue($latest);
+            $item->setOriginalObject($original_object);
+            $items[$item_id] = $item;
+          }
+        }
+      }
+      elseif ($object instanceof RdfInterface) {
+        if ($object->isPublished()) {
+          /** @var \Drupal\rdf_entity\Entity\RdfEntitySparqlStorage $rdf_storage */
+          $rdf_storage = $this->entityTypeManager->getStorage('rdf_entity');
```
brummbar: What about creating some helper to do this? Maybe at some point we should.

web/modules/custom/joinup_search/src/Plugin/search_api/processor/JoinupEntityPublishedRevision.php
```diff
@@ -0,0 +1,111 @@
...
+   * @var \Drupal\Core\Entity\EntityTypeManagerInterface
+   */
+  protected $entityTypeManager;
+
+  /**
+   * The state machine revisions manager service.
+   *
+   * @var \Drupal\state_machine_revisions\RevisionManagerInterface
+   */
+  protected $revisionManager;
```
brummbar: The revision manager is not used in this class actually.

web/profiles/joinup/config/install/block.block.rdfentitiesinproposedstate.yml
```diff
@@ -0,0 +1,32 @@
+langcode: en
+status: true
+dependencies:
+  module:
+    - joinup_core
+    - user
+  theme:
+    - joinup_theme
+id: rdfentitiesinproposedstate
```
brummbar: It's minor, but I have noticed that block ids can be improved manually. I would try to use underscores here: `rdf_entities_in_proposed_state`.

web/profiles/joinup/config/install/block.block.unpublishedcontentofgroup.yml
```diff
@@ -0,0 +1,35 @@
+langcode: en
+status: true
+dependencies:
+  module:
+    - ctools
+    - joinup_core
+  theme:
+    - joinup_theme
+id: unpublishedcontentofgroup
```
brummbar: Same here, some underscores to make it readable.

build.solr.xml
```diff
@@ -101,24 +101,41 @@
...
-                <copy todir="${solr.core.dir}/conf" overwrite="true">
+                <copy todir="${solr.core.published.dir}/conf" overwrite="true">
                     <fileset dir="${solr.config.dir}" />
                 </copy>
-                <!-- Restart Solr so our configuration changes are picked up. -->
-                <phingcall target="restart-apache-solr" />
             </else>
         </if>
+        <if>
+            <available file="${solr.core.unpublished.dir}" type="dir" property="solr.core.unpublished.dir.available" />
```
brummbar: This whole task is duplicated, can we extract the definition and add parameters to it so we can call it with the two different sets?

tests/features/custom_page/custom_page.view.feature
```diff
@@ -21,8 +21,9 @@ Feature:
...
     But I should not see the "About means" tile
 
-    # Collection facilitators can see the content created inside their collections.
+    # Collection facilitators can see the content created inside their collections
+    # but not in the search because the content are not indexed.
     When I am logged in as a facilitator of the "Drafty things" collection
     And I enter "About" in the header search bar and hit enter
     Then I should see the "About places" tile
-    And I should see the "About means" tile
+    And I should not see the "About means" tile
```
idimopoulos: This is because the test was looking for the item in the search page. After this ticket, the search page should only show items indexed in the published index.
I will extend the test so that the user checks as well the collection overview page and finds it.

web/modules/custom/collection/collection.module
```diff
@@ -144,15 +143,15 @@ function collection_entity_bundle_field_info_alter(&$fields, EntityTypeInterface
...
+  if ($entity->getEntityTypeId() !== 'rdf_entity' || $entity->bundle() !== 'collection') {
     return;
   }
   collection_entity_update($entity);
 
   // When a new collection is proposed, give the user the 'group owner' role.
   // This will not occur if the user is the anonymous, i.e. the entity was
   // created by some batch process or some importing.
-  if (Og::isGroup($entity->getEntityTypeId(), $entity->bundle()) && !\Drupal::currentUser()->isAnonymous()) {
+  if (Og::isGroup($entity->getEntityTypeId(), $entity->bundle()) && !empty($entity->getOwner())) {
```
brummbar: Ok the membership is not found (a few lines later) when the user is anonymous, so the code bails out. It works.

web/modules/custom/joinup_core/src/Plugin/Block/EntityUnpublishedBlock.php
```diff
@@ -0,0 +1,221 @@
...
+   * Loads the latest revision on an entity.
+   *
+   * @param int $entity_id
+   *   The content id.
+   *
+   * @return \Drupal\node\NodeInterface
+   *   The loaded node.
+   */
+  protected function getLatestRevision($entity_id) {
+    $storage = $this->entityTypeManager->getStorage('node');
```
idimopoulos: I will dependency injection do

web/modules/custom/joinup_core/src/Plugin/Block/EntityUnpublishedBlock.php
```diff
@@ -0,0 +1,221 @@
...
+   */
+  public function getCacheContexts() {
+    return Cache::mergeContexts(parent::getCacheContexts(), ['og_role']);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getCacheTags() {
+    return Cache::mergeTags(parent::getCacheTags(), ['node_list']);
```
idimopoulos: buh..

web/modules/custom/joinup_search/src/Plugin/search_api/processor/JoinupEntityLatestRevision.php
idimopoulos: This is the only instance in this class where the entity type manager is used. If you are referring to the complete 'get rdf version' maybe we should as I struggled as well but I think we can leave it for now. We have created something similar for the delete but I don't want to make this addition right now in the scope of such a big ticket.

build.solr.xml
idimopoulos: I left it like this only because I would have to put the definitions in separate variables in order to group them and then if in the future we 'd have to add extra data in only one of the cores, we would have to revert simple work that doesn't hurt us having a bit duplicated. It is just 2 commands working in different directories etc.

PR #582, opened 2017-06-19
Double-check the presence of global Drupal - noissue

kalinchernev:
The changes not related to the global Drupal check are coming from [prettier](https://github.com/prettier/prettier). Please tell me if I am to revert them.


kalinchernev:
WIP, has to be improved.

PR #583, opened 2017-06-19
ISAICP-3028: Remove 3th party fonts

tom-waat:
Merge into develop.


web/themes/joinup/joinup_theme.libraries.yml
```diff
@@ -25,13 +25,3 @@ form-components:
     - core/jquery.ui.slider
     - core/drupal
     - core/drupalSettings
-
-fonts:
```
kalinchernev: Please also remove the records from the info file as well:
```
  - joinup_theme/icons
  - joinup_theme/fonts
```

web/themes/joinup/prototype/sass/base/_fontfaces.sass
```diff
@@ -0,0 +1,98 @@
+// Typography
```
kalinchernev: Please include all types of fonts.
As @brummbar suggested, you can use: https://google-webfonts-helper.herokuapp.com/fonts

web/themes/joinup/prototype/sass/components/_icon.sass
```diff
@@ -35,7 +35,7 @@
 //
 // Styleguide 2.8
 
-$waat-icons-font-path: '../images/fonts/waat-icons'
+$waat-icons-font-path: '../images/fonts/waat-icons/waat-icons'
```
kalinchernev: Please move `fonts` in root of theme. Fonts are same asset as css, js, etc. but they are surely not images.

web/themes/joinup/prototype/sass/base/_fontfaces.sass
tom-waat: I already used this tool and included all fonts, which were previously fetched fromhttps://fonts.googleapis.com

kalinchernev: google webfonts helper

There's an option to include files in addition to the `woff` and `woff2`, which are needed in case the [browser wouldn't support them](https://creativemarket.com/blog/the-missing-guide-to-font-formats).

The google apis has a mechanisms that delivers only what is necessary, but since we don't have it, it's necessary to include all types.

tom-waat: Ok, I added fonts for older browsers.

PR #587, opened 2017-06-20
ISAICP-2338: Print option

piotrsmykaj:


web/themes/joinup/prototype/sass/print.sass
```diff
@@ -0,0 +1,22 @@
+@media print
```
kalinchernev: This is fine, though I'd suggest you the following:

- make a mixin:
```
@mixin hidden-print() {
  @media print {
    display: none !important;
  }
}
```
- hide in a per-component basis:
```
@include ecl-hidden-print();
```
Optionally, you can make also a class like `.hidden-print` or something similar using this mixin.

Main goal: make your solution more flexible for others.

kalinchernev:
Hi @piotrsmykaj thanks for the updates!
Please also remove the unrelated code from the pull request, i.e. diff with #589.
At the moment, #589 is blocking this one, I wouldn't merge this one before #589.

piotrsmykaj:
Hi @kalinchernev. Now it should be ok, please check.

PR #589, opened 2017-06-20
ISAICP-3457: Style the homepage tiles

piotrsmykaj:


web/themes/joinup/prototype/sass/components/_featured-lg.sass
```diff
@@ -61,6 +62,11 @@
       width: 200px
       margin-right: 40px
 
+    a
```
kalinchernev: What is the reason of styling this rule?

piotrsmykaj: zrzut ekranu 2017-06-20 o 22 12 47
It's overridden default link color for this attached items.

kalinchernev: Thanks!

PR #593, opened 2017-06-22
Issue #ISAICP-3530: Improve error reporting in ADMS-AP validator.

sandervd:


brummbar:
Tests are green, they are marked as red because of an unrelated error log.

PR #606, opened 2017-06-28
ISAICP-3017: Fix file types and settings on Format and Representation Technique fields

idimopoulos:


brummbar:
Tests are red due issues in CPHP. Behat tests and one of the two phpunit are green, so merging.

PR #613, opened 2017-06-29
ISAICP-3551: Style footer menu

tom-waat:
Merge ISAICP-3551 into develop


brummbar:
Tests look red but they are green.

PR #625, opened 2017-07-05
Isaicp 3490

piotrsmykaj:


kalinchernev:
Small problem with positioning:

offset_more_button

web/themes/joinup/joinup_theme.theme
```diff
@@ -406,6 +417,13 @@ function joinup_theme_preprocess_form_element(&$variables) {
 }
 
 /**
+ * Implements hook_preprocess_HOOK() for form-element--textfield--search-bar.html.twig.
+ */
+function joinup_theme_preprocess_form_element__textfield__search_bar(&$variables) {
```
kalinchernev: Could you please add a small explanation of this operation?

web/themes/joinup/joinup_theme.theme
```diff
@@ -740,6 +762,10 @@ function joinup_theme_theme_suggestions_input_alter(array &$suggestions, array $
  */
 function joinup_theme_theme_suggestions_form_element_alter(array &$suggestions, array $variables) {
   $suggestions[] = 'form_element__' . str_replace('-', '_', $variables['element']['#type']);
+
+  if (isset($variables['element']['#attributes']['data-twig-suggestion'])) {
```
kalinchernev: A comment explaining what is happening here will be appreciated.

web/themes/joinup/joinup_theme.theme
```diff
@@ -992,3 +1018,18 @@ function joinup_theme_preprocess_table__file_formatter_table(&$variables) {
...
 }
+
+/**
+ * Implements hook_form_FORM_ID_alter().
+ *
+ * Add theme suggestions directly to the comment body field.
+ */
+function joinup_theme_form_views_exposed_form_alter(&$form, FormStateInterface $form_state, $form_id) {
+  if (strpos($form['#id'], 'views-exposed-form-search-page-') !== FALSE) {
+    $form['#info']['filter-search_api_fulltext']['label'] = '<span class="search-bar__icon icon icon--search"></span>';
```
kalinchernev: Could you please try to find a way to have an actual textual (even hidden) label? The current way of adding an icon directly without any hint to screen readers might be a problem having [the guidelines](https://www.w3.org/WAI/tutorials/forms/labels/).

Briefly checking the code, there are some visible hints which can be improved:

search_accessibility

web/themes/joinup/prototype/sass/components/_featured.sass
```diff
@@ -156,6 +156,13 @@
     .stats__column--border
       padding-right: 30px
 
+.featured--navy-blue
+  max-width: 1200px
+  justify-content: center
+  //max-width: 970px
```
kalinchernev: You can remove unnecessary code here.

web/themes/joinup/templates/form/form-element--textfield--search-bar.html.twig
```diff
@@ -0,0 +1,104 @@
+{#
+/**
+ * @file
+ * Theme override for a form element.
+ *
+ * Available variables:
+ * - attributes: HTML attributes for the containing element.
```
kalinchernev: Please fix indentation here.

web/themes/joinup/joinup_theme.theme
piotrsmykaj: @kalinchernev Please add your browser and OS X name and version information to can reproduce issue with positioning because on latest Chrome ver 59 and OS X 10.12 it works correctly.

kalinchernev: [System info](https://gist.github.com/anonymous/7d9aa148483b81d37d89c405adbd62da) and chrome i Version 59.0.3071.115 (Official Build) (64-bit), although [FF 54](https://www.mozilla.org/en-US/firefox/54.0/releasenotes/) has the same issue. I'm not logged in.

web/themes/joinup/joinup_theme.theme
```diff
@@ -992,3 +1022,16 @@ function joinup_theme_preprocess_table__file_formatter_table(&$variables) {
...
 }
+
+/**
+ * Implements hook_form_FORM_ID_alter().
+ *
+ * Add theme suggestions and classes in search form.
+ */
+function joinup_theme_form_views_exposed_form_alter(&$form, FormStateInterface $form_state, $form_id) {
+  if (strpos($form['#id'], 'views-exposed-form-search-page-') !== FALSE) {
+    $form['keys']['#attributes']['data-twig-suggestion'] = 'search_bar';
```
brummbar: Please move this to another array key that starts with #. Mangling with the #attributes means that they will be printed in the HTML, which in this case is not wanted. 
`$form['keys']['#extra_suggestion'] = 'search_bar';`
And then you can check that this "property" is set on the render array on the template suggestion hook.

web/themes/joinup/templates/form/container--edit-actions--search-bar.html.twig
```diff
@@ -0,0 +1,35 @@
...
+    has_parent ? 'form-wrapper',
+    is_collection ? 'mdl-grid',
+    is_collection ? 'mdl-grid--no-spacing',
+    is_collection ? 'vertical-tabs__actions',
+    'form__actions-wrapper',
+  ]
+%}
+
+
+    {{ children }}
```
kalinchernev: This modification needs some attention:
- classes are set but never used
- only children is shown, where-as the correction most probably is possible without stripping the rest

Please try to approach the problem with CSS rather than strip-down template.

PR #631, opened 2017-07-08
3580: Entities with long string IDs are causing errors in key value store

claudiu-cristea:


pfrenssen:
Looking good, thanks!

I had some comments on the upstream issue but they are strictly cosmetic and do not affect the functionality. For our purposes the patch in the current state is fine.

Also good to rename those include files, so that Simpletest doesn't get confused by having stray files inside the test folders :+1:

PR #639, opened 2017-07-11
ISAICP-3436: Style solution about page

tom-waat:
Merge ISAICP-3436 into develop


web/modules/custom/contact_information/contact_information.info.yml
```diff
@@ -15,6 +15,7 @@ dependencies:
 config_devel:
   - core.entity_form_display.rdf_entity.contact_information.default
   - core.entity_form_display.rdf_entity.contact_information.inline
+  - core.entity_view_display.rdf_entity.contact_information.about
```
kalinchernev: 2 notes on current display: 
- In wireframe, grey areas are 1 area (owner and contact)
- when multiple items in lists, markup is not showing a list, it's not the best semantically

I'm not going to stop the pull for these 2, but please check whether they are a big problems for next stages of approval.

solution_about

web/modules/custom/owner/config/install/core.entity_view_display.rdf_entity.owner.about.yml
```diff
@@ -0,0 +1,63 @@
+langcode: en
```
tom-waat: It is fixed. But in all these cases I needed to manually edit code which was created by drush config-devel-export

web/themes/joinup/templates/rdf_entity/rdf-entity--about.html.twig
```diff
@@ -0,0 +1,13 @@
+{#
+/**
+ * @file
+ * Themeing for rdf entities.
+ */
+#}
+
+
+<div{{ attributes }}>
```
tom-waat: I decided to remove this template and remove the label in the preprocess function.

web/modules/custom/contact_information/contact_information.info.yml
tom-waat: > In wireframe, grey areas are 1 area (owner and contact)
I think that it is better if the fields are separated, it would be not consistent if some fields would have margin and some don't have.

> when multiple items in lists, markup is not showing a list, it's not the best semantically
It is default template from the drupal core, and there are divs instead of list.

web/themes/joinup/templates/rdf_entity/rdf-entity--about.html.twig
kalinchernev: Thank you!

PR #641, opened 2017-07-12
ISAICP-3485: mdl input/textarea improvements: sass/tpl

luke-waat:


kalinchernev:
Adding some screenshots.
- Sizing is not convenient:
comments area
- The prototype needs the new label container (that's only for your use, so it's not a blocker)
general_inputs

luke-waat:
It looks like behat issue occurs also on develop branch. Could you please have a look @kalinchernev ?

PR #657, opened 2017-07-17
3608: Purge SPARQL before starting the Drupal installation

claudiu-cristea:


src/Phing/RdfPurge.php
```diff
@@ -0,0 +1,22 @@
...
+
+  /**
+   * Purges the whole RDF backend.
+   */
+  public function main() {
+    // Purge RDF backend.
+    $this->execute('sparql DELETE { GRAPH ?g { ?entity ?field ?value } } WHERE { GRAPH ?g { ?entity ?field ?value . } };');
+    // Purge Solr index.
+    foreach (['drupal_published', 'drupal_unpublished'] as $core) {
+      exec('curl "http://localhost:8983/solr/' . $core . '/update?stream.body=<delete><query>*:*</query></delete>&commit=true"');
```
sandervd: The solr path is not local on the servers

claudiu-cristea: This runs only in the `install-dev` context and I saw no Solr property to make it configurable

PR #669, opened 2017-07-18
Isaicp 3497: Collection notification scheme

idimopoulos:


web/modules/custom/joinup_notification/src/EventSubscriber/CommunityContentSubscriber.php
```diff
@@ -249,13 +242,17 @@ protected function getConfigurationName() {
...
     $actor = $this->entityTypeManager->getStorage('user')->load($this->currentUser->id());
     $actor_first_name = $arguments['@actor:field_user_first_name'];
     $actor_last_name = $arguments['@actor:field_user_family_name'];
     $motivation = isset($this->entity->motivation) ? $this->entity->motivation : '';
 
     $arguments['@transition:motivation'] = $motivation;
+    $parent = $this->relationManager->getParent($entity);
+    if (empty($parent)) {
+      return $arguments;
+    }
```
idimopoulos: This is merely a precaution. Normally there should not be call for an entity without a parent but this is the email handler. I don't think it's nice to throw an exception in the argument generator method.

I moved the actor's first+last name higher so that it is not blocked by this if.

PR #672, opened 2017-07-19
3613: Turn Solr connection into a build property

claudiu-cristea:


src/Phing/RdfPurge.php
```diff
@@ -49,7 +49,7 @@ public function setSolrUrl($solr_url) {
    * @param string[] $solr_cores
    *   A comma delimited list of Solr core names.
    */
-  public function setSolrCores($solr_cores) {
+  public function setSolrCores(array $solr_cores) {
```
claudiu-cristea: Actually this is not correct. The `$solr_core` var is a string (then the docs is wrong s/string[]/string/). That is the value we are passing in the .xml file.

claudiu-cristea:
@pfrenssen, pPleased that you have avoided sticking the beehive, by changing `build.continuousphp.xml` (at least in this phase) :)

For reviewer: I discussed with @pfrenssen, and I'm OK with all the improvements that were made to the original PR.

pfrenssen:
The PHPUnit tests also need to have the Solr configuration set.

pfrenssen:
Ignore the CLA thing, I merged the latest HEAD from upstream (https://github.com/pfrenssen/drupal-project) - I needed to split up the Phing target that uncomments the inclusion of the `settings.local.php` file and contributed this upstream.

PR #679, opened 2017-07-21
ISAICP-3633: Add dropdown for leave button

tom-waat:
Merge ISAICP-3633 into develop.

### DIFF
--- a/web/modules/custom/collection/src/Form/JoinCollectionForm.php
+++ b/web/modules/custom/collection/src/Form/JoinCollectionForm.php
@@ -50,26 +50,48 @@ class JoinCollectionForm extends FormBase {
     // confirmation form, disguised as a form submit button. The confirmation
     // form should open in a modal dialog for JavaScript-enabled browsers.
     if (Og::isMember($collection, $user)) {
-      $form['leave'] = [
-        '#type' => 'link',
-        '#title' => $this->t('Leave this collection'),
-        '#url' => Url::fromRoute('collection.leave_confirm_form', [
-          'rdf_entity' => $collection->id(),
-        ]),
-        '#attributes' => [
-          'class' => [
-            'use-ajax',
-            'button',
-            'button--small',
-            'button--blue-light',
-            'mdl-button',
-            'mdl-js-button',
-            'mdl-button--raised',
-            'mdl-js-ripple-effect',
-            'mdl-button--accent',
+      $form['leave_wrapper'] = [
+        '#type' => 'dropbutton',
+        '#links' => [
+          'member' => [
+            'type' => 'item',
+            'title' => $this->t("You're a member"),
+            'attributes' => [
+              'class' => [
+                'button',
+                'button--small',
+                'button--blue-light',
+                'mdl-button',
+                'mdl-js-button',
+                'mdl-button--raised',
+                'mdl-js-ripple-effect',
+                'mdl-button--accent',
+              ],
+              'disabled' => TRUE,
+            ],
           ],
-          'data-dialog-type' => 'modal',
-          'data-dialog-options' => Json::encode(['width' => 'auto']),
+          'leave' => [
+            'type' => 'link',
+            'title' => $this->t('Leave this collection'),
+            'url' => Url::fromRoute('collection.leave_confirm_form', [
+              'rdf_entity' => $collection->id(),
+            ]),
+            'attributes' => [
+              'class' => [
+                'use-ajax',
+                'button',
+                'button--small',
+                'button--blue-light',
+                'mdl-button',
+                'mdl-js-button',
+                'mdl-button--raised',
+                'mdl-js-ripple-effect',
+                'mdl-button--accent',
+              ],
+              'data-dialog-type' => 'modal',
+              'data-dialog-options' => Json::encode(['width' => 'auto']),
+            ],
+          ],
         ],
       ];
       $form['#attached']['library'][] = 'core/drupal.ajax';
--- a/web/themes/joinup/prototype/sass/_drupal-hacks.sass
+++ b/web/themes/joinup/prototype/sass/_drupal-hacks.sass
@@ -175,3 +175,28 @@ textarea
     &:before
       +icon-size(15px)
       content: '\e617'
+
+// Additional styling for dropbutton.
+.dropbutton-action
+  .button
+    margin: 0
+  &.secondary-action
+    position: absolute
+    white-space: nowrap
+    .button
+      margin-top: 5px
+
+.dropbutton-toggle
+  .dropbutton-arrow
+    color: #fff
+
+.dropbutton-multiple
+  &.open
+    .dropbutton-arrow
+      top: 40% !important
+  .dropbutton-widget
+    padding-right: 0 !important
+
+.form__actions_inner
+  .dropbutton-multiple
+    @extend .listed-button
--- a/web/themes/joinup/prototype/sass/base/_shared.sass
+++ b/web/themes/joinup/prototype/sass/base/_shared.sass
@@ -55,3 +55,7 @@
 
 hr
   margin: 0
+
+.listed-button
+  justify-content: space-between
+  margin: 15px 15px 0 0
--- a/web/themes/joinup/prototype/sass/components/_featured.sass
+++ b/web/themes/joinup/prototype/sass/components/_featured.sass
@@ -140,6 +140,7 @@
   display: inline-block
   margin: 10px
   z-index: 99
+  vertical-align: middle
 
 // Modifiers
 // --------------------------------------------------
--- a/web/themes/joinup/prototype/sass/components/_button.sass
+++ b/web/themes/joinup/prototype/sass/components/_button.sass
@@ -105,5 +105,3 @@
 .adms-validator-form
   .button
     margin-top: $form-margin
-// States
-// --------------------------------------------------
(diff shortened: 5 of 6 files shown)


web/modules/custom/collection/src/Form/JoinCollectionForm.php
```diff
@@ -50,26 +50,48 @@ public function buildForm(array $form, FormStateInterface $form_state, AccountPr
...
-            'button',
-            'button--small',
-            'button--blue-light',
-            'mdl-button',
-            'mdl-js-button',
-            'mdl-button--raised',
-            'mdl-js-ripple-effect',
-            'mdl-button--accent',
+      $form['leave_wrapper'] = [
+        '#type' => 'dropbutton',
```
kalinchernev: Is there a reason not to use the MDL dropdown already available in the site like: https://getmdl.io/components/index.html#menus-section ?
If there's no specific instruction to avoid it, I'd recommend it as it will be consistent with the rest of the dropdowns like this, like in the login, for example.

web/themes/joinup/prototype/sass/_drupal-hacks.sass
```diff
@@ -175,3 +175,28 @@ textarea
     &:before
       +icon-size(15px)
       content: '\e617'
+
+// Additional styling for dropbutton.
+.dropbutton-action
```
kalinchernev: Honestly, the styles here don't look too consistent to the other elements for few reasons:
- this element looks like a button:
btn1
However, it's not clickable on the whole surface and it's just a span - tricky both for accessibility and usability
- the colors are very different than this button which is just next to the button dropdown:
btn2

In comparison of those 2, the first one looks disabled.

In addition, some attributes for accessibility for this type of element are missing:
https://www.w3.org/TR/2016/WD-wai-aria-practices-1.1-20161214/#menubutton
https://w3c.github.io/aria-practices/#menubutton)
Examples:

https://www.w3.org/TR/2016/WD-wai-aria-practices-1.1-20161214/examples/menu-button/menu-button-1/menu-button-1.html
https://w3c.github.io/aria-practices/examples/menu-button/menu-button-links.html

web/modules/custom/collection/src/Form/JoinCollectionForm.php
tom-waat: Hi @kalinchernev ,
there is already dropdown button in drupal, it is something different than mdl dropdown (there is an example on add custom page form). 
On the designs there is no "You're a member" text, only "Leave collection" button (or "Join collection"). 
I thought that the additional text "You're a member" should be also styled like button, so it would be consistent with other buttons which are in this place (right part of the header).
I used dropdown button, but because "You're a member" is not a real button, it is not clickable (only arrow, because it dropdown buttons only arrow triggers dropdown). So this button is disabled and because of that it has different color. If needed I could remove this disable state (so entire button would trigger dropdown) or change the color for disable state.

I can also change this dropdown and use mdl dropdown. But I'm not sure how should look the "You're a member" text. Label, white area, button?

kalinchernev: Hi @tom-waat 
I synced with Wesley for this:
- the whole button should be clickable
- the MDL dropdown should be used:

correct_animation

tom-waat: Hi @kalinchernev ,
I replaced it by mdl dropdown. 
Do we need also arrow (up/down) in the button? If yes, than probably I would need make changes in mdl repository otherwise it would be hard to get state of the dropdown for an arrow (open/closed).

I also noticed that some dropdowns have orange links and some blue. Maybe it also should be unified (ex. all blue)?

web/themes/joinup/templates/join-collection-form.html.twig
```diff
@@ -0,0 +1,27 @@
+{#
+/**
+ * @file
+ * Template to render the share content form.
```
kalinchernev: Several remarks about this template:
- could be named better, as it's not for joining, but seems mainly concerned about leaving
- comment blocks are wrong
- not placed in correct folder

web/themes/joinup/prototype/sass/components/_featured.sass
```diff
@@ -140,6 +140,7 @@
   display: inline-block
   margin: 10px
   z-index: 99
+  vertical-align: middle
```
kalinchernev: Why adding this?

web/themes/joinup/prototype/sass/base/_shared.sass
```diff
@@ -52,3 +52,7 @@
 
 hr
   margin: 0
+
+.listed-button
```
kalinchernev: Honestly, even after reading the code, I still don't understand the naming.

web/themes/joinup/prototype/sass/_drupal-hacks.sass
```diff
@@ -182,3 +182,28 @@ form .description
...
```
kalinchernev: I don't think you need this additional styling any more?

web/themes/joinup/prototype/sass/components/_dropdown-menu.sass
```diff
@@ -0,0 +1,24 @@
...
+// Component
+// --------------------------------------------------
+//.dropdown-menu
+
+// Elements
+// --------------------------------------------------
+.dropdown-menu__item a
+  white-space: nowrap !important
+  color: $color-blue-default
+  text-decoration: none
```
kalinchernev: `.featured__inner-wrapper a` already applies `text-decoration: none`

web/themes/joinup/prototype/sass/components/_dropdown-menu.sass
```diff
@@ -0,0 +1,24 @@
+// Dropdown menu
+//
+// Dropdown component
+
+// Component
+// --------------------------------------------------
+//.dropdown-menu
```
kalinchernev: if this is the part where you make something reusable for ISAICP-3578 - isn't it worth making the html as well? (for the prototype)
Honestly, when I see this specific styling in a component file, I expect to have a bigger picture what it is for, if I am to reuse it. For now, the prototype shows a dropdown in the circle menu which is a component namespace and implementation by itself.

web/themes/joinup/prototype/sass/_drupal-hacks.sass
kalinchernev: a11y attributes are still missing.
The button is inaccessible with tabbing:
jul 28 2017 1-51 pm

web/modules/custom/collection/src/Form/JoinCollectionForm.php
kalinchernev: Hi @tom-waat the visualization looks fine for the moment. Indeed colors are not same, but I think that's ok since the colors of the buttons are also different. I don't think arrow animations are necessary for the moment, let's try to keep it simple.

web/themes/joinup/templates/join-collection-form.html.twig
tom-waat: > could be named better, as it's not for joining, but seems mainly concerned about leaving

This form is created in /web/modules/custom/collection/src/Form/JoinCollectionForm.php and there are two functionalities - join and leave, there is name "JoinCollection" so I used it also for template.

> not placed in correct folder
There is a bug in the drupal core - https://www.drupal.org/node/2817103. The template is not detected if I place it in subfolder.

web/themes/joinup/prototype/sass/components/_featured.sass
tom-waat: When I used previously the dropdown button, there was a positioning issue. It is not needed anymore.

web/themes/joinup/prototype/sass/base/_shared.sass
tom-waat: I renamed it. It is used when several buttons are inline ex. action button under form.

web/themes/joinup/prototype/sass/_drupal-hacks.sass
tom-waat: It fixes dropdown button on form pages ex. add custom page. It fixes an issue described in ISAICP-3578

web/themes/joinup/prototype/sass/components/_dropdown-menu.sass
tom-waat: Ok, I moved it to _mdl-hacks.sass. It is not related to ISAICP-3578

tom-waat:
The button is accessible with tabbing. It has now the same structure as other dropdown buttons. Are there needed other attributes?

web/themes/joinup/templates/join-collection-form.html.twig
kalinchernev: I quickly checked the [docs](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Render%21theme.api.php/function/hook_theme/8.2.x) and some examples in core, with a bit of specificity, I think it will be possible to move the templates in the `form` folder. I quickly checked with this, please double-check if it also works for you? If yes, could you please move them in the form folder and apply this suggestion?
[forms_path.txt](https://github.com/ec-europa/joinup-dev/files/1190841/forms_path.txt)

web/themes/joinup/prototype/sass/_drupal-hacks.sass
kalinchernev: Please keep this set of changes for ISAICP-3578. There are 2 tickets for separate cases, so please don't mix, it's hard to follow.

PR #680, opened 2017-07-21
Isaicp 3508: Community content notification

idimopoulos:


tests/features/community_content/notification.post_moderated.discussion.feature
```diff
@@ -86,7 +86,7 @@ Feature: Notification test for the discussion transitions on a post moderated pa
     Then the following email should have been sent:
       | recipient | CC member                                                                                                                                                        |
       | subject   | Joinup: Content has been updated                                                                                                                                 |
-      | body      | the Facilitator, CC Facilitator has approved your request to publish the discussion - "CC notify post approve proposed" in the collection: "CC post collection". |
+      | body      | the Facilitator, CC Facilitator has approved your request of publication of the discussion - "CC notify post approve proposed" in the collection: "CC post collection". |
```
pfrenssen: Hmm I think this was better before, I guess this is the typical case where the content editors speak 5 languages but none of them correctly. I am starting to think that having awkward grammar is a hard requirement for the commission. If someone ever receives a mail from the EC which is correctly formatted and spelled then they will distrust it and think it is a phishing attempt.

composer.json
```diff
@@ -160,7 +160,8 @@
             "drupal/state_machine": {
                 "Fire generic events when transition are applied. @see https://www.drupal.org/node/2832415": "https://www.drupal.org/files/issues/fire_generic_events-2832415-2.patch",
                 "Schema validation error when running functional tests. @see https://www.drupal.org/node/2838952": "https://www.drupal.org/files/issues/schema_validation_error-2838952-2.patch",
-                "Disable static cache for workflow_callback. @see https://www.drupal.org/node/2846986": "https://www.drupal.org/files/issues/static_cache_disable_for_workflow_callback-2846986-2.patch"
+                "Disable static cache for workflow_callback. @see https://www.drupal.org/node/2846986": "https://www.drupal.org/files/issues/static_cache_disable_for_workflow_callback-2846986-2.patch",
+                "Allow access to the initialValue protected property. @see https://www.drupal.org/node/2896479": "https://www.drupal.org/files/issues/allow_access_to_initial_value-2896479-2-D8.patch"
```
pfrenssen: Seems like this patch was developed in an earlier iteration but was not used in the final version?

web/modules/custom/joinup_notification/src/EventSubscriber/CommunityContentSubscriber.php
```diff
@@ -249,32 +257,55 @@ protected function generateArguments(EntityInterface $entity) {
...
-
-          if (in_array('administrator', $role_names)) {
-            $arguments['@actor:role'] = t('Owner');
-          }
-          elseif (in_array('facilitator', $role_names)) {
-            $arguments['@actor:role'] = t('Facilitator');
-          }
+    if (empty($parent)) {
+      return $arguments;
+    }
```
pfrenssen: Don't put an exit out of the code flow in the middle of a method, that makes it harder to maintain. This is only good at the very start to provide an early exit if some condition is not met.

web/modules/custom/joinup_notification/src/EventSubscriber/CommunityContentSubscriber.php
```diff
@@ -249,32 +257,55 @@ protected function generateArguments(EntityInterface $entity) {
 
     $arguments['@actor:full_name'] = $actor_first_name . ' ' . $actor_last_name;
     $arguments['@transition:motivation'] = $motivation;
+    $arguments['@entity:hasPublished:status'] = $this->hasPublished ? 'an update of the' : 'a new';
```
pfrenssen: This is OK since Joinup will never be translated to a different language than English, but in a standard multilingual site this wouldn't be good.

composer.json
idimopoulos: It is being used in joinup_notification.module not properly and safely extract the initial state without reloading the unchanged entity :)

tests/features/community_content/notification.post_moderated.discussion.feature
idimopoulos: The second version is so that it becomes a bit more obscure and be used for publication of both new and updated entities.

web/modules/custom/joinup_notification/src/EventSubscriber/CommunityContentSubscriber.php
idimopoulos: Yeah well...

web/modules/custom/joinup_notification/src/EventSubscriber/CommunityContentSubscriber.php
idimopoulos: What follows after this are arguments that should be skipped if anything happens and there is no parent to the entity. This is a safe escape. The lines before are necessary arguments. How should these be handled? split them up in a separate method?

composer.json
pfrenssen: Ahh there it is!

tests/features/community_content/notification.post_moderated.discussion.feature
pfrenssen: I think you misunderstand me. This is really poor language and formatting.

> the

This sentence doesn't start with a capital letter.

> Facilitator,

In English nouns are not capitalized like they are for example in German. The comma here is misplaced.

> CC Facilitator

Here a comma is missing.

> has approved your request of publication of the discussion

This just sounds completely wrong.

> - "CC [...]"

What is this hyphen doing here?

> in the collection: "CC [...]"

What is this double colon doing here?

web/modules/custom/joinup_notification/src/EventSubscriber/CommunityContentSubscriber.php
pfrenssen: No just wrap any code that needs to be skipped in an if statement. Putting a `return` in the middle of a method is poor practice since it can be easily missed when this code is changed in the future. In well written code the return statement is at the end.

tests/features/community_content/notification.post_moderated.discussion.feature
idimopoulos: For the first part, the
> the
Comes after the "Dear <firstname>," which is not mentioned here, and this ends in a comma. The next paragraph does not start with a capital letter as it is considered an extend of this phrase.
> Facilitator,
This is a role name. That depends on the definition of the role but you can argue with Andrea on that.
> CC Facilitator
Indeed a comma could be added there. The previous comma was not misplaced as normally you would have something like "The facilitator, Ilias Dimopoulos, has done something". So the first comma is also applicable.
> has approved your request of publication of the discussion
Surely the 'has approved your request to publish the discussion' sounds better but I do not know if this is wrong grammar (maybe I am wrong).

For the last two parts, hyphens and double colons are used in sentences like this. Again, it is not a best practice but I did not QA the texts. Please open follow ups for this or discuss directly with Andrea. The notification system is the scope here.

pfrenssen: No I think he's doing a great job in writing bad English.

PR #687, opened 2017-07-24
Isaicp 3510: Report notifications

idimopoulos:


web/modules/custom/contact_form/src/EventSubscriber/NotificationSubscriber.php
```diff
@@ -76,19 +63,160 @@ public function contactForm(NotificationEvent $event) {
...
+    }
+
+    $route_parameters = $url->getRouteParameters();
+    if (!isset($route_parameters['node']) && !isset($route_parameters['rdf_entity'])) {
+      return NULL;
+    }
+
+    // Check what type of content we are reporting.
+    $entity_type_id = NULL;
+    if (isset($route_parameters['node'])) {
```
claudiu-cristea: What happens with URIs like this: `/node/123/revision/3`. This will return the entity but not the specific entity

web/modules/custom/contact_form/src/EventSubscriber/NotificationSubscriber.php
```diff
@@ -76,19 +63,160 @@ public function contactForm(NotificationEvent $event) {
...
+   *
+   * Skip generating the arguments during the sending process.
+   */
+  protected function sendUserDataMessages(array $user_data, array $arguments = []) {
+    foreach ($user_data as $template_id => $user_ids) {
+      $values = ['template' => $template_id, 'arguments' => $arguments];
+      $message = Message::create($values);
+      $message->save();
+
+      foreach ($user_ids as $user_id) {
```
claudiu-cristea: Never load entity by entity for performance reasons. We should use here: `foreach ($this->entityTypeManager->getStorage('user')->loadMultiple($user_ids) as $user_id => $user)`.

web/modules/custom/contact_form/src/EventSubscriber/NotificationSubscriber.php
```diff
@@ -76,19 +63,160 @@ public function contactForm(NotificationEvent $event) {
...
+        /** @var \Drupal\user\Entity\User $user */
+        $user = $this->entityTypeManager->getStorage('user')->load($user_id);
+        if ($user->isAnonymous()) {
```
claudiu-cristea: This we can avoid if previously we would: `$user_ids = array_filter($user_ids)`.

web/modules/custom/contact_form/src/EventSubscriber/NotificationSubscriber.php
idimopoulos: Well, this entity will be used to get memberships and a title. It does not matter which revision is reported as if it's an earlier version, it's not meant to be seen by everyone. 
Also, in the same method, there is a check about the canonical path. so in case of the node revision link, it will not be valid anyway.

web/modules/custom/contact_form/src/EventSubscriber/NotificationSubscriber.php
idimopoulos: ahh yeah.. you are right there. it's better to load all of them directly.

PR #688, opened 2017-07-25
ISAICP-3498: User notifications

idimopoulos:


web/modules/custom/joinup_notification/joinup_notification.behat.inc
```diff
@@ -149,6 +149,114 @@ class JoinupNotificationSubContext extends DrupalSubContextBase implements Drupa
...
+   *   Throws an exception when a parameter is not the expected one.
+   *
+   * @Then the following system email should have been sent:
+   */
+  public function assertSystemEmailSent(TableNode $table) {
+    /** @var string $body */
+    /** @var string $recipient */
+    /** @var string $recipient_mail */
+    /** @var string $subject */
+    extract($table->getRowsHash());
```
pfrenssen: There is also the `$html` variable being extracted. BTW this would be better with normal use of the rows so we can use human friendly column names without underscores.

web/modules/custom/joinup_notification/joinup_notification.behat.inc
```diff
@@ -149,6 +149,114 @@ class JoinupNotificationSubContext extends DrupalSubContextBase implements Drupa
...
+
+    // If the 'html' row is not present, assume a 'yes'. Convert 'yes' to TRUE
+    // and 'no' to FALSE.
+    $html = (isset($html) ? $html : 'yes') === 'yes';
```
pfrenssen: you can save one 'yes' :)

`$html = empty($html) || $html === 'yes';`

Never mind, both work fine :)

web/modules/custom/joinup_user/config/schema/joinup_user.schema.yml
```diff
@@ -6,3 +6,9 @@ joinup_user.mail:
   email_admin_update:
     type: mail
     label: 'Account updated by administrator confirmation'
+  og_roles_changed:
+    type: mail
+    label: 'Og roles changes'
```
pfrenssen: Typo here, should be 'OG roles changed' I think.

web/modules/custom/joinup_user/joinup_user.behat.inc
```diff
@@ -29,4 +29,72 @@ class JoinupUserSubContext extends DrupalSubContextBase implements DrupalSubCont
...
+    $user = $this->getUserByName($user);
+    if (empty($user)) {
+      throw new \Exception("User {$user->getAccountName()} was not found.");
+    }
+
+    $this->visitPath(user_pass_reset_url($user) . '/login');
+  }
+
+  /**
+   * Navigates to the one time log in page of the user.
```
pfrenssen: This description is copied from the previous method, needs to be updated.

web/modules/custom/joinup_user/joinup_user.module
```diff
@@ -329,8 +441,42 @@ function joinup_user_entity_base_field_info(EntityTypeInterface $entity_type) {
  * Implements hook_ENTITY_TYPE_presave() for OG Membership entities.
  */
 function joinup_user_og_membership_presave(EntityInterface $entity) {
-  // When a membership changes, clear the user's cache so that the related
-  // content on the user profile page matches the new group memberships.
+  joinup_user_invalidate_user_cache_tags($entity);
```
pfrenssen: Why is this documentation removed? It explains why it is needed to invalidate the cache tags. Without this documentation this line becomes really suspicious.

web/modules/custom/joinup_user/joinup_user.module
```diff
@@ -329,8 +441,42 @@ function joinup_user_entity_base_field_info(EntityTypeInterface $entity_type) {
...
+
+  if ($entity->getEntityTypeId() === 'og_membership') {
+    /** @var \Drupal\user\Entity\User $account */
+    $account = $entity->getUser();
```
pfrenssen: This is not documented.

web/modules/custom/joinup_user/joinup_user.module
```diff
@@ -329,8 +441,42 @@ function joinup_user_entity_base_field_info(EntityTypeInterface $entity_type) {
...
+      $params
+    );
+  }
+}
+
+/**
+ * Implements hook_ENTITY_TYPE_predelete() for OG Membership entities.
+ */
+function joinup_user_og_membership_predelete(EntityInterface $entity) {
+  joinup_user_invalidate_user_cache_tags($entity);
```
pfrenssen: Why is it needed to invalidate cache tags when a membership is deleted? Is there a test for this?

web/modules/custom/joinup_user/joinup_user.module
```diff
@@ -329,8 +441,42 @@ function joinup_user_entity_base_field_info(EntityTypeInterface $entity_type) {
...
+}
+
+/**
+ * Flush the caches of the user that belongs to a OG membership.
+ *
+ * @param \Drupal\Core\Entity\EntityInterface $entity
+ *   The OG membership entity.
+ */
+function joinup_user_invalidate_user_cache_tags(EntityInterface $entity) {
   if ($entity->getEntityTypeId() === 'og_membership') {
```
pfrenssen: If you type hint on `OgMembershipInterface` instead of on the generic `EntityInterface` then you don't need this switch.

web/modules/custom/joinup_user/joinup_user.module
```diff
@@ -342,8 +488,24 @@ function joinup_user_og_membership_presave(EntityInterface $entity) {
...
+function joinup_user_form_user_multiple_cancel_confirm_alter(&$form, FormStateInterface $form_state, $form_id) {
+  joinup_user_form_user_cancel_form_alter($form, $form_state, $form_id);
+}
+
+/**
+ * Implements hook_form_FORM_ID_alter().
+ *
+ * Disallow access to the cancellation method.
+ */
+function joinup_user_form_user_cancel_form_alter(&$form, FormStateInterface $form_state, $form_id) {
```
pfrenssen: Please explain why this is needed. Also, this changes it for all users including UID1. Is it not only needed to restrict access for moderators?

Also as I said yesterday, why do moderators have access to this form in the first place? They have nothing to search here. If it is only for the email messages it would be better to move this to a dashboard entry.

And secondly, this whole exercise is useless since we will make config read-only on production so moderators cannot change ANY of the values in the backend anyway. If they want a change to one of the messages then they will need to create a ticket and this will be updated on the next release.

web/profiles/joinup/joinup.behat.inc
```diff
@@ -1701,6 +1701,15 @@ class JoinupSubContext extends DrupalSubContextBase implements DrupalSubContextI
   }
 
   /**
+   * Wait until the id="updateprogress" element is gone, or until timeout.
```
pfrenssen: Hehe this documentation is only useful for people that know what this ID stands for :)

web/profiles/joinup/src/Form/UserCancelForm.php
```diff
@@ -0,0 +1,113 @@
...
+   * @param \Drupal\Component\Datetime\TimeInterface $time
+   *   The time service.
+   * @param \Drupal\joinup_core\JoinupRelationManager $relation_manager
+   *   The Joinup relation manager.
+   */
+  public function __construct(EntityManagerInterface $entity_manager, EntityTypeBundleInfoInterface $entity_type_bundle_info = NULL, TimeInterface $time = NULL, JoinupRelationManager $relation_manager = NULL) {
+    parent::__construct($entity_manager, $entity_type_bundle_info, $time);
+
+    // Replicate the behaviour of the parent implementation.
+    $this->relationManager = $relation_manager ?: \Drupal::service('joinup_core.relations_manager');
```
pfrenssen: Huh? If the relation manager is required then just move it to the front of the arguments and make it non-optional. A constructor is never part of an interface so you don't need to respect a certain parameter order.

web/profiles/joinup/src/Form/UserCancelForm.php
```diff
@@ -0,0 +1,113 @@
...
+          // @see \Drupal\Core\Entity\ContentEntityConfirmFormBase::actions()
+          'cancel' => ConfirmFormHelper::buildCancelLink($this, $this->getRequest()),
+          '#weight' => 10,
+        ],
+      ];
+
+      // Show a nicer message when the user is trying to delete its own account.
+      if ($this->currentUser()->id() === $this->entity->id()) {
+        $form['pre_warning'] = [
+          '#markup' => $this->t('Dear @name,<br />when processing your request to delete your account, we noticed that you are the sole owner of these collections:', [
```
pfrenssen: Lol this makes it sound like there is a person working in the background checking files and writing a letter with their findings. It's not very typical to address people by name in a confirmation form.

tests/features/user/cancel.feature
```diff
@@ -0,0 +1,71 @@
...
+    And I click "People"
+    And I click "Alicia Potter"
+    And I click "Edit" in the "Entity actions" region
+    And I press "Cancel account"
+    And I press "Cancel account"
+    And I wait for the batch job to finish
+    And the following system email should have been sent:
+      | recipient_mail | AliciaPotter@example.com                                                                                                                                                                                                                                               |
+      | subject        | Your account has been deleted.                                                                                                                                                                                                                                         |
+      | body           | Your account Alicia Potter has been deleted.This action has been done in the framework of moderation activities regularly conducted on the Joinup platform. If you believe that this action has been performed by mistake, please contact the Joinup Community team at |
```
pfrenssen: I am mistaken, we do allow people to choose a unique account name, it is not necessarily the same as the person's real name. Let's just update the test to use a realistic account name then.

web/modules/custom/joinup_core/src/JoinupRelationManager.php
```diff
@@ -123,4 +134,122 @@ public function getParentElibrary(EntityInterface $entity) {
...
+
+    $users = [];
+    foreach ($this->getGroupMemberships($entity, $state) as $membership) {
+      $user = $membership->getUser();
+      if (!empty($user) && $membership->hasRole($role_id)) {
+        $users[$user->id()] = $user;
+      }
+    }
+
+    return array_filter($users);
```
idimopoulos: The array_map above can still put empty values in the keys. The array map above can create something like
```
[ 0 => null, 1 => 2, 2 => 79000, 3 => NULL ];
```
The array filter makes sure that we do not mistake the 0 for the anonymous user.

idimopoulos: Ahh sorry! I was looking to the next method. Indeed this array_filter is not needed. I noticed that as well at some point but the ticket was already in QA.
I will remove it to make sure that no tests are failing.

web/modules/custom/joinup_core/src/JoinupRelationManager.php
```diff
@@ -123,4 +134,122 @@ public function getParentElibrary(EntityInterface $entity) {
...
+   * @return array
+   *   An array of users that are members of the entity group.
+   */
+  public function getGroupUsers(EntityInterface $entity, array $state = [OgMembershipInterface::STATE_ACTIVE]) {
+    $users = array_map(function (OgMembershipInterface $membership) {
+      $user = $membership->getUser();
+      return empty($user) ? NULL : $user->id();
+    }, $this->getGroupMemberships($entity, $state));
+
+    return array_filter($users);
```
idimopoulos: I am not sure how to do that in the array filter. the `return empty($user) ? NULL : $user->id();` does not filter out entries, it just makes sure that the $user->id() is not going to throw an exception if $user is null.

web/modules/custom/joinup_core/src/JoinupRelationManager.php
```diff
@@ -123,4 +134,122 @@ public function getParentElibrary(EntityInterface $entity) {
...
+  public function getGroupMemberships(EntityInterface $entity, array $state = [OgMembershipInterface::STATE_ACTIVE]) {
+    /** @var \Drupal\og\OgMembershipInterface[] $memberships */
+    $memberships = $this->entityTypeManager->getStorage('og_membership')->loadByProperties([
+      'state' => $state,
+      'entity_type' => $entity->getEntityTypeId(),
+      'entity_id' => $entity->id(),
+    ]);
+
+    return $memberships;
+  }
```
idimopoulos: Nope :) it does not! and we stumbled across a few times in the past. Quite weird, no? 
It's good that Francesco put it in a helper class.

web/modules/custom/joinup_core/src/JoinupRelationManager.php
```diff
@@ -123,4 +134,122 @@ public function getParentElibrary(EntityInterface $entity) {
...
+   *   The user to get the memberships for.
+   * @param string $role
+   *   The role id.
+   * @param array $state
+   *   (optional) An array of membership states to retrieve. Defaults to active.
+   *
+   * @return \Drupal\og\OgMembershipInterface[]
+   *   An array of OG memberships that match the criteria.
+   */
+  public function getUserMembershipsByRole(AccountInterface $user, $role, array $state = [OgMembershipInterface::STATE_ACTIVE]) {
```
idimopoulos: Opened it https://github.com//og/issues/294

web/modules/custom/joinup_user/joinup_user.module
idimopoulos: True but then a mistake would throw an exception instead of silently failing.

web/modules/custom/joinup_user/joinup_user.module
idimopoulos: The user_cancel_delete_form is not the admin form. It's the delete confirmation page.
Why is this needed:
  Unlike other entities, the user entity does not have a 'edit own/any', 'delete/own any', 'select cancel method' etc. There is one permission to allow delete any user and this permission is 'administer users'. The same permission gives access to the cancellation method and the ability to choose whether to notify the user etc.

idimopoulos: Also, as for the read-only config, this form is not storing anything in the database and the config storage settings for the user deletion serve only as a default value when the user has the 'administer users' permission.

web/profiles/joinup/joinup.behat.inc
idimopoulos: You can comment that on the blog I copied it from :P I will change it. :)

web/profiles/joinup/src/Form/UserCancelForm.php
idimopoulos: If you check the parent constructor, the other two services are also non mandatory. They are handled the same way as the relation manager is handled here.

web/profiles/joinup/src/Form/UserCancelForm.php
idimopoulos: That was implemented that way because we converted a notification message into a confirmation message (because the other way around would be overkill).

web/modules/custom/joinup_core/src/JoinupRelationManager.php
pfrenssen: Sorry I mean `array_reduce()` instead of `array_filter()`:

```
array_reduce($memberships, function ($users, $membership) {
  if (!empty($user = $membership->getUser())) $users[] = $user;
  return $users;
}, []);
```

web/modules/custom/joinup_user/joinup_user.module
pfrenssen: Yeah I know this, but the person reading the code doesn't know :) Thanks for adding the description :+1:

web/modules/custom/joinup_user/joinup_user.module
pfrenssen: Yes that's the whole point :grin:

web/modules/custom/joinup_user/joinup_user.module
pfrenssen: Ohh I get it now, thanks for explaining. I confused this with the account settings admin form which also shows this option.

web/profiles/joinup/src/Form/UserCancelForm.php
pfrenssen: Yeah but that's part of core and they do it because they cannot break backwards compatibility with earlier versions of Drupal 8 that had fewer arguments to this constructor.

Our code is new so we don't break backwards compatibility and don't need this workaround.

web/profiles/joinup/src/Form/UserCancelForm.php
pfrenssen: OK leave it in, I can imagine some of our users finding this quite amusing :smile:

tests/features/user/cancel.feature
pfrenssen: This was apparently forgotten, I'll address it.

web/modules/custom/joinup_notification/joinup_notification.behat.inc
pfrenssen: This is very minor and only affects test coverage. It's fine like this for now.

web/modules/custom/joinup_user/joinup_user.module
pfrenssen: I think this switch is even completely unneeded since we're implementing the entity type hook for `OgMembership` entities. So instead of documenting this switch we can remove it and update the type hint in the function argument.

PR #689, opened 2017-07-25
ISAICP-3646: Forms unification

tom-waat:
Merge ISAICP-3646 into develop.


web/themes/joinup/joinup_theme.theme
```diff
@@ -39,24 +42,35 @@ function joinup_theme_preprocess_page(&$variables) {
...
   $blue_header_routes = [
+    'adms_validator.adms_validator_form',
     'collection.propose_form',
+    'contact_form.contact_page',
     'entity.node.edit_form',
     'entity.rdf_entity.edit_form',
-    'node.add',
-    'node.add_page',
     'rdf_entity.propose_form',
-    'rdf_entity.rdf_add',
```
kalinchernev: I was testing some more generic routes like this.
For example, the header is not there on:
- `/rdf_entity/add/asset_distribution`
- `rdf_entity/add/asset_release`

and others coming from `rdf_entity/add`. I know content is not meant to be added from these forms, yet the ticket is saying "all forms", so I don't see why not have the header on these as well, especially given that, for example, "Add RDF" button on a collection will lead here.

web/themes/joinup/joinup_theme.theme
```diff
@@ -89,11 +103,27 @@ function joinup_theme_preprocess_page(&$variables) {
 
     if (in_array($rdf_type, ['collection', 'solution'])) {
       $variables['has_vertical_tabs'] = TRUE;
+      $is_form_container = TRUE;
+    }
+  }
+
+  // Check pattern for add pages.
+  $form_pattern = '.add';
```
kalinchernev: Looks a bit unclear and unstable pattern. Can you maybe think of a regex or something that depends on specific elements rather than length of a string?

web/themes/joinup/joinup_theme.theme
```diff
@@ -128,21 +164,26 @@ function joinup_theme_preprocess_page_title(&$variables) {
...
+
   $route_name = $route_match->getRouteName();
 
   // Array contains create pages.
   $create_content_routes = [
     'collection.propose_form',
-    'node.add',
-    'node.add_page',
     'rdf_entity.propose_form',
-    'rdf_entity.rdf_add',
```
kalinchernev: Same comment as above.

web/themes/joinup/joinup_theme.theme
```diff
@@ -128,21 +164,26 @@ function joinup_theme_preprocess_page_title(&$variables) {
...
-    'solution.collection_solution.add',
   ];
 
   // Set a flag when we are on content creation routes
   // in order to customise some templates.
+  $form_pattern = '.add';
+  $route_length = strlen($route_name);
+  $pattern_length = strlen($form_pattern);
+  if ($route_length > $pattern_length) {
+    if (substr_compare($route_name, $form_pattern, $route_length - $pattern_length, $pattern_length) === 0) {
```
kalinchernev: Same comment as above + repetition. A function might be helpful.

web/themes/joinup/joinup_theme.theme
```diff
@@ -116,6 +146,12 @@ function joinup_theme_preprocess_page(&$variables) {
     }
   }
 
+  if ($is_form_container) {
+    $variables['container_class'] = 'content-container--form';
```
kalinchernev: This class is removed from the layout styles - why is it used here?

web/themes/joinup/templates/custom-form.html.twig
```diff
@@ -0,0 +1,16 @@
+{#
+/**
+ * @file
```
kalinchernev: Few questions:
- What is the purpose of this custom form?
- As it seems a bare template, can a template be avoided?
- Can you think of a better naming?

'Custom' is really general and I honestly wouldn't know what to do with it if I don't have context or documentation.

web/themes/joinup/templates/layout/page.html.twig
```diff
@@ -77,7 +77,7 @@
         </aside>
       {% endif %}
 
-      <div class="section section--main mdl-cell {{ has_vertical_tabs ? 'mdl-cell--12-col mdl-grid--no-spacing content-container--form' : 'mdl-cell--8-col content-container' }}">
+      <div class="section section--main {{ container_class }} mdl-cell {{ has_vertical_tabs ? 'mdl-cell--12-col mdl-grid--no-spacing' : 'mdl-cell--8-col' }}">
```
kalinchernev: Not a real problem, but could you please order static and dynamic classes? I mean, container_class is in the middle of static classes and it's a bit unreadable.

web/themes/joinup/joinup_theme.theme
tom-waat: When I created this ticket I was thinking about all forms which are in use. Path like rdf_entity/add/asset_release are not in the normal site workflow, it is not even possible to use them (there are errors). I don't know why this "Add RDF" button exist, some of the forms there cannot be submitted.

web/themes/joinup/joinup_theme.theme
tom-waat: I checked this pattern and I think that it works ok. It doesn't depend only on length of a string. It checks if string is not shorter than '.add', and then function substr_compare checks if '.add' is at the end of a string.
But if you think that it it not good solution, I could replace it by regex.

web/themes/joinup/templates/custom-form.html.twig
tom-waat: > What is the purpose of this custom form?

It allows to separate content and action buttons

> As it seems a bare template, can a template be avoided?

It is in fact only a template for content which is between <form> tags, so other containers are not needed there, because <form> tags are still there.

> Can you think of a better naming?

Hmm, this is a name for all forms which have action buttons separated form the content. Maybe 'buttons-separated-form'?

web/themes/joinup/joinup_theme.theme
kalinchernev: I see your point, as I mentioned I have an idea some forms are out of the workflow. Still, the button is there and the forms are reachable, so please keep them in the list for now.

web/themes/joinup/templates/custom-form.html.twig
kalinchernev: > It is in fact only a template for content which is between form tags, so other containers are not needed there, because form tags are still there.
Can you clarify this please?

web/themes/joinup/joinup_theme.theme
tom-waat: There is content-container--default class, so I think that it would be consistent to have this class in markup even if it is not used in css.

web/themes/joinup/templates/custom-form.html.twig
tom-waat: The template custom-form.html.twig renders content which is inside template core/themes/stable/templates/form/form.html.twig
`<form>`
content from custom-form.html.twig
`</form>`

web/themes/joinup/joinup_theme.theme
kalinchernev: I debugged a bit, and even then I couldn't really understand the checks.
So, yes, please use a better strategy for checking routes.

Also, as mentioned before, these 2 checks are very similar and probably you'll manage to combine them in cascade:

```php
  // Check pattern for add pages.
  $form_pattern = '.add';
  $route_length = strlen($route_name);
  $pattern_length = strlen($form_pattern);
  if ($route_length > $pattern_length) {
    if (substr_compare($route_name, $form_pattern, $route_length - $pattern_length, $pattern_length) === 0) {
      $variables['is_blue_header'] = TRUE;
      $is_form_container = TRUE;
    }
  }

  if (!empty($route_name) && in_array($route_name, $blue_header_routes)) {
    $variables['is_blue_header'] = TRUE;
    $is_form_container = TRUE;
  }
```

Also, the iffing is repeating in other part of the code, you can abstract to a function taking the route as a parameter and telling you whether or not this page is special.

tom-waat: Ok, I moved two checks to a function:
* if page is a create page
* if page has vertical tabs
There is a possibility to extend the function and add new checks in the future.

web/themes/joinup/joinup_theme.theme
```diff
@@ -19,13 +19,77 @@ use Drupal\file\Entity\File;
...
+  }
+
+  if (!empty($route_name) && in_array($route_name, $create_content_routes)) {
+    $types['create'] = TRUE;
+  }
+
+  // Check if page has form vertical tabs.
+  $tabs_routes = [
+    'entity.rdf_entity.edit_form',
+    'rdf_entity.propose_form',
```
kalinchernev: `rdf_entity.propose_form` repeats in both sets. As we spoke today, if there are 2 types of forms, maybe there should be 2 separate functions, or an exclusion logic rather than inclusion. At any case, it's a repetition ..

web/themes/joinup/joinup_theme.theme
```diff
@@ -1014,6 +1061,39 @@ function joinup_theme_theme_suggestions_alter(array &$suggestions, array $variab
...
+    'rdf_entity_collection_propose_form',
+    'rdf_entity_solution_form',
+    'rdf_entity_solution_edit_form',
+    'rdf_entity_collection_add_form',
+    'rdf_entity_solution_add_form',
+  ];
+
+  if (is_array($form['#theme'])) {
+    if (array_intersect($form['#theme'], $custom_forms) && !array_intersect($form['#theme'], $tab_forms)) {
+      $form['#theme'][] = 'buttons_separated_form';
```
kalinchernev: This if - else leads to the same outcome, so it should be an extended if.

web/themes/joinup/joinup_theme.theme
```diff
@@ -19,13 +19,77 @@ use Drupal\file\Entity\File;
...
+  ];
+
+  // Check if page has a create form.
+  $create_content_routes = [
+    'collection.propose_form',
+    'rdf_entity.propose_form',
+  ];
+
+  // Check if string has 'add' on the end.
+  if (preg_match('/.add$/', $route_name)) {
```
tom-waat: I checked all creation routes:
outside workflow:
* create content form has route "node.add"
* create rdf form has route "rdf_entity.rdf_add"
In workflow, content and rdf add on solution and collection level has similar routes:
ex. joinup_news.rdf_entity_news.add, custom_page.collection_custom_page.add, asset_release.solution_asset_release.add

So all of these routes match the pattern. Only propose collection and propose solution forms don't match this pattern, but they are listed in array $create_content_routes

rdf_entity.propose_form

web/themes/joinup/joinup_theme.theme
```diff
@@ -19,13 +19,77 @@ use Drupal\file\Entity\File;
...
+
+  if (!empty($route_name) && in_array($route_name, $tabs_routes)) {
+    // Get entity type on edit form.
+    $entity = $route_match->getParameter('rdf_entity');
+    if (!empty($entity)) {
+      $rdf_type = $entity->getType();
+    }
+    else {
+      // Get entity type on create form.
+      $rdf_type = $route_match->getParameter('rdf_type');
```
tom-waat: There are two ifs, because on create pages like /rdf_entity/add/collection entity variable is empty, so we need to check the parameter.

kalinchernev:
Hi @tom-waat I made a suggestion here https://github.com/ec-europa/joinup-dev/pull/715
It contains:
- refactored functions which gives simple list of routes by topic
- the function for create vs tabs function is touched only a bit to simplify the if statements
- merged develop and resolved conflicts

Please check mainly the last few commits:
- https://github.com/ec-europa/joinup-dev/pull/715/commits/461786f72e015159f71187a5703a01e788eb96e1
- https://github.com/ec-europa/joinup-dev/pull/715/commits/ac301d864486b1b39645ca42cb8d8e3179bb199c
- https://github.com/ec-europa/joinup-dev/pull/715/commits/fe7ec7cc6efb5aaf795dad33573a73ecedac88b4

The rest are develop history that i don't want to rebase. That was the simplest way to show the ideas.

web/themes/joinup/joinup_theme.theme
```diff
@@ -1187,15 +1187,8 @@ function joinup_theme_form_alter(&$form, FormStateInterface $form_state, $form_i
...
-  if (is_array($form['#theme'])) {
-    if (array_intersect($form['#theme'], $custom_forms) && !array_intersect($form['#theme'], $tab_forms)) {
-      $form['#theme'][] = 'buttons_separated_form';
-    }
-  }
-  else {
-    if (in_array($form['#theme'], $custom_forms) && !in_array($form['#theme'], $tab_forms)) {
-      $form['#theme'][] = 'buttons_separated_form';
-    }
+  if (is_array($form['#theme']) && array_intersect($form['#theme'], $custom_forms) && !array_intersect($form['#theme'], $tab_forms)) {
```
kalinchernev: Excellent, thank you!

web/themes/joinup/joinup_theme.theme
```diff
@@ -242,6 +256,12 @@ function joinup_theme_preprocess_page(&$variables) {
     $is_form_container = TRUE;
   }
 
+  // Check if it is add content or rdf overview page.
+  // It needs to be called after other checks.
+  if (!empty($route_name) && in_array($route_name, $not_forms)) {
```
kalinchernev: You can move in 

```
  if (!empty($route_name)) {
    if (in_array($route_name, $other_forms)) {
      $is_form_container = TRUE;
    }

    if (in_array($route_name, $blue_header_routes)) {
      $variables['is_blue_header'] = TRUE;
      $is_form_container = TRUE;
    }

    if (in_array($route_name, $navy_header_routes)) {
      $variables['is_navy_header'] = TRUE;
    }
  }
```

Around line 232 and reuse the `!empty($route_name)`

PR #690, opened 2017-07-25
ISAICP-3630: content issues fix: sass

luke-waat:
Sorry for confusion @kalinchernev, it looks like I have created this branch from grid branch (ISAICP-3560), instead of develop. I would be grateful if you could check ISAICP-3560 first and then review this one once ISAICP-3560 is merged.


kalinchernev:
@tom-waat thanks for taking latest develop!
It seems this branch is off the footer, which adds a dependency.
I'd suggest you cherry pick https://github.com/ec-europa/joinup-dev/pull/690/commits/9460665274c90d4d415ae3bdacfe3978679f1ae7 and open a new pull, that way it becomes much easier for this small change to go forward without any dependencies.

web/themes/joinup/prototype/sass/base/_ordered_list.sass
```diff
@@ -6,7 +6,7 @@
 
 article
   ol
-    padding-left: 0
+    padding-left: 15px
```
kalinchernev: Unfortunately, this is not a sufficient fix to the reported issue:

bullets-still-wrong

luke-waat: The issue is related with bullet points which are vertically centered instead of being inline with first text line, no matter how long the text is. Do I understand it correctly?

kalinchernev: The bullet should be normally aligned and not centered.
Example:
download

luke-waat: This should be done.

kalinchernev: yep, thanks!

PR #676, opened 2017-07-20
ISAICP-3560

tom-waat:
Merge ISAICP-3560 into develop

PR #691, opened 2017-07-25
ISAICP-3593: Implement footer

tom-waat:
Please check once #676 is merged.


web/themes/joinup/prototype/sass/components/_ec-menu.sass
```diff
@@ -0,0 +1,33 @@
+// European Commision menu
```
kalinchernev: Not obligatory, but maybe you could add the footer in the list of components in the style guide similarly to the header.

luke-waat: Added footer under number 2.14 in the style guide. I divided it into two components: EC menu and social media menu. I think that showing it as one component in this default, narrow style guide view might be confusing.

kalinchernev: It's good, thanks!

PR #692, opened 2017-07-27
Isaicp 3519: Notifications for the request/approve/reject operation of a membership

idimopoulos:


tests/features/collection/collection.member_administration.feature
```diff
@@ -52,6 +69,30 @@ Feature: Collection membership administration
...
+    Then I should not see the link "Group"
+    And I check the box "Update the member Kathie Cumbershot"
+    Then I select "Delete the selected membership(s)" from "Action"
+    And I press the "Apply to selected items" button
+    Then I should see the following success messages:
+      | Delete the selected membership(s) was applied to 1 item. |
+    And the following email should have been sent:
+      | recipient | Kathie Cumbershot                                                               |
+      | subject   | Joinup: Your request to join the collection Medical diagnosis was rejected      |
+      | body      | Lisa Cuddy has rejected your request to join the "Medical diagnosis" collection |
```
pfrenssen: What happens if a membership of an active member is deleted? This test only covers if a pending membership is deleted, but active ones can also be deleted through the membership management.

I'm asking because this notification only makes sense in context of a pending membership. Is the deletion of an active membership handled in a separate ticket?

PR #695, opened 2017-07-28
Isaicp 3657: Implement the legal notice page.

idimopoulos:


tests/features/joinup_core/page_access.feature
```diff
@@ -15,3 +15,8 @@ Feature:
     And I should see the text "The page you are looking for does not exist; It may have been moved, or removed altogether. You might want to:"
     And I should see the link "search function"
     And I should see the link "home page"
+
+  Scenario: Legal notice page is accessible by anyone.
+    When I am not logged in
+    And I visit "/joinup/legal-notice"
+    Then I should not see the heading "Page not found"
```
pfrenssen: Lol this is a bit of a poor test :D

Anyway, if you think it is enough to just test access this should be moved to `authentication.feature` and you should also add tests for authenticated users. Now you only know that anonymous users can see it, it might not be accessible by authenticated users.

claudiu-cristea:
I can't believe that we are hardcoding this page as a route with its own controller and with the content as template :(

sandervd:
The controller will suffice for now:
All pages that should be alterable by the moderator can go as custom pages in the Joinup collection later.

PR #697, opened 2017-07-28
ISAICP-3618: Changes in tiles

tom-waat:


web/themes/joinup/templates/content/node--custom-page--listing-item.html.twig
```diff
@@ -54,6 +54,7 @@
  * - title_suffix: Additional output populated by modules, intended to be
  *   displayed after the main title tag that appears in the template.
  * - view_mode: View mode; for example, "teaser" or "full".
+ * - view_mode: View mode; for example, "teaser" or "full".
```
kalinchernev: Seems like the same line of 56?

web/modules/custom/collection/config/install/core.entity_view_display.rdf_entity.collection.view_mode_tile.yml
```diff
@@ -32,8 +32,10 @@ targetEntityType: rdf_entity
 bundle: collection
 mode: view_mode_tile
 content:
-  field_ar_description:
-    weight: 2
+  field_ar_abstract:
```
tom-waat: For now it I think that we should not remove abstract from a tile view mode, because it is not clear if there will be featured view mode or maybe featured tiles will be done in a different way. For now abstract is not displayed in hidden view mode:
`{% if view_mode == 'view_mode_featured' %}`

tom-waat:
> There are still some view modes which contain description field (I re-built and re-installed from scratch)
> Featured:
> 
>     document
>     event
>     solution

As we discussed we don't check featured view mode (but on the wireframes the description is visible there).

>     Tile view mode:
>     event
>     licence
>     solution

I checked it and in event and solution the teaser is hidden if it is not featured view mode:
```
    {% if view_mode == 'view_mode_featured' %}
      <div class="listing__field">
        <p class="listing__teaser">{{ content.field_is_description|render|striptags }}</p>
      </div>
    {% endif %}
```

The licence is missing on the wireframes, so I didn't work on it.

PR #706, opened 2017-08-01
ISAICP-3651: Hide title on overview pages

tom-waat:


web/themes/joinup/joinup_theme.theme
```diff
@@ -130,6 +130,13 @@ function joinup_theme_preprocess_page_title(&$variables) {
   }
   $route_name = $route_match->getRouteName();
 
+  $variables['visually_hidden_class'] = '';
+
+  // Early bail out.
+  if (empty($route_name)) {
```
kalinchernev: This can be even before `$variables['visually_hidden_class'] = '';`, but it's ok.

PR #711, opened 2017-08-01
ISAICP-3578: Fix dropdown button

tom-waat:


web/themes/joinup/prototype/sass/_drupal-hacks.sass
```diff
@@ -182,3 +182,29 @@ form .description
     &:before
       +icon-size(15px)
       content: '\e617'
+
+// Additional styling for dropbutton.
```
kalinchernev: The first button does not receive classes in the same way as the second one, which makes it unnoticeable when tabbing:

edit_buttons

kalinchernev: Mobile is virtually un-usable:
bs_realdroid_mobile_google nexus 6p-7 0-1080x1920

PR #714, opened 2017-08-02
ISAICP-3574: Move local tasks to the featured region

tom-waat:


web/themes/joinup/templates/navigation/menu-local-tasks.html.twig
```diff
@@ -12,10 +12,16 @@
  */
 #}
 {% if primary %}
+  <button id="local-tasks-primary" class="mdl-button mdl-js-button mdl-button--icon">
```
kalinchernev: Please check if the check can be simpler such as:

```
  {% set type = (primary ? 'primary' : 'secondary') %}
  <button id="local-tasks-{{ type }}" class="mdl-button mdl-js-button mdl-button--icon">
    <span class="icon icon--dots"></span>
  </button>
  <h2 class="visually-hidden">{{ 'Tabs navigation'|t }}</h2>
  <ul class="mdl-menu mdl-menu--bottom-right mdl-js-menu mdl-js-ripple-effect" for="local-tasks-{{ type }}">{{ primary ?: secondary }}</ul>
```

The difference seems small enough to allow a shorthand.

tom-waat: I didn't find an example when secondary links are displayed, but if it is in default drupal template I don't won't to remove it. I think that in your code if there are primary and secondary links, then only primary would be displayed (because there is if .. if, not if .. elseif )

web/themes/joinup/prototype/sass/_mdl-hacks.sass
```diff
@@ -46,10 +46,26 @@
...
 
-.join-collection-form
+.join-collection-form, .block-local-tasks-block
   .mdl-menu__item a
     white-space: nowrap
     color: $color-blue-default
   .mdl-menu__container
     margin-top: $field-margin
 
+.block-local-tasks-block
```
tom-waat: This is styling related to mdl dropdown menu, so maybe it should stay in _mdl-hacks.sass

web/themes/joinup/templates/navigation/menu-local-tasks.html.twig
kalinchernev: You're right!

web/themes/joinup/prototype/sass/_mdl-hacks.sass
kalinchernev: I'm ok with that, as long as you it's ok with you.

PR #716, opened 2017-08-02
ISAICP-2219: pinned box on top to show collections

brummbar:


web/modules/custom/joinup_community_content/src/Controller/PinContentController.php
```diff
@@ -0,0 +1,182 @@
...
+   */
+  protected function isFacilitatorInParentCollection(NodeInterface $node, AccountInterface $account) {
+    $collection = $this->relationManager->getParent($node);
+
+    if (!$collection || $collection->bundle() !== 'collection') {
+      return FALSE;
+    }
+
+    $membership = $this->membershipManager->getMembership($collection, $account);
+    return !empty($membership) && $membership->hasRole('rdf_entity-collection-facilitator');
```
claudiu-cristea: Oh, no! Again and again we are repeating this wrong way to check the access. We should never ever, but never, check against the role. Starting from version 4 (or even 3), Drupal has introduced this layer or permissions that allows a higher flexibility. For example, with the actual code, the super-admin (uid1) is not able to pin/unpin content because he's not a group facilitator. But if a permission is checked, rather than a role ID, super-admin has, by default, all the permissions even are not granted explicitly.

So here we need to:
- Create a new collection OG permission: `pin community content` (we can use this permission also for unpinning).
- Grant this permission to collection owner and facilitator OG roles.
- Check here against the new permission. I think `\Drupal::service('og.access')->userAccess($collection, 'pin community content')` is doing the job (but injected). Rename this function to reflect the changes.

web/modules/custom/joinup_community_content/js/pinned_content.js
```diff
@@ -0,0 +1,50 @@
...
+(function ($, Drupal) {
+
+  /**
+   * Theme function that renders a visual cue for pinned content.
+   *
+   * @returns {string}
+   *   The HTML for the pinned content visual cue.
+   */
+  Drupal.theme.pinnedContentCue = function () {
+    return '<div class="pinned-markup"><div class="icon icon--pin"></div></div>';
```
brummbar: A node can be pinned only inside its collection, so there is no need to have an explicative tooltip.

claudiu-cristea:
@brummbar, but what about collection owner OG role? I see that the ticket description says:

> As a collection owner or facilitator, I must be able to PIN a content item [...]

Either the description is accurate, then we'll have to:

- Add the permissions also to the `rdf_entity-collection-administrator` OG role
- Add a case in the Behat test

Or there was a decision to restrict only to facilitator. In this case let's update the description.

brummbar:
While we often use the word "owner" in Joinup, there is no owner role. The "owner" pseudo-role is granted by having both the facilitator and administrator roles. In terms of permissions, an "owner" can do all the actions allowed to a facilitator and in addition he can request the deletion of a collection.
As we discussed multiple times in the last year, it was decided in Joinup to work with the sum of permissions of the roles. So the administrator has only the permission to request the deletion.
That's why the permission has been added only in that role and that's why in the majority of Joinup tests we cover specifically only the facilitator, as it implies the "owner" too.

claudiu-cristea:
Ok, ok, it makes sense. Everytime I'm forgetting about the role complementarity in terms of permissions (probably because it's a wrong design!).

The changes look good. Merging...

PR #717, opened 2017-08-02
Isaicp 3662: Re publish of group should re index children.

idimopoulos:


web/modules/custom/collection/collection.module
```diff
@@ -144,18 +144,18 @@ function collection_entity_bundle_field_info_alter(&$fields, EntityTypeInterface
...
+  if ($entity->bundle() !== 'collection') {
     return;
   }
-  collection_entity_update($entity);
+  collection_rdf_entity_update($entity);
 
   // When a new collection is proposed, give the user the 'group owner' role.
   // This will not occur if the user is the anonymous, i.e. the entity was
   // created by some batch process or some importing.
-  if (Og::isGroup($entity->getEntityTypeId(), $entity->bundle()) && !empty($entity->getOwner())) {
```
idimopoulos: We can if you want to, but this hook is already called only for rdf_entities and we are further filtering that it is a collection! We already know it's a group as this is a project specific functionality! Do you want me to add a check?

PR #718, opened 2017-08-03
Isaicp 2746: Apply unique title descriptions

pfrenssen:


web/modules/custom/asset_distribution/asset_distribution.module
```diff
@@ -284,35 +284,44 @@ QUERY;
...
-    ->condition('rid', 'asset_distribution')
-    ->condition('og_audience', $asset_distribution->get('og_audience')->first()->getValue()['target_id']);
-  if (!empty($asset_distribution->id())) {
-    $query->condition('id', $asset_distribution->id(), '<>');
+  // If the distribution is now added, get the parent from the route.
+  elseif (\Drupal::routeMatch()->getRouteName() === 'asset_distribution.asset_release_asset_distribution.add') {
+    $parent = \Drupal::routeMatch()->getParameter('rdf_entity');
+    if ($parent instanceof RdfInterface && in_array($parent->bundle(), ['solution', 'asset_release'])) {
+      return $parent;
+    }
```
pfrenssen: This doesn't seem right, we shouldn't get the parent from the route parameters. It should be present on the entity even if it is new. If it is not present then we should implement an `#entity_builder` callback that populates the value.

After all not all distributions are created through the form, they might be federated, imported, or created during tests.

web/modules/custom/solution/solution.module
```diff
@@ -585,13 +585,6 @@ function solution_get_collection_ids(RdfInterface $solution) {
 }
 
 /**
- * Implements hook_ENTITY_TYPE_insert().
- */
-function solution_rdf_entity_insert(EntityInterface $entity) {
-  solution_invalidate_collection($entity);
-}
```
pfrenssen: Good find! This is not needed.

web/modules/custom/asset_distribution/asset_distribution.module
pfrenssen: I had a look at how this can be fixed properly:

1. We can put a computed field on the distribution that contains the reference to the parent, similar to the one we put on solutions. It will get the value from the database for existing entities, and from a property on the entity object for new entities.
2. An `#entity_builder` firing in the form submission can populate the value on the computed field. This will cover the standard case of creating a distribution through the UI.
3. Other ways of creating (through federation or when running tests) can populate the value in a similar way, e.g. by setting a property on the entity that is used by the computed field.

This seems like a lot of work to block this ticket on. I will let this pass for now. Maybe we can throw an exception at this point if the value is not present on the route? That will alert us immediately if we are using this incorrectly.

PR #719, opened 2017-08-03
3555: The video should not start automatically

claudiu-cristea:


claudiu-cristea:
There's a random Behat failure.

claudiu-cristea:
This was QAed but, by mistake, was not merged.

PR #728, opened 2017-08-04
Revert "Isaicp 3677: Apply an access check to MenuSubPages block"

idimopoulos:
Reverts ec-europa/joinup-dev#726


claudiu-cristea:
We don't need this anymore.

PR #739, opened 2017-08-07
Isaicp 3659: Change Log in to Sign in

idimopoulos:


idimopoulos:
This fix seems to work. This already integrates the test from https://github.com/ec-europa/joinup-dev/pull/743 so the other one can be closed.

PR #741, opened 2017-08-08
WIP: Attempt to skip user invalidation on membership update.

idimopoulos:


idimopoulos:
This experiment failed. More of this will be checked in https://webgate.ec.europa.eu/CITnet/jira/browse/ISAICP-3698

PR #745, opened 2017-08-08
ISAICP-3679: Style news page

tom-waat:


web/themes/joinup/templates/content/page-title--node--discussion.html.twig
```diff
@@ -17,7 +17,7 @@
   <div class="page__title-wrapper">
       <h1 class="page__title">{{ title }}</h1>
       {% if permalink %}
-          <a class="permalink" href="{{ permalink }}" title="Permalink">#</a>
+        <a class="permalink" href="{{ permalink }}" title="Permalink"><span class="icon icon--anchor"></span></a>
```
kalinchernev: I think it will be good to include some text in the link for screen readers in addition to the icon.

web/themes/joinup/templates/content/page-title--node--news.html.twig
```diff
@@ -15,16 +15,16 @@
 {{ title_prefix }}
 {% if title %}
   <div class="page__title-wrapper">
-      <h1 class="page__title">{{ title }}</h1>
+    <div class="details">
+      <div class="details__element details__element-headline">
+        <div class="value gray">{{ title }}</div>
```
kalinchernev: Hi @tom-waat I was reviewing latest changes and I think it's good to go comparing with Andrea's requirements here.
However, tests are failing because the headings are having changed file structures like following:

Success, as expected:
passing

Failing tests, because the text is not in h1, h2, etc. tag:
failling

Do you think the markup should be updated to match the test assertions or rather the other way around?

Apart from these changes we need to handle, I'd like to merge this one.

PR #748, opened 2017-08-09
ISAICP-3619: header bar improvements

luke-waat:
Rebuild is required.


web/modules/custom/joinup_core/templates/statistics-block.html.twig
```diff
@@ -33,6 +23,16 @@
     </a>
     <div class="stats__description">Collaborate around thematics</div>
   </div>
+  <div class="stats__wrapper stats__wrapper__solutions">
```
kalinchernev: Please include a translate filter to strings in this template to make them accessible for localization.

web/themes/joinup/prototype/sass/components/_search-bar.sass
```diff
@@ -57,39 +55,57 @@
     color: lighten($color-grey-medium, 10%)
 
 .search-bar__input-container
-  margin-left: 40px
+  //width: calc(100% - 50px)
```
kalinchernev: Unnecessary lines could be removed.

web/modules/custom/joinup_core/templates/statistics-block.html.twig
luke-waat: Done.

web/themes/joinup/prototype/sass/components/_search-bar.sass
luke-waat: Done.

PR #749, opened 2017-08-09
ISAICP-3676: Fix grids implemented in blocks

tom-waat:


web/themes/joinup/joinup_theme.theme
```diff
@@ -1052,6 +1052,22 @@ function joinup_theme_theme_suggestions_container_alter(array &$suggestions, arr
       $suggestions[] = 'container__edit_actions__search_bar';
     }
   }
+
+  // Early bail out.
+  if (!isset($variables['element']['#extra_suggestion'])) {
```
kalinchernev: Please make like this:
```
/**
 * Implements hook_theme_suggestions_HOOK_alter() for container.html.twig.
 */
function joinup_theme_theme_suggestions_container_alter(array &$suggestions, array $variables) {
  $has_extra_suggestions = isset($variables['element']['#extra_suggestion']);

  if (isset($variables['element']['#id'])) {
    $suggestions[] = 'container__' . str_replace('-', '_', $variables['element']['#id']);

    // Suggestion for search form submit container.
    if (isset($variables['element']['submit']['#id']) && strpos($variables['element']['submit']['#id'], 'edit-submit-search') !== FALSE) {
      $suggestions[] = 'container__edit_actions__search_bar';
    }
  }

  if ($has_extra_suggestions) {
    switch ($variables['element']['#extra_suggestion']) {
      case 'container_grid':
        $suggestions[] = 'container__grid';
        break;

      case 'container_grid_item':
        $suggestions[] = 'container__grid_item';
        break;
    }
  }
}
```

This way:
- you don't stop the function preventing others to include further logic
- you make the check very visible and readable
- you make the check reusable for others later

web/themes/joinup/joinup_theme.theme
```diff
@@ -1117,6 +1133,16 @@ function joinup_theme_theme_suggestions_details_alter(array &$suggestions, array
 }
 
 /**
+ * Implements hook_theme_suggestions_HOOK_alter() for block.html.twig.
+ */
+function joinup_theme_theme_suggestions_block_alter(array &$suggestions, array $variables) {
+  // Provide suggestion for blocks which have top padding and border.
+  if (isset($variables['elements']['content']['#extra_suggestion']) && $variables['elements']['content']['#extra_suggestion'] == 'separated_block') {
+    $suggestions[] = 'block__separated';
```
kalinchernev: Let's choose one way and make it consistent? Either `separated__block` or `block__separated`?

web/themes/joinup/templates/form/container--grid.html.twig
```diff
@@ -0,0 +1,34 @@
+{#
```
kalinchernev: Are you sure container templates should be in the `form` folder?

web/modules/custom/joinup_news/config/install/core.entity_view_display.node.news.default.yml
```diff
@@ -20,11 +20,29 @@ dependencies:
   module:
     - comment
     - ds
+    - field_group
     - file
     - link
-    - options
     - text
     - user
+third_party_settings:
```
kalinchernev: Can you please help me connect this change to which part of the ticket?

web/themes/joinup/templates/block/block--separated.html.twig
```diff
@@ -0,0 +1,45 @@
+{#
```
kalinchernev: The markup of this template is the same as the original blocks, are you sure a preprocess can't attach the necessary classes without a template?

web/modules/custom/solution/solution.module
```diff
@@ -421,14 +421,7 @@ function solution_build_related_solutions_list(array $ids) {
...
       '#type' => 'container',
-      '#attributes' => [
-        'class' => [
-          'listing__item',
-          'listing__item--tile',
-          'mdl-cell',
-          'mdl-cell--4-col',
-        ],
-      ],
+      '#extra_suggestion' => 'container_grid_item',
```
kalinchernev: Hi @tom-waat adding reply to this note in the jira ticket

> Related solutions - this field is implemented in differently than other entity referenced fields, so I wasn't able to implement block styling there (grid is implemented). In fact in solution manage display there are two similar fields: "Related solutions: explicit + EIRA" and "Related Solutions". Only the first, which is not implemented correctly as entity referenced field, is displayed. I think that is needs to be reviewed.

Spoke with @sandervd, please feel free to modify anything in the `solution_build_related_solutions_list()` to re-use the helpers as in the other entity reference field instances.

web/themes/joinup/templates/form/container--grid.html.twig
tom-waat: Original container.html.twig is in /core/themes/stable/templates/form, usually I place templates in the same folder as it is in the base theme.

web/modules/custom/joinup_news/config/install/core.entity_view_display.node.news.default.yml
tom-waat: It is about adding entity_reference__listing_tile suggestion to entity reference fields (it adds top margin and border to these fields)

web/themes/joinup/templates/block/block--separated.html.twig
tom-waat: There are added new classes and title has different tag, so there are several changes. But if needed I can move it to preprocess.

web/modules/custom/solution/solution.module
tom-waat: Ok, I change type of this block from form_element to container and I added there the same markup as for other blocks.

web/themes/joinup/joinup_theme.theme
```diff
@@ -1052,6 +1054,16 @@ function joinup_theme_theme_suggestions_container_alter(array &$suggestions, arr
       $suggestions[] = 'container__edit_actions__search_bar';
     }
   }
+
+  // Early bail out.
```
kalinchernev: Please remove this, it's not needed, as the check already checks this, and it's not useful to exit function.

web/themes/joinup/joinup_theme.theme
```diff
@@ -1052,6 +1054,16 @@ function joinup_theme_theme_suggestions_container_alter(array &$suggestions, arr
       $suggestions[] = 'container__edit_actions__search_bar';
```
kalinchernev: I'm not sure this suggestion is necessary.
Also, `strpos($variables['element']['submit']['#id'], 'edit-submit-search') !== FALSE` could possibly a positive check to be shorter. Just please make sure this is actually needed at all.

web/themes/joinup/templates/block/block--separated.html.twig
kalinchernev: Yes please, let's keep it simple with a function as long as only classes are managed.

web/themes/joinup/joinup_theme.theme
tom-waat: I removed it, it was an oversight.

web/themes/joinup/joinup_theme.theme
tom-waat: This suggestion is used on advanced search page.

web/themes/joinup/templates/block/block--separated.html.twig
tom-waat: Ok, template is removed, code is added in preprocess function.

kalinchernev:
@tom-waat thanks!

PR #756, opened 2017-08-11
ISAICP-3671: Implement document page

tom-waat:


web/themes/joinup/templates/content/page-title--node--document.html.twig
```diff
@@ -0,0 +1,35 @@
...
+{{ title_prefix }}
+{% if title %}
+  <div class="page__title-wrapper">
+    <h1 class="page__title">{{ title }}</h1>
+    {% if permalink %}
+      <a class="permalink" href="{{ permalink }}" title="Permalink"><span class="icon icon--anchor"></span></a>
+    {% endif %}
+    <div class="details">
+      <div class="details__element details__element-headline">
+        <div class="value gray">{{ additional_content.field_short_title }}</div>
```
kalinchernev: Please unify with the way the News was made with a heading.

web/themes/joinup/templates/content/page-title--node--document.html.twig
```diff
@@ -0,0 +1,35 @@
...
+ * - title: The page title, for use in the actual content.
+ * - title_suffix: Additional output populated by modules, intended to be
+ *   displayed after the main title tag that appears in the template.
+ */
+#}
+{{ title_prefix }}
+{% if title %}
+  <div class="page__title-wrapper">
+    <h1 class="page__title">{{ title }}</h1>
+    {% if permalink %}
```
kalinchernev: Regression:

permalink_regression

web/themes/joinup/joinup_theme.theme
```diff
@@ -456,10 +456,11 @@ function joinup_theme_preprocess_block(&$variables) {
         $node = $route_match->getParameter('node');
         if (!empty($node)) {
           $node_types = [
-            'news',
-            'event',
-            'discussion',
```
kalinchernev: :)

web/themes/joinup/joinup_theme.theme
```diff
@@ -833,6 +834,30 @@ function joinup_theme_preprocess_page_title__node__discussion(&$variables) {
 }
 
 /**
+ * Implements hook_preprocess_page_title__node__document().
```
kalinchernev: The hook is not really `hook_preprocess_page_title__node__document()` I believe

web/themes/joinup/joinup_theme.theme
```diff
@@ -833,6 +834,30 @@ function joinup_theme_preprocess_page_title__node__discussion(&$variables) {
...
+ */
+function joinup_theme_preprocess_page_title__node__document(&$variables) {
+  /** @var \Drupal\node\NodeInterface $node */
+  $node = $variables['entity'];
+
+  /* @var \Drupal\Core\Entity\EntityViewBuilderInterface $viewBuilder */
+  $viewBuilder = \Drupal::entityTypeManager()->getViewBuilder('node');
```
kalinchernev: Are you sure you can't reuse one invocation for all the values at once?
`$viewBuilder = \Drupal::entityTypeManager()->getViewBuilder('node');` should be be giving the same results.

I think that would be sufficient:

```php
  /** @var \Drupal\node\NodeInterface $node */
  $node = $variables['entity'];

  /* @var \Drupal\Core\Entity\EntityViewBuilderInterface $viewBuilder */
  $viewBuilder = \Drupal::entityTypeManager()->getViewBuilder('node');

  $field_document_publication_date = $viewBuilder->viewField($node->field_document_publication_date, 'default');
  $variables['additional_content']['field_document_publication_date'] = $field_document_publication_date;

  /* @var \Drupal\Core\Entity\EntityViewBuilderInterface $viewBuilder */
  $field_short_title = $viewBuilder->viewField($node->field_short_title, 'default');
  $variables['additional_content']['field_short_title'] = $field_short_title;

  /* @var \Drupal\Core\Entity\EntityViewBuilderInterface $viewBuilder */
  $author = $node->get('uid')->entity;
  $author_inline = $viewBuilder->view($author, 'inline');
  $variables['additional_content']['author'] = $author_inline;
```

PS I haven't tried it though, please double-check

kalinchernev:
I spent some time trying to add a test for the document content type, but couldn't.
The following error:
```
The website encountered an unexpected error. Please try again later.
TypeError: Argument 2 passed to simplenews_node_view() must implement interface Drupal\node\NodeInterface, instance of Drupal\user\Entity\User given in simplenews_node_view() (line 71 of modules/contrib/simplenews/simplenews.module).
```
It appears both on `@Given I am viewing a/an :type( content):` and `And document content:` style of creating content. The content created by those 2 is `Drupal/node/Entity/Node` whereas `simplenews_node_view()` expects `NodeInterface`. I am not sure if there is a difference, but at any case, the document created by behat I can't view at the moment.

For reference, here's the scenario I tried with last:

```
  Scenario: Document page visualization.
    Given the following licence:
      | title       | Beer licence                                     |
      | description | Offer a beer to the developer when you meet him. |
      | type        | Attribution                                      |
    And the following collection:
      | title | Traveller tools |
      | state | validated       |
    Given I am viewing a "document" content:
      | title            | VAT refund sample document              |
      | document type    | document                                |
      | short title      | VAT refund fac-simile                   |
      | file type        | upload                                  |
      | file             | text.pdf                                |
      | body             | Valid for people living outside the EU. |
      | collection       | Traveller tools                         |
      | licence          | Beer licence                            |
      | state            | validated                               |
    Then show last response
```
I'm not sure how to create the date correctly, but at the moment that shouldn't be the problem as the content type is not a enabled to be a simplenews content type.

Will possibly need help from back-end to make a test for this.
CC @brummbar

web/themes/joinup/joinup_theme.theme
tom-waat: Hmm, so its should be only Implements hook_preprocess_page_title() ?

web/themes/joinup/joinup_theme.theme
tom-waat: I removed one $viewBuilder = \Drupal::entityTypeManager()->getViewBuilder('node');,
but the third view builder is for user not for node.

web/themes/joinup/joinup_theme.theme
kalinchernev: Most of the inherited hooks like this are https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Render%21theme.api.php/function/hook_preprocess_HOOK/8.2.x

web/themes/joinup/joinup_theme.theme
kalinchernev: Alright, I didn't notice this small details, thanks

web/themes/joinup/joinup_theme.theme
tom-waat: Ok, it is changed, but maybe we could create ticket to review and fix documentation in other preprocess functions.

PR #767, opened 2017-08-16
ISAICP-3675: Create new template for vertical tabs

tom-waat:


web/themes/joinup/templates/form/details.html.twig
```diff
@@ -14,24 +14,26 @@
...
 <details{{ attributes }}>
+  {%
+    set summary_classes = [
+      required ? 'js-form-required',
+      required ? 'form-required',
+      'form__label',
+    ]
+  %}
   {%- if title -%}
-    <summary{{ summary_attributes }}>{{ title }}</summary>
```
kalinchernev: Was checking the [whitespace control](https://twig.symfony.com/doc/2.x/templates.html#whitespace-control) in the operator, can you also please add it in the title if the control is really necessary?

PR #760, opened 2017-08-14
ISAICP-3723: take back necessary changes

kalinchernev:
Changes from:
- #737
- #748
- #751

Were taken out in #752. This pull adds them back.

PR #768, opened 2017-08-16
Isaicp 3661: Revert wrong merge that reverted many PRs.

idimopoulos:
Some of them where already reverted in https://github.com/ec-europa/joinup-dev/pull/760

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -9482,7 +9482,11 @@
                 },
                 "drupal": {
                     "version": "8.x-1.0-alpha6+0-dev",
-                    "datestamp": "1493675943"
+                    "datestamp": "1493675943",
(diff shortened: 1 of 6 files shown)


pfrenssen:
This is removing too much, amongst other things this removes the test coverage that checks the race condition that was discovered in RDF Entity last week.

We should simply revert the merge conflict that caused the problem, and not touch unrelated commits.

pfrenssen:
Updated the branch with the correct fix. This was made more difficult because this was partially fixed by manually reapplying parts of the lost work in #760. So this couldn't be reverted any more. Fixed it by reapplying the original code by cherry-picking the merge commit:

```
$ git cherry-pick 2524c1a -m 1
`

PR #771, opened 2017-08-17
ISAICP-3713

luke-waat:


web/themes/joinup/prototype/sass/components/_form-table.sass
```diff
@@ -36,13 +44,53 @@ thead th.form-table__header
...
+      min-height: auto
+      margin: 0
+      padding: 0
+    .dropbutton
+      display: flex
+    .dropbutton-widget
+      position: static
+    .dropbutton-toggle
+      display: none
+      visibility: none
```
kalinchernev: Should be `visibility: ninja` or `visibility: hidden`, although I don't see why adding it when `display: none`

web/themes/joinup/prototype/sass/components/_form-table.sass
```diff
@@ -36,13 +44,53 @@ thead th.form-table__header
...
+    .dropbutton-action
+      position: static
```
kalinchernev: What does this line do?

web/themes/joinup/prototype/sass/components/_form.sass
```diff
@@ -98,5 +99,33 @@
 .form__description--checkbox
   margin-left: 25px
 
+// Dependencies
+// --------------------------------------------------
+.form__actions
```
kalinchernev: What are these changes related to? If they are the ones from the drupal hacks, why moving them here as they are still very specific to the drupal markup?

web/themes/joinup/prototype/sass/components/_form-table.sass
luke-waat: Done. It was duplicate, `display: none` left to avoid additional space leftovers.

web/themes/joinup/prototype/sass/components/_form-table.sass
luke-waat: My job was to style action links in the `form_table`. Unfortunately these links have dropdown markup which couldn't be changed, so I had to reset the styling and adjust links to requirements. In order to do so, first of all, to avoid overwriting of the already overwritten styles, I added a dependency to the existing styles for dropdown. These styles have been used in case of actions buttons in `form__actions` element which is completely different. After that I have added second dependency of the dropdown, but this time for `form__table`.

In the meantime, working on the `form__table`, I realised that this component is used in many places, some of the styling are duplicated and there is a chance for regression, so I decided to make minor adjustments and unifications.

web/themes/joinup/prototype/sass/components/_form.sass
luke-waat: I split the styles between `form` and `form-table` because I did not want to keep these styles in one place because it was completely different components but with the same markup. If you would like I can move all the styling to `_drupal-hacks` file, there is no problem.

PR #773, opened 2017-08-17
ISAICP-3697: banners text transform fix: sass

luke-waat:


luke-waat:
@kalinchernev we are already using shared classes like _bold_, _italic_, so I added new one: _uppercase_. In my opinion it might be more useful than adding modifier to elements --bold, --italic or --uppercase. What do you think?

kalinchernev:
@luke-waat on naming, anything is ok with me as long as you feel ok with it. The important task at hand in this issue is that one area should not be uppercase any more, whereas others should remain uppercase.

PR #775, opened 2017-08-17
Isaicp 3497: UAT remarks on the collection notifications

idimopoulos:
This PR has also two extra fixes. One is related to converting all links into a tags (commit 4ee81ef37218e362fd7e6d0fd2e4c93e486e7685) and the other one related to a bug which resulted html markup to end up in the email (commit 121a756)


web/modules/custom/joinup_notification/config/install/message.template.col_arc_del_no_request.yml
```diff
@@ -6,7 +6,7 @@ label: 'Collection archive/delete without request'
 description: 'Template id 13 - This will be sent in case the deletion/archival happened without request'
 text:
   1:
-    value: "<p>Your @entity:bundle \"@entity:title\" has been @transition:request_action:past.</p>\r\n\r\n<p>The reason for being @transition:request_action:past is: @transition:motivation.</p>\r\n"
+    value: "<p>Your @entity:bundle \"@entity:title\" has been @transition:request_action:past.</p>@transition:archive:extra:owner"
```
idimopoulos: This change is because in case of a delete, the message that was removed should be removed as well until we have the delete confirmation page.

PR #781, opened 2017-08-18
ISAICP-3446: Use Euro style date formats.

pfrenssen:


composer.json
```diff
@@ -31,7 +31,7 @@
         "drupal/migrate_run": "~1.0-alpha4",
         "drupal/migrate_spreadsheet": "~1.0-beta4",
         "drupal/og": "dev-ISAICP-2709 as dev-1.x",
-        "drupal/og_menu": "~1.0@alpha",
+        "drupal/og_menu": "^1.0-alpha3",
```
pfrenssen: This commit is actually for ISAICP-3688 but I committed it here by mistake, I mixed up my branches. It is harmless. The core update will be unlocked in ISAICP-3755.

PR #786, opened 2017-08-22
ISAICP-3733: Style the social icons and inputs

tom-waat:


web/themes/joinup/prototype/sass/components/_featured.sass
```diff
@@ -141,6 +141,28 @@
   margin: 10px
   z-index: 99
 
+// Style social media links for user
+.featured__social-media
```
kalinchernev: Hi @tom-waat the visualization is really ok, both in the back-end and in the front-end.
Please correct the following and the pull request will be merge-able:
- follow BEM naming conventions on this one: you definitely have access to all `ul`, `li` and `a` tags in the template. Use this opportunity to make it right as `social-media__list`, `social-media__item` and `social-media__link`.
- i don't see a reason why you would name this pattern under featured, as this way of visualization could be ok on any other place out of the featured context

tom-waat: There is already social media component in _footer-social-media.sass, but the styling is completely different, so I decided to create a new component - inline-social-media

PR #787, opened 2017-08-22
ISAICP-3700: Implement dots styling for all contextual links

tom-waat:


web/themes/joinup/prototype/sass/_drupal-hacks.sass
```diff
@@ -81,24 +82,47 @@ textarea
...
+    left: 0
+  .contextual .contextual-links
+    border-radius: 0 4px 4px 4px
+    float: left
+    left: 0
+    right: auto
+
+
+ul.contextual-links
+  +z-index(highlight+2)
```
kalinchernev: screenshot from 2017-08-25 11-46-54

tom-waat: It is fixed. But probably we need a ticket for z-index fixes, because sometimes dropdown menus in header are overlapping on smaller screens.

kalinchernev: I tend to disagree about having another ticket as the z-index-es were refreshed really recently in ISAICP-3692. There, z-indexes got organized into levels which should be consistent enough that we just follow these from this point on. This menu is [an exception](https://webgate.ec.europa.eu/CITnet/jira/browse/ISAICP-3700?focusedCommentId=2285839&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-2285839) as there is a test assuming it's always to be displayed, ok, we accept super high z-index for it. For the rest, it just does not make sense to add new levels on each z-index being wrong. Rather, let's try to fix these contextual links which are too high and allow only the special menu to be always visible.

I honestly don't consider adding a new "just a bit higher" item to the list a fix. Why did we make the list in first place?

tom-waat: The contextual button are ok now, I fixed them in the last commit. Is it not ok to have z-index: contextual? I cannot use z-index "navigation" there because some contextual buttons would not be clickable (ex. in sidebar menu).

> I honestly don't consider adding a new "just a bit higher" item to the list a fix. Why did we make the list in first place?
This z-index was created in ISAICP-3692 and it is used in many places so I think that I should not touch it in this ticket, @luke, what do you think?

kalinchernev: > This menu is an exception as there is a test assuming it's always to be displayed, ok, we accept super high z-index for it. 

I am ok to have super-specific z-index for the sidebar menu contextual nav. Tests are expecting this. For the rest, I'd expect contextual navs to be on the level of navigations. As they are navigations, no?

tom-waat: Ok, I changed z-index for contextual button - it is now "navigation". I added custom z-index values only for button in the sidebar and in the top menu (the second fix is only for admin user).

PR #789, opened 2017-08-22
Isaicp 3680: Restrict kicker length to 30 characters

idimopoulos:


web/modules/custom/joinup_news/joinup_news.module
```diff
@@ -23,3 +24,18 @@ function joinup_news_entity_bundle_field_info_alter(&$fields, EntityTypeInterfac
     }
   }
 }
+
+/**
+ * Implements hook_form_alter().
+ *
+ * Restrict the kicker(title) to 30 characters long.
+ */
+function joinup_news_form_alter(&$form, FormStateInterface $form_state, $form_id) {
```
claudiu-cristea: Why are we coding so much? We should stop re-inventing the wheel. We should use contrib modules that are already doing this job. We should switch from maintaining code to maintaining config as much as we can. There's a good module doing this job: https://www.drupal.org/project/maxlength

PR #799, opened 2017-08-24
Remove patch #2892809

claudiu-cristea:
See https://www.drupal.org/node/2892809


pfrenssen:
Looking good, will wait for test results before merging.

PR #802, opened 2017-08-24
Isaicp 3720

pfrenssen:


web/modules/custom/joinup_user/joinup_user.module
```diff
@@ -534,3 +535,43 @@ function joinup_user_form_user_cancel_form_alter(&$form, FormStateInterface $for
...
+    $account = User::load($account->id());
+  }
+
+  // Default to the account name.
+  $name = $account->getAccountName();
+
+  /** @var \Drupal\user\UserInterface $account */
+  $name_components = [];
+  foreach (['field_user_first_name', 'field_user_family_name'] as $field_name) {
+    if (!empty($account->get($field_name)->first())) {
```
claudiu-cristea: This check has a dedicated API:

```php
if (!$account->get($field_name)->isEmpty()) { ... }
```

web/modules/custom/joinup_user/joinup_user.module
```diff
@@ -534,3 +535,43 @@ function joinup_user_form_user_cancel_form_alter(&$form, FormStateInterface $for
...
+      $name_components[] = $account->get($field_name)->first()->value;
```
claudiu-cristea: Simply: `$account->get($field_name)->value`.

web/modules/custom/joinup_user/src/UserFullNameFieldItemList.php
```diff
@@ -22,17 +22,7 @@ protected function createItem($offset = 0, $value = NULL) {
...
-
-    if (empty($first_name) || empty($family_name)) {
-      $value = $entity->getDisplayName();
-    }
-    else {
-      $value = "$first_name $family_name";
-    }
-
-    return parent::createItem($offset, $value);
+    return parent::createItem($offset, joinup_user_get_display_name($entity));
```
claudiu-cristea: More clean and more understandable from my point of view:

```php
return parent::createItem($offset, $entity->getDisplayName());
```
The procedural function is only a helper but `->getDisplayName()` is part of the User entity contract and we should use it.

PR #805, opened 2017-08-25
ISAICP-3767: Changes in a document tile

tom-waat:


web/themes/joinup/joinup_theme.theme
```diff
@@ -641,6 +641,23 @@ function joinup_theme_preprocess_field_multiple_value_form(&$variables) {
...
+function joinup_theme_preprocess_field__comma_separated(&$variables) {
+  // Only in listing modes.
+  $view_modes = [
+    'view_mode_tile',
+    'view_mode_featured',
+  ];
+
+  // Limit number of displayed items.
+  $items_limit = 3;
+  if ($variables['field_name'] == 'field_keywords' && in_array($variables['element']['#view_mode'], $view_modes)) {
```
kalinchernev: Unfortunately, the limitation settings don't apply:

tags_limit

This is with latest version of the branch + latest develop + `drupal cr all` which is reinforced during the demo.

The `items_limit` can be hoisted in the beginning of the function as any other variable definition.

PR #813, opened 2017-08-30
ISAICP-3810: Fix collection title in solution header

tom-waat:


web/themes/joinup/prototype/sass/components/_featured.sass
```diff
@@ -110,6 +110,11 @@
   background-color: #fff
   padding: 0 15px
   +border-radius(3px)
+  color: $color-base
```
kalinchernev: This set of rules in the `.featured__stats` is getting really cheese. You might want to file a ticket to globally refactor this, which I'll be ok for the moment to move faster:
- naming is not ok - you have `featured__stats` and then `stats_something` these are 2 different components.
- the link is too global of a rule
- the inline-block + vertical-align is repeated too many times, not to mention you might reuse a flex added by @luke-waat earlier

In overall, I truly believe these can be simplified ...

PR #825, opened 2017-09-02
Dockerfile: Prepare docker instance

idimopoulos:
This pull request includes:

* Exclude files Dockerfile and docker-compose.yml in project root for git.
* Alter the solr related commands to not require the base path of solr as it is not needed and messes with solr container.


idimopoulos:
I guess this is not valid anymore.

PR #830, opened 2017-09-05
ISAICP-3855: Fix for logic error during migration.

sandervd:


pfrenssen:
This has been closed because this was a workaround for an issue during migration which probably wasn't deemed necessary.

PR #836, opened 2017-09-07
3826: Advanced Search content type relevance

pfrenssen:


web/modules/custom/joinup_search/config/install/search_api.index.published.yml
```diff
@@ -415,13 +416,41 @@ processor_settings:
...
+  type_boost:
+    boosts:
+      'entity:node':
+        datasource_boost: !!float 1
+        bundle_boosts:
+          custom_page: !!float 5
+          discussion: !!float 2
+          document: !!float 0
+          event: !!float 3
+          news: !!float 4
```
claudiu-cristea: While this boost value works correctly it can me misleading when you go to the admin UI, at `/admin/config/search/search-api/index/published/processors`:
boost

This happens because the entered values are not in the standard list of the select dropdown, which is: 0.0,
 0.1, 0.2, 0.3, 0.5, 0.8, 1.0, 2.0, 3.0, 5.0, 8.0, 13.0, 21.0. If somebody will save that form the types with non-standard values will be set to NULL.

I would suggest that we use values from the standard list for `node:news`, `rdf_entity:collection` and `rdf_entity:solution`.

- `node:custom_page`: 8.0.
- `node:news`: 5.0.
- `node:event`: 3.0.
- `node:discussion`: 2.0.

Also "Default boost for items from this datasource" (`datasource_boost`) should be 0.0, according to the ticked.

web/modules/custom/joinup_search/config/install/search_api.index.published.yml
```diff
@@ -415,13 +416,41 @@ processor_settings:
...
+          newsletter: !!float 0
+          video: !!float 0
+      'entity:rdf_entity':
+        datasource_boost: !!float 1
+        bundle_boosts:
+          asset_distribution: !!float 0
+          asset_release: !!float 0
+          collection: !!float 7
```
claudiu-cristea: Same as for `node:news`. This can go with 21.0. And `datasource_boost` should be 0.0 according to the ticket.

web/modules/custom/joinup_search/config/install/search_api.index.published.yml
```diff
@@ -415,13 +416,41 @@ processor_settings:
...
+          licence: !!float 0
+          solution: !!float 6
```
claudiu-cristea: Same as for `node:news`. This can go with 13.0.

web/modules/custom/joinup_search/config/install/search_api.index.published.yml
```diff
@@ -415,13 +416,41 @@ processor_settings:
...
+      'entity:user':
+        datasource_boost: !!float 1
```
claudiu-cristea: This should be 0.0.

PR #841, opened 2017-09-08
ISAICP-3871: Implement content type icon and text in tiles

tom-waat:


web/themes/joinup/templates/content/node--news--listing-item.html.twig
```diff
@@ -90,6 +90,10 @@
       <div class="listing__field listing__tags">{{ content.field_policy_domain }}</div>
     {% endif %}
     <div class="listing__field listing__stats mdl-card__actions">
+      <div class="listing__stat">
```
kalinchernev: Looks ok comparing with the tiles wireframe from the ticket requirements. However, in [ISAICP-3125](https://webgate.ec.europa.eu/CITnet/jira/browse/ISAICP-3125) I think there was an implementation in the shared visual cue, which at least I tested earlier this year. It's missing now:

missing_shared_visual_cue

PR #849, opened 2017-09-13
ISAICP-3800: Add keywords in news tiles

tom-waat:


kalinchernev:
I tried to write a small behat test for the display of the tile, but after struggling to run the tests locally for 2 hours, I give up for now. Checking with the BE team, there is currently a new version of gherkin which yields some errors which will soon be solved.

PR #855, opened 2017-09-13
ISAICP-3883: Changes in distribution tile

tom-waat:


web/themes/joinup/templates/rdf_entity/rdf-entity--asset-distribution--listing-item.html.twig
```diff
@@ -15,11 +15,7 @@ set classes = [
 <div class="{{ classes|join(' ') }}">
   <div class="listing__metadata">
     <div class="listing__field">
-      {% if file_name %}
```
kalinchernev: Thanks, please remove also:
- https://github.com/ec-europa/joinup-dev/pull/855/files#diff-b2909915abaa19a802586179a85c8677R1403
- 4d78fc4

PR #862, opened 2017-09-15
ISAICP-3807: Harmonize tiles height

tom-waat:


web/themes/joinup/joinup_theme.theme
```diff
@@ -1400,6 +1406,12 @@ function joinup_theme_preprocess_rdf_entity(&$variables) {
   if ($variables['view_mode'] == 'about') {
     $variables['is_label_hidden'] = TRUE;
   }
+
+  // Add bundle class in listing templates.
```
kalinchernev: There is a clear repetition for solving the same task. Could be extracted into a function with a parameter being a reference variable to the `$variables`.

web/themes/joinup/templates/content/node--document--listing-item.html.twig
```diff
@@ -86,9 +86,7 @@
       {{ title_prefix }}
       <h2 class="listing__title"><a href="{{ url }}">{{ label }}</a></h2>
       {{ title_suffix }}
-      {% if content.field_type|render %}
```
kalinchernev: Honestly, I don't quite understand why in some places the check is added, such as in `{% if content.field_keywords|render %}` and in others it's removed.

web/themes/joinup/templates/rdf_entity/rdf-entity--asset-distribution--listing-item.html.twig
```diff
@@ -4,21 +4,34 @@
...
 #}
-<div class="listing__card listing__card--detail-container mdl-card">
+{%
+  set classes = [
+    'listing__card',
+    'mdl-card',
+    'listing__card--detail-container',
+  ]
+%}
+{# @todo Implement changes - https://webgate.ec.europa.eu/CITnet/jira/browse/ISAICP-3883 #}
```
kalinchernev: Sorry, but todos are not acceptable. I already marked the ticket as being blocked by those tickets.

web/themes/joinup/templates/rdf_entity/rdf-entity--asset-distribution--listing-item.html.twig
```diff
@@ -4,21 +4,34 @@
...
+      {% if content.field_ad_file_size|render %}
+        <div class="listing__weight listing__single">
+            <div class="label gray">{{ 'Weight'|t }}:</div>
+            <div class="value bold">{{ content.field_ad_file_size }}</div>
+        </div>
+      {% endif %}
+      {% if content.field_ad_licence|render %}
+        <div class="listing__licence listing__single">
+          <div class="label gray">{{ 'Licence'|t }}:</div>
+          <div title="{{ content.field_ad_licence|render|striptags|trim }}" class="value bold underline">{{ content.field_ad_licence }}</div>
```
kalinchernev: This set of classes `value bold underline` is no BEM by any way. Please try to refactor at least the `value` which is way too generic.

kalinchernev:
Adding a reference to the latest display I achieved if I have to show that to someone in the meantime.

all_tiles

video_tile

web/themes/joinup/joinup_theme.theme
luke-waat: Agree that this might be done better, but maybe let's create follow up ticket to review the main template, remove repetitions and create generic, reusable functions.

web/themes/joinup/templates/content/node--document--listing-item.html.twig
luke-waat: Probably Tom added these checks only for fields which are not mandatory to avoid printing empty markup?

web/themes/joinup/templates/rdf_entity/rdf-entity--asset-distribution--listing-item.html.twig
luke-waat: I agree with you that these classes are not perfect, but we should leave it as it is, because they are used across the project. Alteration in one place will not change much, but it might creates confusion. Let's create follow up ticket to sort the generic classes and replace them with BEM like. In my opinion it's better to improve this once and for all.

kalinchernev:
Merged to `UAT-ready` https://github.com/ec-europa/joinup-dev/commit/05647cf4e18ba41ccd9442b7c4ec24affac6b889

kalinchernev:
Will resolve the conflicts here

PR #865, opened 2017-09-15
ISAICP-3834: Fix share box styling

tom-waat:


kalinchernev:
The visualization is ok, only 2 small remarks while making the funcitonal tests:
- There is some scrolling which seems wrong in mobile:
scroll
- Button area is too big
share_button

I've made a small update in the label of the share button.

luke-waat:
- scrolling effect has been removed;
- that large area comes from MDL and it is responsible for ripple effect after clicking on the button; it does not increase the clickable surface.

web/themes/joinup/prototype/sass/components/_share-box.sass
```diff
@@ -1,60 +1,90 @@
...
     min-width: $screen-sm
 
+.share-box__title
+  margin-top: 0
+  color: $color-base
+
+.share-box__social
+  margin-bottom: 20px
+  +breakpoint(sm)
+    display: inline-block
```
kalinchernev: Not a stopper, though it's probably unnecessary to use `inline-block` + width when you have the flexbox.

kalinchernev:
@luke-waat thanks for your updates, only 1 remark - the checkbox effect is not fully visible on mobile:
share_checkbox_limitation
Together with the scrolling it gives an unpolished look, I'd be glad if you could fix it. The rest of the changes are ok.
I know it's not time for perfecting things, but you might check if flexbox will solve the current inline block parts.

web/themes/joinup/prototype/sass/components/_share-box.sass
luke-waat: Can you please take a look if you still can see additional scroll?

kalinchernev:
Thank you @luke-waat !

PR #871, opened 2017-09-16
Isaicp 3901: Restrict pages for the authenticated user.

idimopoulos:


idimopoulos:
This ticket was accepted in UAT. Merging.

PR #873, opened 2017-09-18
ISAICP-3876

luke-waat:


web/themes/joinup/prototype/sass/components/_fieldset.sass
```diff
@@ -55,11 +55,12 @@
 
 .fieldset--border-bottom
   position: relative
-  padding-bottom: 18px
+  padding-bottom: 27px
```
kalinchernev: I'd recommend you separate spacing out of a selector named "border-bottom".
It can be better adding a modifier responsible for spacing.
Some other modifiers such as the `fieldset--box` also have spacing which can be delegated to the same modifier.
In overall, the rules about padding/margin in this component don't look very consistent. This is not a stopper, but if I had to maintain this I would be very afraid to change anything in these rules not being able to follow what's really going on.

web/themes/joinup/templates/content/node--document--full.html.twig
```diff
@@ -74,7 +74,7 @@
   {{ content.body }}
   {{ content.group_categorisation }}
   {{ content.group_attachment }}
-  <div class="action-link__wrapper right">
+  <div class="fieldset action-link__wrapper right">
```
kalinchernev: Do we really need all these fieldsets for adding spacing in the case where these are not really fieldset components only?

web/themes/joinup/templates/field/field--fieldset-wrapper.html.twig
```diff
@@ -0,0 +1,65 @@
...
+ * - field_type: The type of the field.
+ * - label_display: The display settings for the label.
+ *
+ * @see template_preprocess_field()
+ */
+#}
+
+{% if label_hidden %}
+  {% if multiple %}
+    <div{{ attributes.addClass('fieldset') }}>
```
kalinchernev: Ok, here I'm confused. A bit earlier you make a [preprocess function for the global field](https://github.com/ec-europa/joinup-dev/pull/873/files#diff-b2909915abaa19a802586179a85c8677R647) and here you override a template for adding a class on the `attributes` if I understand correctly? Why adding a template for adding classes if you have already a preprocessor being able to manage existing variables?

web/themes/joinup/templates/rdf_entity/rdf-entity--asset-release--full.html.twig
```diff
@@ -7,7 +7,7 @@
 <article{{ attributes }}>
   <div class="page__content">
     {{ content.field_isr_release_notes }}
-    <div class="version-wrapper">{{ content.field_isr_is_version_of }}</div>
+    <div class="fieldset fieldset--version-wrapper">{{ content.field_isr_is_version_of }}</div>
```
kalinchernev: `fieldset--version-wrapper` seems unnecessary and also too specific. Not to speak about the fact there is already a `fieldset` wrapper ...

web/themes/joinup/templates/field/field--fieldset-wrapper.html.twig
luke-waat: Agree that some changes are unclear. I wouldn't touch this changes because I am not sure whether this will not lead to a regression as this was made by Tom. I am quite sure that any possible duplicates have been made due to time pressure. This has to be reviewed.

kalinchernev:
As other pull requests wrongly targeting and using the history of `UAT-ready`, #913 was open as a follow-up towards `develop`.

PR #880, opened 2017-09-19
ISAICP-3890

luke-waat:


kalinchernev:
The title of the wireframe is in uppercase and the actual title is lowercase, but I think it's not a problem.
I think this one is good to be merged when ISAICP-3834 is merged with its comments.

PR #883, opened 2017-09-20
Issue #ISAICP-3879: Disable image resizer.

sandervd:


brummbar:
I have been digging a bit in the [CKEditor source code](https://github.com/ckeditor/ckeditor-dev/blob/fc149eeb4b92ba9ea64c6becee9e0ddff906097f/plugins/image2/plugin.js#L432) and found out that the resizer can be disable by passing  `image2_disableResizer = true` in the format configuration.
See [this gist](https://gist.github.com/brummbar/dc3e02f3820474cc076dba28258946b6) with a practical example on how that works.

PR #886, opened 2017-09-20
ISAICP-3868

luke-waat:


web/themes/joinup/templates/rdf_entity/rdf-entity--asset-release--compact.html.twig
```diff
@@ -28,5 +35,8 @@
         <div class="timeline__listing-item">{{ 'No distributions yet.'|t }}</div>
       {% endif %}
     </div>
+    <div class="timeline__read-more">
+      <a href="{{ url }}">{{ 'Read more'|t }}</a>
```
kalinchernev: Please:
- change to "more"
- uppercase
- all alt/title attributes with title for a11y

web/themes/joinup/templates/rdf_entity/rdf-entity--asset-release--compact.html.twig
```diff
@@ -18,7 +18,14 @@
   <div class="timeline__release-content mdl-cell mdl-cell--9-col">
     <div class="timeline__meta">
       <h2 class="timeline__release-title"><a class="bold" href="{{ url }}" rel="bookmark">{{ full_label }}</a></h2>
-      <p class="timeline__release-abstract">{{ content.field_isr_release_notes }}</p>
+      <div class="timeline__release-abstract">
+
+        {% set abstract = content.field_isr_release_notes|render|striptags %}
+        {{ abstract|length > 200 ? abstract|slice(0, 200) ~ '...' : abstract }}
+
+        <a href="{{ url }}">{{ 'Read more'|t }}</a>
```
kalinchernev: Please change to "more" only and include alt/title attr for a11y.

web/themes/joinup/templates/rdf_entity/rdf-entity--asset-release--compact.html.twig
```diff
@@ -18,7 +18,14 @@
   <div class="timeline__release-content mdl-cell mdl-cell--9-col">
     <div class="timeline__meta">
       <h2 class="timeline__release-title"><a class="bold" href="{{ url }}" rel="bookmark">{{ full_label }}</a></h2>
-      <p class="timeline__release-abstract">{{ content.field_isr_release_notes }}</p>
+      <div class="timeline__release-abstract">
+
+        {% set abstract = content.field_isr_release_notes|render|striptags %}
+        {{ abstract|length > 200 ? abstract|slice(0, 200) ~ '...' : abstract }}
```
kalinchernev: This technique looks really nice, however it does not take into account the fact that the abstract contains words. Please use something [like this](https://www.lullabot.com/articles/trim-a-string-to-a-given-word-count) in order to cut by word count to preserve the words.

luke-waat: You are right, trim on word boundary is not possible in pure twig. I think that we should stay with current, temporary solution and create follow up ticket for the next sprint for Tom. He is more familiar with the project and your code standard. I am sure that he will propose the best solution for this issue, more particular like field preprocess or maybe more generic like creation of an custom formatter for text fields trimming on word boundary.

kalinchernev: Sure, [ISAICP-3918](https://webgate.ec.europa.eu/CITnet/jira/browse/ISAICP-3918) created as a follow-up.

PR #891, opened 2017-09-21
ISAICP-3667: about page display adjustment

luke-waat:


kalinchernev:
In overall, changes are :+1: 

Noticed the following specifics though:
- long emails will overflow in the contacts area:

solution_about overflowing_texts

- Fields in hidden area are:
about solution in validated state joinup

The ticket requests a different order:

```
In case the Solution has the Solution type field with one of these values: TestComponent, TestService and TestScenario, an additional block should be added as a grey block with the title TRR Information that contains:

1. Test resource type
2. Actor (with a label, this field is a list of values, one or more)
3. Business process (with the label)
4. Product type (with the label)
5. Standardization level (with the label)

```

kalinchernev:
Merged into `UAT-ready` 7ecbad62fd80be65f4b0976857b3d89261336dce

PR #897, opened 2017-09-22
Isaicp 3812: Remove the licence field from solutions

idimopoulos:


tests/features/solution/homepage.feature
```diff
@@ -11,6 +11,9 @@ Feature: Solution homepage
     And the following contact:
       | name  | Placebo             |
       | email | Placebo@example.com |
+    And the following licence:
+      | title       | LGPL                                |
```
brummbar: Why this was added? It seems that is not used anywhere in the test, only in the data.

tests/features/solution/homepage.feature
```diff
@@ -39,7 +42,7 @@ Feature: Solution homepage
     When I go to the homepage of the "Information sharing protocol" solution
     # I should see only the related release.
     # And the distribution directly associated.
-    Then I should see the following tiles in the correct order:
+    And I should see the following tiles in the correct order:
```
brummbar: This should stay a "then".

PR #899, opened 2017-09-22
ISAICP-3890

kalinchernev:
#880

PR #907, opened 2017-09-22
ISAICP-3890

kalinchernev:
Cherry picking commits only for ISAICP-3890 issue.
This is assuming that ISAICP-3834 is already available, because previously #899 and #880 were merging changes about the global restyling of the modal on which these changes here are dependent on.


kalinchernev:
[UAT acceptance](https://webgate.ec.europa.eu/CITnet/jira/browse/ISAICP-3890?focusedCommentId=2332169&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-2332169).

PR #913, opened 2017-09-25
ISAICP-3876

kalinchernev:
#873 without history of `UAT-ready`


web/themes/joinup/prototype/sass/components/_fieldset.sass
```diff
@@ -16,31 +19,35 @@
...
 
 .fieldset__field-label
-  margin-top: 0
   font-weight: $font-weight-medium
+  margin-top: 0
 
 .fieldset__title
   margin-top: 0
 
+.fieldset
```
kalinchernev: This seems to be a repetition of https://github.com/ec-europa/joinup-dev/pull/913/files#diff-a500538f142a2994ab3faa0b48a55e26R7

Also, it seems unnecessary:
screenshot from 2017 09 25 10 48 25

web/themes/joinup/prototype/sass/components/_fieldset.sass
```diff
@@ -55,11 +62,12 @@
 
 .fieldset--border-bottom
   position: relative
-  padding-bottom: 18px
+  margin-bottom: 25px
+  padding-bottom: 27px
```
kalinchernev: Looks really `cheesy` - are you sure you really need these rules having all the other top/bottom spacing rules?

web/themes/joinup/prototype/sass/components/_fieldset.sass
```diff
@@ -71,6 +79,18 @@
       right: -60px
       left: -60px
 
+.fieldset--textarea
+  p
+    overflow: hidden
```
kalinchernev: Seems really risky to do that.

web/themes/joinup/prototype/sass/components/_fieldset.sass
```diff
@@ -71,6 +79,18 @@
...
+  img
+    margin-bottom: 16px
+  img.align-left
```
kalinchernev: Also look very unstable assumption.
Also, it's not BEM at all.

web/themes/joinup/templates/content/node--event--full.html.twig
```diff
@@ -10,15 +10,19 @@
 <article{{ attributes }}>
   <div class="page__content">
     {{ content.body }}
-    {{ content.field_event_spatial_coverage }}
+    {% if content.field_event_spatial_coverage|render %}
+      <div class="fieldset">
```
kalinchernev: Fieldset does not seem to be the right name for a solution which just works with text areas.

web/themes/joinup/templates/field/field--fieldset-wrapper.html.twig
```diff
@@ -0,0 +1,65 @@
+{#
```
kalinchernev: This template is unnecessary having the preprocess function which can manage attributes.

web/themes/joinup/prototype/sass/components/_fieldset.sass
luke-waat: Duplicate removed.

web/themes/joinup/prototype/sass/components/_fieldset.sass
luke-waat: It is only for the grey line fieldsets. It should be bulletproof, any issues with this?

web/themes/joinup/prototype/sass/components/_fieldset.sass
luke-waat: Implementation of the `fieldset--textarea` and `fieldset--border-bottom` have been slightly changed to avoid adding any styles on the paragraph.

web/themes/joinup/prototype/sass/components/_fieldset.sass
luke-waat: These classes comes from WYSIWYG.

kalinchernev: Can we move these rules into either drupal hacks or some other `editor` style sheet?

web/themes/joinup/templates/field/field--fieldset-wrapper.html.twig
kalinchernev: Yes, I had this comment [previously](https://github.com/ec-europa/joinup-dev/pull/873#discussion_r139714564) as well indeed. Though, I have not created a ticket for this, as here I have asked for a specific change, not a refactoring in the whole system.

kalinchernev:
In UAT-ready branch https://github.com/ec-europa/joinup-dev/commit/73665e3321c13802448811d5ddccfd68e521043a

PR #916, opened 2017-09-25
Isaicp 3932: Homepage text

idimopoulos:


web/profiles/joinup/src/Plugin/Block/RecommendedContentBlock.php
```diff
@@ -94,6 +94,14 @@ public function defaultConfiguration() {
    * {@inheritdoc}
    */
   public function build() {
+    $build = [
+      '#attributes' => ['class' => ['listing', 'listing--grid', 'mdl-grid']],
```
kalinchernev: I don't think these classes are necessary, because line 127 already adds these.
Also, it seems these classes are not optimal, paragraph goes out of the tiles' alignment:
attributes_current

You can use the grid "column" classes like this to correct this:
attributes_paragraph

kalinchernev:
@idimopoulos I merged to UAT-ready at https://github.com/ec-europa/joinup-dev/commit/0ad2242926b4eb2c1ba49291d4749e2de773b503 because Marco asked me to make them quickly for the release this week.

I've added the CSS classes in the template as I saw an example like this in core, but please feel free to change this to a better way when you're back.

kalinchernev:
1db87bb99d09fee19ae06a4a2c06fa7ca7f1cd6c a70ed6de708d6c9fc0a8af4c0de4d6fd83fde5fe 07712d591ee879795d0c7eeacd2ea88e0d0a1115 and 8566f94f0ebdda78dceace08a7825851018197c1 merged as a follow-up on initial merge to uat-ready

PR #918, opened 2017-09-27
ISAICP-3931: Show only first row of location field

tom-waat:
DO NOT MERGE before ISAICP-3907


web/themes/joinup/prototype/sass/components/_details.sass
```diff
@@ -71,6 +71,10 @@
     color: $color-grey-medium
     font-weight: $font-weight-medium
 
+.details__element--single
```
kalinchernev: Since we change details "component":
- the avatar's head is cut:
screenshot from 2017-09-27 17-05-27
- inline of adjustment elements:
screenshot from 2017-09-27 17-06-44

tom-waat: "the avatar's head is cut:"
It is cut because of image style. The image there is rounded so image style have square shape.

"inline of adjustment elements:"
Hmm, I am not sure what should be changed there. I am cutting the location text in css, in the same way as for tiles.

kalinchernev:
Added to uat at https://github.com/ec-europa/joinup-dev/commit/609bc5c1f9853bcbdd5ec4eb8bcc1634a1e14bae

PR #930, opened 2017-10-03
3962: Upload distribution files with the same name in different folders

claudiu-cristea:


web/modules/custom/joinup_core/tests/src/Functional/FileUrlFieldTest.php
```diff
@@ -41,36 +44,75 @@ class FileUrlFieldTest extends RdfWebTestBase {
...
 
+    $rdf_entity = Rdf::create([
+      'rid' => 'file_url',
+      'label' => 'Foo',
+    ]);
+    $rdf_entity->save();
+
     // Test file for new entities.
-    $this->drupalGet('rdf_entity/add/file_url');
+    $this->drupalGet($rdf_entity->toUrl('edit-form'));
```
pfrenssen: :+1:

web/modules/custom/joinup_core/tests/src/Functional/FileUrlFieldTest.php
```diff
@@ -41,36 +44,75 @@ class FileUrlFieldTest extends RdfWebTestBase {
...
+    $final_uri = $rdf_entity_file->getFileUri();
+    $this->assertFileExists($final_uri);
 
     // Ensure the file can be downloaded.
-    $this->drupalGet(file_create_url($rdf_entity_file->getFileUri()));
+    $this->drupalGet(file_create_url($final_uri));
     $this->assertSession()->statusCodeEquals(200);
 
+    // Check that the same file is uploaded to different locations.
+    $this->assertNotEquals($initial_uri, $final_uri);
```
pfrenssen: Is this important to be tested? This seems like an implementation detail, because the persisting of the original filenames can also be solved in a different way.

On the other hand, this also verifies that the second file hasn't overwritten the first file, so it is probably useful after all.

web/modules/custom/joinup_core/tests/src/Functional/FileUrlFieldTest.php
```diff
@@ -41,36 +44,75 @@ class FileUrlFieldTest extends RdfWebTestBase {
...
+    $rdf_entity_file = FileUrlHandler::urlToFile($rdf_entity->{$field_name}->target_id);
+    $initial_uri = $rdf_entity_file->getFileUri();
+    $this->assertFileExists($initial_uri);
+
+    // Ensure the file can be downloaded.
+    $this->drupalGet(file_create_url($initial_uri));
+    $this->assertSession()->statusCodeEquals(200);
+
+    // Upload the same file again to test if the file is saved with in new
+    // location while still keeping the same file basename.
```
pfrenssen: This sentence with has one word too many.

web/modules/custom/joinup_core/tests/src/Functional/FileUrlFieldTest.php
```diff
@@ -41,36 +44,75 @@ class FileUrlFieldTest extends RdfWebTestBase {
...
     // Ensure the file can be downloaded.
-    $this->drupalGet(file_create_url($rdf_entity_file->getFileUri()));
+    $this->drupalGet(file_create_url($final_uri));
     $this->assertSession()->statusCodeEquals(200);
 
+    // Check that the same file is uploaded to different locations.
+    $this->assertNotEquals($initial_uri, $final_uri);
+
+    /** @var \Drupal\Core\File\FileSystemInterface $file_system */
+    $file_system = $this->container->get('file_system');
```
pfrenssen: If we move this into the test setup we will certainly be blessed by the container injection gods of the olden days.

web/profiles/joinup/config/install/file.settings.yml
```diff
@@ -0,0 +1,6 @@
+description:
+  type: 'textfield'
+  length: 128
+icon:
+  directory: 'core/modules/file/icons'
+make_unused_managed_files_temporary: true
```
pfrenssen: Yeah let's make sure old stuff gets cleaned up, or we will end up with jiggabytes of cruft.

PR #931, opened 2017-10-04
ISAICP-3963: additional sorting condition on recommended content on the homepage

brummbar:


claudiu-cristea:
Merged in UAT-ready

PR #951, opened 2017-10-16
4017: Create admin UI to move content

claudiu-cristea:


web/modules/custom/custom_page/custom_page.services.yml
```diff
@@ -4,3 +4,7 @@ services:
     arguments: ['@entity.manager']
     tags:
      - { name: event_subscriber }
+  custom_page.og_menu_link_updater:
+    class: Drupal\custom_page\CustomPageOgMenuLinksUpdater
+    arguments: ['@entity_type.manager', '@plugin.manager.menu.link']
+    lazy: true
```
pfrenssen: Services aren't lazy by default? Then we probably need to apply this on many different places.

claudiu-cristea: No, they are instantiated on each request. As a general rule, every service that doesn't use a factory class is eligible to be made lazy. Unless we know that we need it in each request. I was about to open a ticket for that.

web/modules/custom/joinup_core/src/EventSubscriber/GroupManageContentSubscriber.php
```diff
@@ -0,0 +1,22 @@
...
+ * Alters the 'view.group_content_management.manage' route.
+ */
+class GroupManageContentSubscriber extends RouteSubscriberBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function alterRoutes(RouteCollection $collection) {
+    if ($route = $collection->get('view.group_content_management.manage')) {
+      $route->setOption('_admin_route', TRUE);
```
pfrenssen: Here the overview is declared as an admin page but the moderators do not have access to the admin theme. Do you think this would be worthwhile to add? Perhaps in a separate issue?

claudiu-cristea: @pfrenssen Generally I think it's a good idea. Opened https://webgate.ec.europa.eu/CITnet/jira/browse/ISAICP-4045

claudiu-cristea: About making admin path a View path, I just realised that I've opened a core ticket long time ago, in Drupal DevDays Milan https://www.drupal.org/node/2719797. That would be the correct way.

pfrenssen: @claudiu-cristea just mentioned this issue which is relevant: https://www.drupal.org/node/2719797

web/modules/custom/custom_page/custom_page.services.yml
claudiu-cristea: Opened https://webgate.ec.europa.eu/CITnet/jira/browse/ISAICP-4047 to make several services lazy.

pfrenssen: @brummbar explained this to me. By default services themselves are actually lazy loaded (instantiated on demand by the service container), but the dependencies of the service are instantiated directly, even when they may not be needed by all methods on the service. So marking them as lazy and providing the proxy class allows to lazy load _the dependencies of the service_ on demand. This sounds really interesting for many of the services we have in Joinup, especially the helper services that have a lot of dependencies.

web/modules/custom/custom_page/src/CustomPageOgMenuLinksUpdater.php
```diff
@@ -0,0 +1,125 @@
...
+              $instance->deleteLink();
+              // Search for children of deleted menu link.
+              $mids = $this->menuLinkContentStorage->getQuery()
+                ->condition('parent', "menu_link_content:{$instance->getDerivativeId()}")
+                ->execute();
+              if ($mids) {
+                // Remove the relationship to the deleted parent menu link.
+                foreach (MenuLinkContent::loadMultiple($mids) as $menu_link_content) {
+                  $menu_link_content->set('parent', NULL)->save();
+                }
```
pfrenssen: I'm not entirely sure if this is desired on a functional level. What is expected by the end user if a parent item is being moved to a new collection? Do the child items move along? Or should they be orphaned like is happening here?

This situation is not described in the original ticket, should we discuss this with the functional team first?

web/modules/custom/custom_page/tests/src/Kernel/CustomPageOgMenuLinkUpdaterTest.php
```diff
@@ -0,0 +1,239 @@
...
+    $this->assertEmpty($links);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function tearDown() {
+    foreach ([1, 2] as $index) {
+      Rdf::load("http://example.com/$index")->delete();
+    }
```
pfrenssen: The `parent::tearDown()` should also be called here, it does a lot of important clean up work that preserves memory during the test runs.

web/modules/custom/joinup_core/src/Plugin/Action/ChangeGroupAction.php
```diff
@@ -0,0 +1,191 @@
...
+        // Prevent notification dispatching.
+        // @see joinup_notification_dispatch_notification()
+        $node->skip_notification = TRUE;
+        $node->set('og_audience', $group_id)->save();
+
+        $status = ['#markup' => $this->t("The group of @type '%title' has been changed to '%group'.", $args)];
+        if ($is_custom_page) {
+          $status[] = [
+            '#prefix' => ' ',
+            '#markup' => $this->t('The custom page menu link is disabled in the new group and it should be manually enabled.'),
```
pfrenssen: Is this a functional requirement to make the custom pages disabled by default? I don't see this mentioned in the original ticket.

web/modules/custom/custom_page/custom_page.services.yml
claudiu-cristea: That's correct. More details here https://www.drupal.org/node/1973618

claudiu-cristea:
@pfrenssen 

> When a page has been visited before being moved (i.e. it is cached) then the group header of the old group is still showing until the cache is cleared.

This looks like a caching issue, a bug. I think we should investigate this in a separate issue.

PR #953, opened 2017-10-17
ISAICP-3612: Temporarily pin Search API on the previous stable version

pfrenssen:
…to avoid patch conflicts.

This will be unpinned again in ISAICP-3942.


pfrenssen:
This is no longer needed, this is apparently already fixed in develop.

PR #955, opened 2017-10-17
ISAICP-4033: Disable the update module on production

pfrenssen:
This probably got enabled manually during initial deployment.

### DIFF
--- a/web/modules/custom/joinup_migrate/joinup_migrate.post_update.php
+++ b/web/modules/custom/joinup_migrate/joinup_migrate.post_update.php
@@ -1,0 +1,13 @@
+<?php
+
+/**
+ * @file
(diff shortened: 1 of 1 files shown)


web/modules/custom/joinup_migrate/joinup_migrate.install
```diff
@@ -260,3 +260,10 @@ function joinup_migrate_source_db_schema($table) {
 
   return $schema[$table];
 }
+
+/**
+ * Disable the Update module.
+ */
+function joinup_migrate_update_8100() {
```
claudiu-cristea: Is there a reason why this not in a `post_update` function? We are just using the API in a straight way, we're not repairing any structure, storage and also we are not upgrading any API that needs backed repairing.

PR #957, opened 2017-10-18
ISAICP-3942: Update locked dependencies

pfrenssen:


claudiu-cristea:
@pfrenssen, just a question here. Shouldn't we run `composer install` in `scripts/update_production.sh`, on the very top, just before running any other updates?

pfrenssen:
No, the composer dependencies are part of the tarball that is deployed on production.

The general rule is that anything that can be done offline should not be done prior to deploying on production so that we do not spend time waiting on things that can be avoided. Also composer installs can fail due to external resources not being available so it is risky.

PR #959, opened 2017-10-19
4040: Admin tool to visit a RDF page given its ID

claudiu-cristea:


web/modules/custom/joinup_core/joinup_core.routing.yml
```diff
@@ -54,3 +54,13 @@ joinup_core.id_redirect:
...
   requirements:
     _access: 'TRUE'
+
+joinup_core.go_to_rdf_entity_page:
+  path: '/rdf'
+  defaults:
+    _form: 'Drupal\joinup_core\Form\GoToRdfEntityPage'
+    _title: 'Go to RDF entity page'
+  requirements:
+    _permission: 'administer nodes'
```
idimopoulos: This should depend either on 'administer rdf entity' which is defined in the Rdf entity annotation, or a new permission that is specific to this page.

web/modules/custom/joinup_core/joinup_core.routing.yml
```diff
@@ -54,3 +54,13 @@ joinup_core.id_redirect:
     _controller: '\Drupal\joinup_core\Controller\IdRedirect::redirectToRdfEntity'
   requirements:
     _access: 'TRUE'
+
+joinup_core.go_to_rdf_entity_page:
+  path: '/rdf'
```
idimopoulos: Can we have it as a bit more unique url? Rdf is a very common thing to use. I know it is project specific as it is in joinup_core but since this is a distribution, I think it's better to not use this keyword for the url. Maybe rdf-redirect or something similar.

web/modules/custom/joinup_core/joinup_core.routing.yml
claudiu-cristea: The idea was to provide a handy tool, an "easy to type" thing. If we put there a long URL will not help. Right now, on our project, this URL is not used. If it will be in the future, we can change it.

web/modules/custom/joinup_core/joinup_core.routing.yml
claudiu-cristea: The idea was to provide a handy tool, an "easy to type" thing. If we put there a long URL will not help. Right now, on our project, this URL is not used. If it will be in the future, we can change it.

web/modules/custom/joinup_core/joinup_core.routing.yml
claudiu-cristea: Yeah, that was my first thought. But I realised that is not used anywhere and no role is granted with that permission. Ideally we would need a custom permission, something like "administer content" that should cover nodes and RDF entities. I picked up "administer nodes" because that is assigned to "administrator" role.

claudiu-cristea:
After discussion with @idimopoulos I made next changes:

* Added a custom permission: `access uri converter`
* Granted this permission to administrators and moderators.
* Now you can paste a URL (base path or aliased) and get back the decoded ID. The URL can be any  that was built based on a route with the `{rdf_entity}` parameter.

PR #960, opened 2017-10-20
3968: Fix files with the same name that were renamed

claudiu-cristea:


web/modules/custom/joinup_migrate/joinup_migrate.post_update.php
```diff
@@ -159,3 +158,86 @@ function joinup_migrate_post_update_more_redirects() {
...
+    }
+  }
+
+  // Keep only duplicates.
+  array_walk($file_groups, function (array &$file_group) {
+    $file_group = array_filter($file_group, function (array $files) {
+      return count($files) > 1;
+    });
+  });
+  $file_groups = array_filter($file_groups, function (array $file_group) {
```
brummbar: What is this needed for? Is there more to filter? I couldn't understand 😭 
Anyway, if you are trying to remove empty arrays, it's enough to call array_filter() alone.

web/modules/custom/joinup_migrate/joinup_migrate.post_update.php
```diff
@@ -159,3 +158,86 @@ function joinup_migrate_post_update_more_redirects() {
...
+
+  /** @var \Drupal\Core\Database\Query\SelectInterface $query */
+  $query = Database::getConnection()->select('file_managed', 'f')
+    ->fields('f', ['fid', 'uri'])
+    ->fields('d6f', ['filename', 'filepath'])
+    ->orderBy('f.uri')
+    ->condition('f.uri', 'public://distribution/%', 'LIKE');
+  $query->join("$legacy_db_name.files", 'd6f', 'f.fid = d6f.fid');
+  $file_groups = [];
+  foreach ($query->execute()->fetchAll() as $row) {
```
brummbar: Maybe drop a comment to explain what is the result of the query and what is the structure we are building.

web/modules/custom/joinup_migrate/joinup_migrate.post_update.php
```diff
@@ -159,3 +158,86 @@ function joinup_migrate_post_update_more_redirects() {
...
+        if (!$file = file_move($file, "$destination/$file_name")) {
+          throw new \Exception("Can't move file $file_name (fid $fid).");
+        }
+        $redirects[$source_path] = "base:/sites/default/files/" . str_replace('public://', '', $file->getFileUri());
+      }
+    }
+  }
+
+  // Delete existing redirects.
+  if ($rids = $redirect_storage->getQuery()
```
brummbar: So we are creating new redirects for the moved files, so the links that are out already are not turned into dead ends. This is really clever! But is it needed the step of deleting existing redirects? As far as I know there should not be any others to files. Just asking.

web/modules/custom/joinup_migrate/joinup_migrate.post_update.php
claudiu-cristea: Done.

web/modules/custom/joinup_migrate/joinup_migrate.post_update.php
claudiu-cristea: I've created redirects to all migrated files so that we lower the broken links. Such files were migrated, thus the have redirects. Now we want to override those redirects.

web/modules/custom/joinup_migrate/joinup_migrate.post_update.php
claudiu-cristea: That was just for removing empty dirs resulted from the previous step. You're right, `array_filter()` is enough but the function was left there from a previous iteration when I tried to fix both steps in a single statement. However, I removed the line because it's useless.

web/modules/custom/joinup_migrate/joinup_migrate.post_update.php
brummbar: 👍

PR #961, opened 2017-10-24
Check if the solution exists in the SolutionOwnerCheckSubscriber

claudiu-cristea:


pfrenssen:
Hotfix for a fatal error that occurs during the update of v1.16. Looks good.

PR #969, opened 2017-11-02
4057: Copy about text to distributions for CTT (Center Technology Transfer Spain)

pfrenssen:


web/modules/custom/joinup_migrate/joinup_migrate.post_update.php
```diff
@@ -159,3 +161,45 @@ function joinup_migrate_post_update_more_redirects() {
...
+      if (empty($descriptions)) {
+        continue;
+      }
+
+      // Load the distributions that are associated with this solution.
+      $distribution_ids = array_map($get_reference_id, iterator_to_array($affiliate->get('field_is_distribution')));
+      foreach (Rdf::loadMultiple($distribution_ids) as $distribution) {
+        /** @var \Drupal\rdf_entity\RdfInterface $distribution */
+        foreach ($descriptions as $langcode => $description) {
+          $translated_entity = $distribution->getTranslation($langcode);
```
brummbar: I have a minor remark here. In Joinup we have only the english translation, so everything is fine, but if we retrieve a translation that doesn't exist for a certain distributions (in the case the solution has been translated but not the distribution) this will thrown an exception. To be super-safe, we should wrap this with an `if ($distribution->hasTranslation($langcode)) then {}`.
It's optional tho, so feel free to not change it.

web/modules/custom/joinup_migrate/joinup_migrate.post_update.php
```diff
@@ -159,3 +161,45 @@ function joinup_migrate_post_update_more_redirects() {
...
+function joinup_migrate_post_update_copy_about_text_ctt_spain() {
+  // Define a reusable anonymous function to retrieve the entity reference ID
+  // from a field item.
+  $get_reference_id = function (EntityReferenceItem $item) {
+    return $item->getValue()['target_id'];
+  };
+
+  // Load all affiliated content for the CTT Spain collection.
+  $collection = Rdf::load('http://administracionelectronica.gob.es/ctt');
+  $affiliate_ids = array_map($get_reference_id, iterator_to_array($collection->get('field_ar_affiliates')));
```
claudiu-cristea: We don't need this. Entity reference in Drupal 8 is awesome. You can just:
```php
foreach ($collection->get('field_ar_affiliates')->referencedEntities() as ...)
```

claudiu-cristea:
This was QA'd and approved in UAT already. What I did is only resolving the conflict, so I will merge it directly in `develop`.

PR #975, opened 2017-11-09
ISAICP-3980: Add a post update script for ownership transfer. 

idimopoulos:


web/modules/custom/joinup_migrate/joinup_migrate.post_update.php
```diff
@@ -159,3 +161,63 @@ function joinup_migrate_post_update_more_redirects() {
     }
   }
 }
+
+/**
+ * Switches the ownership of the 'Sharing and reuse of IT solutions' collection.
```
claudiu-cristea: Because this file will grow in the next period with a lot of requests to fix conten, I decided in ISAICP-3962 (PR https://github.com/ec-europa/joinup-dev/pull/960/files) to add the ticket number in the docblock. This is true only for *joinup_migrate.post_update.php* file, not other module. The idea is that the migration module will be removed from the codebase and we will clear even the Git history. But not I'm trying to keep track somehow about these updates. Also, the first word, unlike on other functions/methods, should not end with "s". This is true only for `hook_update_N()` and hook_post_update_NAME()` functions. So s/Switches/Switch.

claudiu-cristea: Because this file will grow in the next period with a lot of requests to fix content, I decided in ISAICP-3962 (PR https://github.com/ec-europa/joinup-dev/pull/960/files) to add the ticket number in the docblock. This is true **only for joinup_migrate.post_update.php** file, not for other modules. The idea is that the migration module will be removed from the codebase and we will clear even the Git history. But not I'm trying to keep track somehow about these updates. Also, the first word, unlike on other functions/methods, should not end with "s". This is true only for `hook_update_N()` and `hook_post_update_NAME()` functions. So s/Switches/Switch.

web/modules/custom/joinup_migrate/joinup_migrate.post_update.php
```diff
@@ -159,3 +161,63 @@ function joinup_migrate_post_update_more_redirects() {
...
+ */
+function joinup_migrate_post_update_srits_change_owner() {
+  $collection_id = 'http://data.europa.eu/w21/e5f0febe-f29c-428a-a1a2-6bb0ccd37949';
+  // The 'Joinup moderator'.
+  $current_owner_id = '700003';
+  // The 'Joinup editor'.
+  $new_owner_id = '6363';
+  /** @var \Drupal\og\MembershipManagerInterface $membership_manager */
+  $membership_manager = \Drupal::service('og.membership_manager');
+  /** @var \Drupal\Core\Entity\EntityTypeManager $entity_storage */
```
claudiu-cristea: Seems to be `$entity_type_manager`, not `$entity_storage`, right? And should typehint as interface. So, s/EntityTypeManager/EntityTypeManagerInterface.

web/modules/custom/joinup_migrate/joinup_migrate.post_update.php
```diff
@@ -159,3 +161,63 @@ function joinup_migrate_post_update_more_redirects() {
...
+  $entity_type_manager = \Drupal::service('entity_type.manager');
+
+  $collection = $entity_type_manager->getStorage('rdf_entity')->load($collection_id);
+  if (empty($collection)) {
+    throw new Exception(t('Collection with id :id was not found.', [
```
claudiu-cristea: As general rule: Do not use `t()` in exception messages. The reason is that, when we throw an exception we want to avoid side effects. Calling a function might obscure the original error cause. During exception throw we want to exit as quick as we can with a clear message. Also translation makes no sense because such messages are for developers. Using a double quote string is just fine to embed a variable in such cases: `"Collection with id $collection_id was not found."`. More, root level namespaced classes should always be called with the qualified namespace: `\Exception` (like `\Drupal`).

web/modules/custom/joinup_migrate/joinup_migrate.post_update.php
```diff
@@ -159,3 +161,63 @@ function joinup_migrate_post_update_more_redirects() {
...
+      ':id' => $collection_id,
+    ]));
+  }
+
+  $current_owner = $entity_type_manager->getStorage('user')->load($current_owner_id);
+  if (empty($current_owner)) {
+    throw new Exception(t('User with id :id was not found.', [
```
claudiu-cristea: See above.

web/modules/custom/joinup_migrate/joinup_migrate.post_update.php
```diff
@@ -159,3 +161,63 @@ function joinup_migrate_post_update_more_redirects() {
...
+      ':id' => $current_owner_id,
+    ]));
+  }
+
+  $new_owner = $entity_type_manager->getStorage('user')->load($new_owner_id);
+  if (empty($new_owner)) {
+    throw new Exception(t('User with id :id was not found.', [
```
claudiu-cristea: See above.

web/modules/custom/joinup_migrate/joinup_migrate.post_update.php
```diff
@@ -159,3 +161,63 @@ function joinup_migrate_post_update_more_redirects() {
...
+
+/**
+ * Switches the ownership of the 'Sharing and reuse of IT solutions' collection.
+ */
+function joinup_migrate_post_update_srits_change_owner() {
+  $collection_id = 'http://data.europa.eu/w21/e5f0febe-f29c-428a-a1a2-6bb0ccd37949';
+  // The 'Joinup moderator'.
+  $current_owner_id = '700003';
+  // The 'Joinup editor'.
+  $new_owner_id = '6363';
```
claudiu-cristea: Why are `$current_owner_id` and `new_owner_id`  strings? They are integer according to 'user' entity base fields definition.

web/modules/custom/joinup_migrate/joinup_migrate.post_update.php
```diff
@@ -159,3 +161,63 @@ function joinup_migrate_post_update_more_redirects() {
...
+  $entity_type_manager = \Drupal::service('entity_type.manager');
+
+  $collection = $entity_type_manager->getStorage('rdf_entity')->load($collection_id);
+  if (empty($collection)) {
+    throw new Exception(t('Collection with id :id was not found.', [
+      ':id' => $collection_id,
+    ]));
+  }
+
+  $current_owner = $entity_type_manager->getStorage('user')->load($current_owner_id);
```
claudiu-cristea: Let's save the use storage into a variable, as we are using it 2 times.

web/modules/custom/joinup_migrate/joinup_migrate.post_update.php
```diff
@@ -159,3 +161,63 @@ function joinup_migrate_post_update_more_redirects() {
...
+  if (empty($new_owner)) {
+    throw new Exception(t('User with id :id was not found.', [
+      ':id' => $new_owner_id,
+    ]));
+  }
+
+  // Revoke the administrator role from the 'Joinup moderator'.
+  $membership = $membership_manager->getMembership($collection, $current_owner);
+  if (!empty($membership)) {
+    $membership->revokeRole(OgRole::getRole('rdf_entity', 'collection', OgRoleInterface::ADMINISTRATOR));
```
claudiu-cristea: You can leave its as it is. For the future: we have `$membership->revokeRoleById()` :)

web/modules/custom/joinup_migrate/joinup_migrate.post_update.php
```diff
@@ -159,3 +161,63 @@ function joinup_migrate_post_update_more_redirects() {
...
+
+  // Make user 'Joinup editor' the owner for the 'Sharing and reuse of IT
+  // solutions' collection.
+  $membership = $membership_manager->getMembership($collection, $new_owner);
+  if (empty($membership)) {
+    $membership = $membership_manager->createMembership($collection, $new_owner);
+  }
+
+  $roles = [
+    'rdf_entity-collection-' . OgRoleInterface::AUTHENTICATED,
```
claudiu-cristea: I don't think we should add this role explicitly. I think it works just as site-wide "authenticated user".

web/modules/custom/joinup_migrate/joinup_migrate.post_update.php
idimopoulos: Ow.. I did not know that :/ I think there are some already to be accepted (merged in UAT) that have been without the ticket number. :/

web/modules/custom/joinup_migrate/joinup_migrate.post_update.php
```diff
@@ -159,3 +161,66 @@ function joinup_migrate_post_update_more_redirects() {
...
+  $current_owner_id = 700003;
+  // The 'Joinup editor'.
+  $new_owner_id = 6363;
+  /** @var \Drupal\og\MembershipManagerInterface $membership_manager */
+  $membership_manager = \Drupal::service('og.membership_manager');
+  /** @var \Drupal\Core\Entity\EntityTypeManager $entity_type_manager */
+  $entity_type_manager = \Drupal::service('entity_type.manager');
+
+  /** @var \Drupal\rdf_entity\RdfInterface $collection */
+  $collection = $entity_type_manager->getStorage('rdf_entity')->load($collection_id);
```
claudiu-cristea: Could you add a condition to not run the update if $collection === NULL? Because this breaks the update process on UAT, where the this collection doesn't exist.

PR #977, opened 2017-11-09
ISAICP-4086: subscribing discussions basic functionality

brummbar:


tests/features/joinup_discussion/discussion.subscribe.feature
```diff
@@ -0,0 +1,32 @@
+@api
+Feature: Subscribing to discussions
+  As a member of a collection
+  I want to subscribe to interesting discussions
+  So that I can stay up to date with its evolvements.
```
pfrenssen: @brummbar points out that this is not limited to members of a collection. Any authenticated user can subscribe to a discussion. We need to find a different persona.

web/profiles/joinup/joinup.profile
```diff
@@ -312,6 +312,9 @@ function joinup_install_tasks_alter(&$tasks, $install_state) {
   $tasks['joinup_remove_simplenews_defaults'] = [
     'function' => [JoinupCustomInstallTasks::class, 'removeSimpleNewsDefaults'],
   ];
+  $tasks['joinup_delete_message_subscribe_defaults'] = [
```
claudiu-cristea: Shouldn't this go also in a post-update hook?

brummbar: Removed in https://github.com/ec-europa/joinup-dev/pull/977/commits/a8dc369548265226727894a9300a4c4c133978c1 due to the incompatibility of config_sync with deleted entities.

PR #980, opened 2017-11-12
4132: Upload screenshot artifacts to S3

claudiu-cristea:


tests/src/Context/ScreenshotContext.php
```diff
@@ -17,6 +17,44 @@
...
+   *   An Amazon S3 bucket URI, such as s3://<bucket name>/path/to/artifacts,
+   *   where to store the screenshots. If an empty value is received, no AWS S3
+   *   upload will occur.
+   *
+   * @see tests/behat.yml.dist
+   */
+  public function __construct($screenshots_dir, $artifacts_s3_uri) {
+    $screenshots_dir = trim($screenshots_dir);
+    // If a directory has been passed, ensure the directory exists.
+    if (!empty($this->screenshotsDir) && !is_dir($screenshots_dir)) {
```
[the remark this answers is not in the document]
claudiu-cristea: Should be fixed.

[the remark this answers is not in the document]
idimopoulos: Oops.. Fixed.

build.properties
```diff
@@ -76,6 +76,18 @@ behat.options.strict = false
...
 
+# The location where to save the screenshots. If empty, the screenshot will not
+# be created.
+# Example:
+# behat.screenshots_dir = ${tmp.dir}/behat
+behat.screenshots_dir =
+
+# The AWS S3 URI location where to upload the Behat screenshot artifacts. This
+# property is disabled by default. You'll need to explicit configure it in the
+# continuous integration system.
```
pfrenssen: Huh? What does this mean, we need to explicitly configure it in the CI system?

build.properties
```diff
@@ -76,6 +76,18 @@ behat.options.strict = false
 # shows exception backtraces, 3 shows debugging information.
 behat.options.verbosity = 2
 
+# The location where to save the screenshots. If empty, the screenshot will not
+# be created.
+# Example:
+# behat.screenshots_dir = ${tmp.dir}/behat
+behat.screenshots_dir =
```
pfrenssen: I wouldn't duplicate this line, just leave it commented out. This is the typical pattern in config files, see for example `php.ini`.

build.properties
```diff
@@ -76,6 +76,18 @@ behat.options.strict = false
...
+
+# The AWS S3 URI location where to upload the Behat screenshot artifacts. This
+# property is disabled by default. You'll need to explicit configure it in the
+# continuous integration system.
+# Example:
+# behat.artifacts_s3_uri = s3://joinup2/cphp-behat-artifacts
```
pfrenssen: I wouldn't hardcode it to our S3 bucket, there is no value in that, just provide an example path like `s3://mybucket/mypath`.

tests/src/Context/ScreenshotContext.php
```diff
@@ -17,6 +17,44 @@
...
+   *   The directory where the screenshots are saved. The value is
+   *   passed in behat.yml.
+   * @param string $artifacts_s3_uri
+   *   An Amazon S3 bucket URI, such as s3://<bucket name>/path/to/artifacts,
+   *   where to store the screenshots. If an empty value is received, no AWS S3
+   *   upload will occur.
+   *
+   * @see tests/behat.yml.dist
+   */
+  public function __construct($screenshots_dir, $artifacts_s3_uri) {
```
pfrenssen: These values are optional so they should be marked as such.

tests/src/Context/ScreenshotContext.php
```diff
@@ -17,6 +17,44 @@
...
+  protected $artifactsS3Uri;
+
+  /**
+   * Constructs a new ScreenshotContext context.
+   *
+   * @param string $screenshots_dir
+   *   The directory where the screenshots are saved. The value is
+   *   passed in behat.yml.
+   * @param string $artifacts_s3_uri
+   *   An Amazon S3 bucket URI, such as s3://<bucket name>/path/to/artifacts,
```
pfrenssen: This is not an URI but an URL 😛

tests/src/Context/ScreenshotContext.php
```diff
@@ -26,7 +64,7 @@ class ScreenshotContext extends RawMinkContext {
    * @Then (I )take a screenshot :name
    */
   public function takeScreenshot($name = NULL) {
-    $file_name = sys_get_temp_dir() . DIRECTORY_SEPARATOR . $name;
+    $file_name = $this->screenshotsDir . DIRECTORY_SEPARATOR . $name;
```
pfrenssen: Since `$this->screenshotsDir` is being checked in `::createScreenshot()` to see if it is valid we should not rely on it here.

tests/src/Context/ScreenshotContext.php
```diff
@@ -17,6 +17,44 @@
...
+   *   upload will occur.
+   *
+   * @see tests/behat.yml.dist
+   */
+  public function __construct($screenshots_dir, $artifacts_s3_uri) {
+    $screenshots_dir = trim($screenshots_dir);
+    // If a directory has been passed, ensure the directory exists.
+    if (!empty($screenshots_dir) && !is_dir($screenshots_dir)) {
+      $screenshots_dir = rtrim($screenshots_dir, '/');
+      @mkdir($screenshots_dir, 0777, TRUE);
```
pfrenssen: `mkdir()` might return `FALSE` if the creation of the folder failed but the return value is not checked.

tests/src/Context/ScreenshotContext.php
```diff
@@ -124,12 +165,23 @@ public function createScreenshot($file_name, $message, $ext = TRUE) {
...
+      }
+      catch (DriverException $e) {
       }
-      $html_data = $this->getSession()->getPage()->getContent();
-      file_put_contents($file_name, $html_data);
     }
+
+    if ($this->artifactsS3Uri) {
+      $output = ["aws s3 cp $file_name {$this->artifactsS3Uri}/"];
+      exec($output[0], $output);
```
pfrenssen: What is this proprietary command? Where is this being passed the AWS region, secret and key to use?

Let's use the AWS SDK which is included in our vendor folder like we do in all our other code that interacts with AWS.

PR #995, opened 2017-11-22
Isaicp 4094: Set the 'from' option in contact forms

idimopoulos:


web/modules/custom/contact_form/contact_form.module
```diff
@@ -96,3 +96,17 @@ function contact_form_tokens($type, $tokens, array $data, array $options, Bubble
...
+
+/**
+ * Implements hook_mail_alter().
+ */
+function contact_form_mail_alter(&$message) {
+  if ($message['key'] !== 'contact_form_submission') {
+    return;
+  }
+
+  $from = \Drupal::service('config.factory')->get('contact_form.settings')->get('contact_form_from');
```
pfrenssen: Please document why we are altering the from address.

PR #1001, opened 2017-11-23
4087: Notify discussion subscribers on new comment

claudiu-cristea:


web/modules/custom/joinup_discussion/src/EventSubscriber/SubscribedDiscussionCommentSubscriber.php
```diff
@@ -0,0 +1,195 @@
...
+    return [NotificationEvents::COMMENT_CRUD => 'commentCrudProxy'];
+  }
+
+  /**
+   * Acts as a proxy by passing the control to a dedicated method.
+   *
+   * Normally we should have split the events into more specific events, like
+   * 'comment create', 'comment update', 'comment delete'. Just 'comment crud'
+   * is too generic and is not actually an event. In order to keep the
+   * compatibility with Joinup Notification module, we use this proxy method to
```
pfrenssen: I agree fully with this, maybe we should make a ticket to refactor this in the future so it uses distinct events?

web/modules/custom/joinup_discussion/src/EventSubscriber/SubscribedDiscussionCommentSubscriber.php
```diff
@@ -0,0 +1,195 @@
...
+      /** @var \Drupal\comment\CommentInterface $original_comment */
+      $original_comment = $this->comment->original;
+      // An anonymous comment has been just approved.
+      if (!$original_comment->isPublished() && $this->comment->isPublished()) {
+        // Deliver the notification message.
+        $this->messageDelivery
+          ->createMessage('discussion_comment_new')
+          ->setArguments($this->getArguments())
+          ->setRecipients($this->recipients)
+          ->sendMail();
```
pfrenssen: This call is a duplicate of the one in `::onCommentCreate()`.

web/modules/custom/joinup_discussion/src/EventSubscriber/SubscribedDiscussionCommentSubscriber.php
```diff
@@ -0,0 +1,195 @@
...
+  protected function onCommentCreate(): void {
+    // Do not send the notification if the comment is not yet published. This
+    // normally happens when an anonymous user is posting a comment because such
+    // comments are subject of approval.
+    if ($this->comment->isPublished()) {
+      // Deliver the notification message.
+      $this->messageDelivery
+        ->createMessage('discussion_comment_new')
+        ->setArguments($this->getArguments())
+        ->setRecipients($this->recipients)
```
pfrenssen: If we also call a method here like `::getRecipients()` instead of relying on the fact that the list of recipient was already built in a previous step this class would become more resilient to bugs if it is expanded or refactored in the future.

web/modules/custom/joinup_notification/joinup_notification.services.yml
```diff
@@ -1,36 +1,43 @@
...
     class: Drupal\joinup_notification\EventSubscriber\CommunityContentSubscriber
-    arguments: ['@entity_type.manager', '@config.factory', '@current_user', '@og.group_type_manager', '@og.membership_manager', '@joinup_core.workflow.helper', '@joinup_core.relations_manager', '@message_notify.sender', '@state_machine_revisions.revision_manager']
+    parent: joinup_notification.notification_subscriber_base
+    arguments: ['@state_machine_revisions.revision_manager']
     tags:
       - { name: event_subscriber }
   joinup_notification.collection.rdf.subscriber:
     class: Drupal\joinup_notification\EventSubscriber\CollectionRdfSubscriber
-    arguments: ['@entity_type.manager', '@config.factory', '@current_user', '@og.group_type_manager', '@og.membership_manager', '@joinup_core.workflow.helper', '@joinup_core.relations_manager', '@message_notify.sender']
+    parent: joinup_notification.notification_subscriber_base
```
pfrenssen: Nifty trick!

web/modules/custom/joinup_notification/src/JoinupMessageDelivery.php
```diff
@@ -0,0 +1,145 @@
...
+   */
+  public function setRecipientsAsEmails(array $mails): JoinupMessageDeliveryInterface {
+    $this->mails = $mails;
+    return $this;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function sendMail(): JoinupMessageDeliveryInterface {
```
pfrenssen: Since this is part of a service this shouldn't rely on data which is stored in properties on the object. The properties might be set during a previous usage of the service and may be outdated.

I made a followup to fix this: ISAICP-4169.

web/modules/custom/joinup_notification/src/JoinupMessageDelivery.php
```diff
@@ -0,0 +1,145 @@
...
+    }, $this->accounts));
+
+    // Merge E-mail addresses extracted from passed from user accounts with
+    // those passed directly as recipient E-mail addresses. Ensure uniqueness.
+    $mails = array_unique(array_merge($mails, $this->mails));
+
+    // Send E-mail messages.
+    array_walk($mails, function (string $mail): void {
+      $options = ['save on success' => FALSE, 'mail' => $mail];
+      $this->messageNotifier->send($this->message, $options);
```
pfrenssen: Sending of a message might fail. This should return `FALSE` if a message could not be sent so that the calling code can log this or display a message to the user.

web/modules/custom/joinup_notification/joinup_notification.behat.inc
```diff
@@ -173,6 +173,48 @@ class JoinupNotificationSubContext extends DrupalSubContextBase implements Drupa
     }
   }
 
+  /**
+   * Asserts that an email has been sent.
```
pfrenssen: Has not been sent ;)

claudiu-cristea: dammit 👍

PR #1008, opened 2017-11-28
Isaicp 4108

pfrenssen:


pfrenssen:
This diverged dramatically from the main branch (3700+ lines!), so I rebased it on top of the latest and greatest code in the discussions epic (ISAICP-4091) and force pushed. We'll need to rework `SubscribedDiscussionSubscriber`, it has duplicated code, and we can no longer rely on setting properties on the object now that it is used for two events.

PR #1010, opened 2017-11-28
Hotfix to support migration database when restoring from the producti…

idimopoulos:
We try to restore the production databases in the acceptance environment so the execute-migration target does not run. Thus, we need to manually run the setup-migration and import-legacy-db targets.


idimopoulos:
I merged myself only because @claudiu-cristea sent me the same solution, so we kinda accepted each other's solution by default.

PR #1019, opened 2017-12-05
4091: Notify subscribers when a discussion gets updated

claudiu-cristea:


tests/features/joinup_discussion/discussion.subscribe.feature
```diff
@@ -100,3 +102,48 @@ Feature: Subscribing to discussions
...
+    And I fill in "Motivation" with "Reporting this content..."
+    And I press "Report"
+    Then the following email should not have been sent:
+      | recipient | dale@example.com                                                                  |
+      | subject   | Joinup: The discussion "Rare Butter" was updated in the space of "Dairy products" |
+      | body      | The discussion "Rare Butter" was updated in the "Dairy products" collection.      |
+    And the following email should not have been sent:
+      | recipient | flash@example.com                                                                  |
+      | subject   | Joinup: The discussion "Rare Butter" was updated in the space of "Dairy products" |
+      | body      | The discussion "Rare Butter" was updated in the "Dairy products" collection.      |
```
pfrenssen: Let's put another line here "Then 0 e-mails should have been sent" - that makes the test more bullet proof if we change the email template in the future.

claudiu-cristea: We cannot. Because other notifications are, beside the scope of this ticket, are sent.

PR #1021, opened 2017-12-07
ISAICP-4189: Allow updating modules with new service dependencies.

pfrenssen:


claudiu-cristea:
Just read the upstream PR. Stunning! I'm trying to follow there.

pfrenssen:
OK this is green but only by accident. We have a dependency that is causing a strange failure on the bot:

```
[error]  The "filter_image_style" plugin does not exist.
```

I cannot replicate this error locally. I spent some time bisecting the dependencies but it is slow work because we need to wait for results from the bot. I managed to work around it now by manipulating the `composer.json` and `composer.lock` files to only apply the updates for affected parts and their direct dependencies.

So this is working right now, but as soon as it is merged in develop it will break again on the next update. I'm opening a separate issue for this and we'll discuss during standup how to proceed.

PR #1022, opened 2017-12-08
4198: FilterImageStyle missing plugin bug

claudiu-cristea:


claudiu-cristea:
Before QAing, see the last comment from the issue ticket.

PR #1026, opened 2017-12-13
ISAICP-4199: Update dependencies

pfrenssen:


claudiu-cristea:
If we are updating dependencies, I wonder if we could update also some patches. For example, https://www.drupal.org/node/2914478 has a new version that helps reduce some noise with Drush 9.

pfrenssen:
So it turns out that pinning commits on development branches is an unsupported feature of Composer and is known to be problematic. This is not going to be easy to fix. I made an issue about it to have a look at it at a later moment since we really need this, we don't have control over when our dependencies will make new releases: https://webgate.ec.europa.eu/CITnet/jira/browse/ISAICP-4228

It's a good suggestion to update the patches. I will do this.

pfrenssen:
I have started to go through the patches but I had to give up on it, it takes too much time. The problem is that the patches need individual attention, they might not apply since they are rolled for Drupal 8.5.x, or the newer patches might even contain regressions compared to the ones we are already using (e.g. https://www.drupal.org/project/drupal/issues/2845361). So all patches actually need to be reviewed before updating, that's too much work for this ticket.

PR #1030, opened 2017-12-20
4229: Contact form attachments

claudiu-cristea:


build.project.xml
```diff
@@ -62,8 +62,29 @@
         <setmysqlautoincrement />
         <phingcall target="writable-settings-folder" />
         <phingcall target="writable-settings-php" />
+
+        <!-- Avoid duplicates in settings.php, if the file exists. -->
+        <reflexive>
```
pfrenssen: Let's not hack this into the installer. This should be a separate Phing target, like all the other instances where we make a change to the settings file (e.g. `enable-config-read-only`, `disable-persistent-sessions-cookies` etc).

Also the location of the private files folder is environment specific so this should be in `settings.local.php` instead.

build.properties
```diff
@@ -64,6 +64,8 @@ tmp.dir = ${project.basedir}/tmp
 # Exported configuration.
 drupal.config.dir = ${project.basedir}/config
 
+# Drupal private files directory (relative to ${project.basedir}).
+drupal.private_files.dir = private
```
pfrenssen: Let's use the full relative path here, so that we don't need to construct this every time it is used.

But again, we don't need to deal with this on a build level. If people want to use a different path they can manually change their settings file, we don't need to impose anything here.

web/modules/custom/joinup_core/joinup_core.install
```diff
@@ -60,3 +60,41 @@ function joinup_core_requirements($phase) {
...
+  // The private files directory should have been already created.
+  if (!$private_files_directory = realpath(DRUPAL_ROOT . '/../private')) {
+    throw new \RuntimeException("Directory '$private_files_directory' has to be created prior running this script.");
+  }
+
+  // As the private files directory is outside the web tree, we really don't
+  // need a .htaccess file but we still provide it to comply with the Drupal
+  // security policy in the unlikely case this directory will be moved under the
+  // web tree in the future.
+  file_save_htaccess($private_files_directory);
```
pfrenssen: I wouldn't mess with this stuff in an update hook. This is not the responsibility of the application, but of the system administrators.

web/modules/custom/joinup_core/joinup_core.install
```diff
@@ -60,3 +60,41 @@ function joinup_core_requirements($phase) {
...
+  // Save current settings in a local variable.
+  $settings = Settings::getAll();
+
+  // Add the 'file_private_path' in settings.php.
+  $private_files_directory_settings['settings']['file_private_path'] = (object) [
+    'value' => $private_files_directory,
+    'required' => TRUE,
+  ];
+  chmod($settings_file, 0666);
+  drupal_rewrite_settings($private_files_directory_settings, $settings_file);
```
pfrenssen: If a website is correctly configured then this should be impossible. The site should not be allowed to overwrite its own settings file, since this would make it easier to do privilege escalation attacks.

pfrenssen:
Having a new look now that I know that D8 does require the private files folder to be created manually outside the web root, I think we can probably handle this entirely in a Phing target, then we don't need the `joinup_core_update_8101()` update hook.

web/modules/custom/joinup_core/joinup_core.post_update.php
```diff
@@ -25,3 +25,35 @@ function joinup_core_post_update_install_vbo() {
...
+
+  $message_storage = \Drupal::entityTypeManager()->getStorage('message');
+  $ids = $message_storage->getQuery()
+    ->condition('template', 'contact_form_submission')
+    ->exists('field_contact_attachment')
+    ->execute();
+
+  foreach ($message_storage->loadMultiple($ids) as $message) {
+    /** @var \Drupal\file\FileInterface $attachment */
+    if ($attachment = $message->field_contact_attachment->entity) {
```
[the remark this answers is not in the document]
claudiu-cristea: > If I am not mistaking, the new Drupal Coding standards are not allowing assignment in the if statement. I might be wrong though.

Could you point me to the docs? At least coder is not complaining and now Coder is following the official Coding Standards. That would be DX regression :(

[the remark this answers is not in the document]
claudiu-cristea: > Also, we should also check if the file actually exists.

This looks as a good precaution. Just that we always have the file system service in the `$file_system` variable. Also, I wonder if we need to compute the real path, `file_exists()` should be able to deal also with relative paths, no?

build.dist.xml
```diff
@@ -35,6 +35,16 @@
         </if>
     </target>
 
+    <!-- Make settings.local.php writable. -->
+    <target name="writable-settings-local-php">
+        <if>
+            <available file="${website.settings.local.php}" type="file" property="website.settings.local.php.available" />
+            <then>
+                <chmod mode="0644" failonerror="false" verbose="true" quiet="false" file="${website.settings.local.php}" />
```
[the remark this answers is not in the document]
claudiu-cristea: I only duplicate the code from `writable-settings-php` :) But I guess, we don't want to fail here. We still want to try to write to the file in the target that uses this as a dependency. Have no string idea (I guess the caller is working also without this dependency)

web/modules/custom/joinup_core/joinup_core.post_update.php
```diff
@@ -25,3 +25,42 @@ function joinup_core_post_update_install_vbo() {
...
+      if (empty($attachment)) {
+        continue;
+      }
+      $realpath = \Drupal::service('file_system')->realpath($attachment->getFileUri());
+      if (empty($realpath) || !is_file($realpath)) {
```
claudiu-cristea: We've already referred the file system service in the `$file_system` variable. There's no need to call the container again and for each file. Also, from [the PHP official site](http://php.net/manual/en/function.realpath.php):

> realpath() returns FALSE on failure, e.g. if the file does not exist.

I guess we can improve this entire check as:
```php
if (!$attachment || ($file_system->realpath($attachment->getFileUri()) === FALSE)) {
```

PR #1034, opened 2017-12-21
HOTFIX #ISAICP-4247: Don't use redis for semaphores.

sandervd:


pfrenssen:
Thanks, let's find a conclusive fix in ISAICP-4245.

PR #1038, opened 2018-01-03
QA 4108 against 4091 DO NOT MERGE

claudiu-cristea:


web/modules/custom/joinup_discussion/src/EventSubscriber/SubscribedDiscussionSubscriber.php
```diff
@@ -76,56 +99,159 @@ public static function getSubscribedEvents(): array {
...
+  /**
+   * Returns the fully loaded User entity for the current user.
+   *
+   * @return \Drupal\user\UserInterface
+   *   The current user.
+   */
+  protected function getCurrentUser(): UserInterface {
+    try {
+      /** @var \Drupal\user\UserInterface $user */
+      $user = $this->entityTypeManager->getStorage('user')->load($this->currentUser->id());
```
claudiu-cristea: I wonder why we don't use the shortcut and avoid entity type manager injection:

```
use Drupal\user\Entity\User;
...
$user = User::load($this->currentUser->id());
```

web/modules/custom/joinup_discussion/src/EventSubscriber/SubscribedDiscussionSubscriber.php
```diff
@@ -76,56 +99,159 @@ public static function getSubscribedEvents(): array {
...
+      return $user;
+    }
+    catch (InvalidPluginDefinitionException $e) {
+      // The storage for the 'user' entity type should exist. Ignore this.
```
claudiu-cristea: True, but it would be nice to return an explicit `NULL`. At least the IDE would not complain :)

web/modules/custom/joinup_discussion/src/EventSubscriber/SubscribedDiscussionSubscriber.php
```diff
@@ -76,56 +99,159 @@ public static function getSubscribedEvents(): array {
...
+   * Returns the group the discussion belongs to.
+   *
+   * @param \Drupal\node\NodeInterface $discussion
+   *   The discussion for which to return the group.
+   *
+   * @return \Drupal\rdf_entity\RdfInterface|null
+   *   The group (which is an RDF entity of type 'collection' or 'solution'),
+   *   or NULL if no group is found.
+   */
+  protected function getDiscussionGroup(NodeInterface $discussion) : ?RdfInterface {
```
claudiu-cristea: No space before ":".

web/modules/custom/joinup_discussion/src/EventSubscriber/SubscribedDiscussionSubscriber.php
```diff
@@ -76,56 +99,159 @@ public static function getSubscribedEvents(): array {
...
-  protected function getArguments(): array {
-    return [
-      '@entity:title' => $this->discussion->label(),
-      '@group:label' => $this->group->label(),
-      '@group:bundle' => $this->group->bundle(),
-      '@entity:url' => $this->discussion->toUrl('canonical', [
-        'absolute' => TRUE,
-      ])->toString(),
-    ];
+  protected function getArguments(NodeInterface $discussion) : array {
```
claudiu-cristea: No space before ":".

web/modules/custom/joinup_discussion/src/EventSubscriber/SubscribedDiscussionSubscriber.php
```diff
@@ -76,56 +99,159 @@ public static function getSubscribedEvents(): array {
...
+  protected function sendMessage(NodeInterface $discussion, string $message_template): bool {
+    try {
+      return $this->messageDelivery
+        ->createMessage($message_template)
+        ->setArguments($this->getArguments($discussion))
+        ->setRecipients($this->getSubscribers($discussion))
+        ->sendMail();
+    }
+    catch (\Exception $e) {
+      $this->logger->critical('Unexpected exception thrown when sending a message for a discussion.',
```
claudiu-cristea: Why using the `'php'` channel here? Wouldn't be more appropriate to use the `'mail'` channel as `MailManager` is doing. And I like that they are injecting the logger factory instead a specific logger channel.

tests/src/Traits/EntityTrait.php
```diff
@@ -49,7 +55,37 @@ protected function getEntityByLabel($entity_type, $label, $bundle = NULL) {
...
+  protected static function entityTypeAliases() {
+    return [
+      'content' => 'node',
+    ];
+  }
+
+  /**
+   * Translates human readable entity types to machine names.
+   *
+   * @param string $entity_type
```
claudiu-cristea: Let's rename to `$entity_type_label` to make it clear.

web/modules/custom/joinup_discussion/src/EventSubscriber/SubscribedDiscussionSubscriber.php
claudiu-cristea: https://github.com/drupal/core/blob/8.5.x/lib/Drupal/Core/Mail/MailManager.php#L300

tests/src/Traits/EntityTrait.php
pfrenssen: Renamed to `$entity_type_id` and also updated the trait to use strict typing.

web/modules/custom/joinup_discussion/joinup_discussion.module
```diff
@@ -73,3 +74,16 @@ function joinup_discussion_node_view(array &$build, EntityInterface $entity, Ent
     ],
   ];
 }
+
+/**
+ * Implements hook_ENTITY_TYPE_predelete().
+ */
+function joinup_discussion_node_predelete(EntityInterface $entity) : void {
```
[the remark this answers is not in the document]
pfrenssen: Created followup issue: ISAICP-4256

web/modules/custom/joinup_discussion/joinup_discussion.module
```diff
@@ -73,3 +74,16 @@ function joinup_discussion_node_view(array &$build, EntityInterface $entity, Ent
...
+  if ($entity->bundle() !== 'discussion') {
+    return;
+  }
+
+  // Send a message to subscribed users to inform that the discussion was
+  // deleted.
+  joinup_notification_dispatch_notification('delete', DiscussionEvents::DELETE, $entity);
```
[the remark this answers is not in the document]
pfrenssen: I fully agree with this, I also think `joinup_notification_dispatch_notification()` is too closely linked to workflow notifications, I cannot for the life of me remember why I went with this approach.

web/modules/custom/joinup_discussion/src/EventSubscriber/SubscribedDiscussionSubscriber.php
```diff
@@ -76,56 +99,159 @@ public static function getSubscribedEvents(): array {
...
-      ->setArguments($this->getArguments())
-      ->setRecipients($this->getRecipients())
-      ->sendMail();
+  /**
+   * Sends notification to subscribed users when a discussion is deleted.
+   *
+   * @param \Drupal\joinup_notification\Event\NotificationEvent $event
+   *   The notification event object.
+   */
+  public function notifyOnDiscussionDeletion(NotificationEvent $event) : void {
```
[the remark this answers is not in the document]
pfrenssen: 👍

web/modules/custom/joinup_discussion/src/EventSubscriber/SubscribedDiscussionSubscriber.php
pfrenssen: I can answer this! I was in fact the one that added these `Entity::load()` and `Entity::loadMultiple()` static methods to core, back in 2014 (ref [Issue #2190313: Add $entityType::load() and ::loadMultiple() to simplify loading entities](https://www.drupal.org/node/2190313)).

These methods are were originally provided as convenient alternatives for the D7 functions `entity_load()` and `entity_load_multiple()` to help people in porting procedural D7 code to D8. They are not intended to be used in modern OO code.

In fact this will also retrieve the entity storage from the container, but since it is a static method it gets it from `\Drupal::entityManager()` in the global namespace. This pattern should not be used in OO code since it makes it a real pain to mock in unit tests.

web/modules/custom/joinup_discussion/src/EventSubscriber/SubscribedDiscussionSubscriber.php
pfrenssen: Implementing this I found that this method shouldn't return NULL because then the problem just cascades up the call stack and the IDE starts complaining in other methods. We positively need to be certain that this method returns a user. I found a creative solution for it that satisfies PHPStorm :D

web/modules/custom/joinup_discussion/src/EventSubscriber/SubscribedDiscussionSubscriber.php
pfrenssen: Very good point, I wasn't aware of the existence of this channel. I copied this approach from somewhere. I'll also have a look at injecting the factory, that sounds like a clean solution.

web/modules/custom/joinup_discussion/src/Event/DiscussionDeleteEvent.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\joinup_discussion\Event;
+
+/**
+ * An event to fire whenever a discussion is deleted.
+ */
+class DiscussionDeleteEvent extends DiscussionEventBase {
```
claudiu-cristea: Love it ❤️

web/modules/custom/joinup_discussion/src/EventSubscriber/SubscribedDiscussionSubscriber.php
```diff
@@ -57,75 +69,191 @@ class SubscribedDiscussionSubscriber implements EventSubscriberInterface {
...
+
+  /**
+   * Sends the notification to the recipients.
+   *
+   * @param \Drupal\node\NodeInterface $discussion
+   *   The discussion for which to send the notification.
+   * @param string $message_template
+   *   The ID of the message template to use.
+   *
+   * @return bool
```
claudiu-cristea: In fact the caller never uses this return value. I'm not sure we need to return something here. Also, even Drupal send successfully an email, there's no guarantee the the message has been delivered by the system mailing and we cannot do anything in case of failure. I would make this function `:void` and just catch and log a potential exception (?)

web/modules/custom/joinup_subscription/joinup_subscription.behat.inc
```diff
@@ -0,0 +1,60 @@
...
+    }
+  }
+
+  /**
+   * Returns the Joinup subscription service.
+   *
+   * @return \Drupal\joinup_subscription\JoinupSubscriptionInterface
+   *   The subscription service.
+   */
+  protected function getSubscriptionService() : JoinupSubscriptionInterface {
```
claudiu-cristea: No space before `:`.

pfrenssen: I seem to keep writing it like this -.- we need a code sniffer rule for this

claudiu-cristea: I remember, you've opened an issue in d.o, in coder?

PR #1039, opened 2018-01-04
Isaicp 4231: Update VBO

idimopoulos:


composer.json
```diff
@@ -56,7 +56,7 @@
         "drupal/token": "~1.0",
         "drupal/video_embed_field": "~1.5",
         "drupal/view_unpublished": "~1.0@alpha",
-        "drupal/views_bulk_operations": "dev-1.x#6106b1c7a172d273e0407c49a4c4a8186c929573",
+        "drupal/views_bulk_operations": "dev-1.x",
```
claudiu-cristea: Don't like to rely on dev-1.x, that is changing very often. Till this will be release we could grab changes that would potentially break again the site. I see `8.x-2.0-beta1` has been released in December 28. Could you test, please, if that version includes the changes that we need?

PR #1046, opened 2018-01-08
ISAICP-4222: Refactor the invitation form to align with the WF

brummbar:


web/modules/custom/joinup_invite/src/Form/InviteFormBase.php
```diff
@@ -46,145 +49,237 @@ public static function create(ContainerInterface $container) {
...
+   * @return string
    *   A string version of user's full name.
    */
-  protected function getAccountName(UserInterface $user) : TranslatableMarkup {
-    return $this->t('@name (@email)', [
-      '@name' => $user->get('full_name')->value,
-      '@email' => $user->getEmail(),
-    ]);
+  protected function getAccountName(UserInterface $user): string {
+    return $user->get('full_name')->value;
```
idimopoulos: I do not really like this drawback. I see that now when you start typing you get the full name and the user name in the parenthesis. I don't think most cases will be done using user names.
I would suggest that we do show the email even instead of the user name if we cannot show both. 
If there will be two people named "John Smith" with usernames "john_smith" and "smith_john" it can be very weird. And that's the case where the user names match and it is not starlord and monkey respectively. 
An e-mail would be absolute and it would also cover the case where the users communicated externally through the mailing system and the facilitator would have to distinguish the users.

web/modules/custom/joinup_invite/src/Form/InviteToGroupForm.php
```diff
@@ -51,46 +59,71 @@ public static function create(ContainerInterface $container) {
...
-    $users = array_filter($form_state->getValue('users'));
-    $group = $form_state->get('group');
-    $role_id = $group->getEntityTypeId() . '-' . $group->bundle() . '-facilitator';
-    $facilitator_role = $this->entityTypeManager->getStorage('og_role')->load($role_id);
-
-    foreach ($users as $uid) {
-      $user = $this->entityTypeManager->getStorage('user')->load($uid);
-      $membership = $this->ogMembershipManager->getMembership($group, $user);
+    $users = $this->getUserList($form_state);
+    $role_id = implode('-', [
```
idimopoulos: Is it generally the best option to use an implode to concatenate 3 pieces of known strings in a static array?
Just asking, why not `$role_id = $this->rdf_entity->getEntityTypeId() . '-' . $this->rdf_entity->bundle() . '-' . $form_state->getValue('role');` ?

web/themes/joinup/prototype/sass/components/_chip.sass
```diff
@@ -0,0 +1,54 @@
...
+  display: inline-block
+  vertical-align: middle
+  overflow: hidden
+  text-align: center
+
+  &:focus,
+  &:active
+    background-color: #4e4e4e !important
+
+  .icon
```
idimopoulos: The icons seem to be a bit tilted on the left. If we add here a
```
position: relative;
left: 1px;
```
the issue is solved.

web/modules/custom/joinup_core/src/Element/Chip.php
```diff
@@ -0,0 +1,79 @@
+<?php
+
+namespace Drupal\joinup_core\Element;
```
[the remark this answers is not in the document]
brummbar: This is a form element type, so it cannot go to the theme. If you would change the theme, the backend code would not work anymore. The fact that is a component found to the MDL design doesn't mean that it cannot be used on a bootstrap theme for example.

web/modules/custom/joinup_core/templates/chip.html.twig
```diff
@@ -0,0 +1,18 @@
+{#
```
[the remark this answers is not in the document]
brummbar: Same as above. Cannot define something used in the backend on the theme level.

web/modules/custom/joinup_core/templates/input--chip-button.html.twig
```diff
@@ -0,0 +1,15 @@
+{#
```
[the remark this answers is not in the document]
brummbar: Same as above

web/modules/custom/joinup_invite/src/Form/InviteFormBase.php
```diff
@@ -46,145 +49,237 @@ public static function create(ContainerInterface $container) {
...
-    $form['filter_container']['email'] = [
+    $form['#id'] = Html::getId($this->getFormId());
+    $form['#attributes']['class'][] = 'invite-form';
+
+    $form['autocomplete'] = [
       '#type' => 'textfield',
-      '#title' => $this->t('Email or name'),
-      '#default_value' => $form_state->getValue('email') ?: '',
+      '#title' => $this->t('E-mail'),
+      '#description' => $this->t('Enter a Joinup user e-mail address or start typing to search for a user and select it with the mouse or the keyboard.'),
```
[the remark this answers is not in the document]
brummbar: The label was discussed with the project manager and can be updated later by the functional team. Now it's fine like this.

web/modules/custom/joinup_invite/src/Form/InviteFormBase.php
brummbar: Please read description in the JIRA ticket for why the PM decided like this.

web/modules/custom/joinup_invite/src/Form/InviteToGroupForm.php
brummbar: I find it more clear to read.

web/themes/joinup/prototype/sass/components/_chip.sass
brummbar: As explained in the JIRA ticket the icons are temporary and no time should be spent on it. A related front-end ticket will take care of the styling before release.

web/modules/custom/joinup_core/templates/chip.html.twig
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * @file
+ * Default theme implementation to display a chip component from Material Design library..
```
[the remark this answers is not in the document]
brummbar: To be removed indeed.

pfrenssen:
This was depending on code from a ticket in the discussion epic (#1056). This has now been accepted, so I rebased this to be merged into the epic branch itself (#1079).

PR #1047, opened 2018-01-08
ISAICP-4260: Drop cache tables immediately after importing a database dump

pfrenssen:


src/Phing/DropCacheTablesTask.php
```diff
@@ -0,0 +1,57 @@
...
+declare(strict_types = 1);
+
+namespace DrupalProject\Phing;
+
+require_once 'phing/tasks/ext/pdo/PDOTask.php';
+
+/**
+ * Phing task to drop Drupal's cache tables from the database.
+ */
+class DropCacheTablesTask extends \PDOTask {
```
claudiu-cristea: Drop OR Truncate. I see that it's truncate. Let's use everywhere the correct term: truncate.

PR #1050, opened 2018-01-15
ISAICP-4207: ADMS-AP ETL pipline

sandervd:


sandervd:
This PR was created against develop instead of ISAICP-4206

PR #1058, opened 2018-01-22
ISAICP-4207: ADMS-AP ETL pipeline

sandervd:


web/modules/custom/rdf_etl/src/ProcessStepBase.php
```diff
@@ -0,0 +1,61 @@
...
+abstract class ProcessStepBase extends PluginBase implements EtlProcessStepInterface {
+  use StringTranslationTrait;
+
+  /**
+   * {@inheritdoc}
+   */
+  public function __construct(array $configuration, $plugin_id, $plugin_definition) {
+    parent::__construct($configuration, $plugin_id, $plugin_definition);
+
+    $this->configuration += $this->defaultConfiguration();
```
claudiu-cristea: This should be replaced by `$this->setConfiguration($configuration);` and the merge of default configuration should be mode there:

```php
public function setConfiguration(array $configuration): void {
  $this->configuration = $configuration + $this->defaultConfiguration();
  // Or even use NestedArray::mergeDeep()....
}
```

web/modules/custom/joinup_communities/spain_ctt/src/Plugin/EtlDataPipeline/SpainCttDataPipeline.php
```diff
@@ -0,0 +1,26 @@
...
+ *  label = @Translation("Spain - Center for Technology Transfer"),
+ * )
+ */
+class SpainCttDataPipeline extends EtlDataPipelineBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function initStepDefinition(): void {
+    $this->steps = new PipelineStepDefinitionList();
```
claudiu-cristea: Can be moved in the base class, as is needed for every pipeline, no?

web/modules/custom/rdf_etl/src/Plugin/EtlDataPipelineBase.php
```diff
@@ -0,0 +1,60 @@
...
+   * {@inheritdoc}
+   */
+  public function setActiveStepDefinition(int $sequence): void {
+    $this->steps->seek($sequence);
+  }
+
+  /**
+   * Initializes steps to a PipelineStepDefinitionList.
+   */
+  abstract protected function initStepDefinition(): void;
```
claudiu-cristea: What is the reason this in not in the `EtlDataPipelineInterface` interface?

web/modules/custom/rdf_etl/src/Plugin/EtlDataPipelineBase.php
```diff
@@ -0,0 +1,60 @@
...
+    $this->initStepDefinition();
+  }
+
+  /**
+   * Get steps.
+   *
+   * @return \Drupal\rdf_etl\PipelineStepDefinitionList
+   *   The step definition.
+   */
+  public function stepDefinitionList(): PipelineStepDefinitionList {
```
claudiu-cristea: This should be added to the interface as is a public API method,

web/modules/custom/rdf_etl/src/EtlState.php
```diff
@@ -0,0 +1,55 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\rdf_etl;
+
+/**
+ * Class representing a state of a pipeline.
+ *
+ * @package Drupal\rdf_etl
```
claudiu-cristea: I see the `@package` tag used everywhere in this PR but Drupal Coding Standards, the _API documentation and comment standards_ doesn't support this tag https://www.drupal.org/docs/develop/coding-standards/api-documentation-and-comment-standards

web/modules/custom/rdf_etl/src/EtlState.php
```diff
@@ -0,0 +1,55 @@
...
+    $this->sequence = $sequence;
+  }
+
+  /**
+   * Get the persisted position within the pipeline.
+   *
+   * @return int
+   *   The current position of within the pipeline.
+   */
+  public function sequence(): int {
```
claudiu-cristea: Is there a reason why sequences are not string identifiers? Probably because we use the iterator? Otherwise is nice to have strings for steps, it would be more meaningful.

web/modules/custom/rdf_etl/src/Plugin/EtlProcessStep/ManualUploadStep.php
```diff
@@ -0,0 +1,64 @@
...
+   */
+  public function validateConfigurationForm(array &$form, FormStateInterface $form_state): void {
+    parent::validateConfigurationForm($form, $form_state);
+    $form_state->setError($form['adms_file'], $this->t('Invalid file'));
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function submitConfigurationForm(array &$form, FormStateInterface $form_state): void {
```
claudiu-cristea: I think we should move this in the abstract and add a new interface method like `extractDataFromFormSubmit()`. That we implement in each implementation so that we avoid to duplicate code around build info.

web/modules/custom/rdf_etl/src/EtlState.php
```diff
@@ -0,0 +1,55 @@
...
+ */
+class EtlState {
+
+  /**
+   * The pipeline plugin id.
+   *
+   * @var string
+   */
+  protected $pipeline;
+  protected $sequence;
```
claudiu-cristea: Needs docs.

claudiu-cristea:
Problems related to the general architecture of `rdf_etl`:

- I went 'out of bounds' with the pipeline and I wasn't able to use anymore the pipeline because the sequence is stored in the state. So, even a cache rebuild won’t help. It keeps sayin':
  ```Exception: Cannot get out of bound element from pipeline definition list. in Drupal\rdf_etl\PipelineStepDefinitionList->current() (line 47)```
- The data persisted across steps is stored in state vars. That means 2 users will overlap one on each other. We discussed 2 possible solutions here:
  1. Use `user.private_tempstore` instead of state vars and make graph per-user `http://etl-sink/{user}`. Implement a garbage-collection policy that will clear unfinished pipelines. We can even popup a status message (warning) that alerts the user "You have an unfinished RDF file import process. Click _here_ to finish the process or _reset/clear the import_".
  1. Use the Lock API to allow only one import operation at a time. TBD what happens if the work is abandoned and the lock is ON, how to release the lock?
-  ~What if the user takes a lunch break after a step and the whole process waits for user input in a step X with UI? Supposing that he even close and restart the computer. When he goes back to `/admin/content/etl` shouldn't he be redirected directly to the step X?~ EDIT: It seems that this works as expected.

web/modules/custom/rdf_etl/src/Plugin/EtlProcessStepManager.php
```diff
@@ -0,0 +1,48 @@
...
+
+  /**
+   * {@inheritdoc}
+   *
+   * @throws \Exception
+   *   Plugin does not adhere to interface.
+   */
+  public function createInstance($plugin_id, array $configuration = []): EtlProcessStepInterface {
+    $data_pipeline = parent::createInstance($plugin_id, $configuration);
+    if (!$data_pipeline instanceof EtlProcessStepInterface) {
```
claudiu-cristea: Really? I wonder how this could be possible. I think the plugin factory already knows to do this as we've already passed the interface in the constructor. See https://github.com/drupal/core/blob/8.4.x/lib/Drupal/Component/Plugin/Factory/DefaultFactory.php#L100. I think we can drop the whole method.

web/modules/custom/rdf_etl/src/Plugin/EtlDataPipelineManager.php
```diff
@@ -0,0 +1,48 @@
...
+
+  /**
+   * {@inheritdoc}
+   *
+   * @throws \Exception
+   *   Plugin does not adhere to interface.
+   */
+  public function createInstance($plugin_id, array $configuration = []): EtlDataPipelineInterface {
+    $data_pipeline = parent::createInstance($plugin_id, $configuration);
+    if (!$data_pipeline instanceof EtlDataPipelineInterface) {
```
claudiu-cristea: Same as for `\Drupal\rdf_etl\Plugin\EtlProcessStepManager::createInstance()`

PR #1060, opened 2018-01-23
Isaicp 4202: Automatic subscription to group's community content

idimopoulos:


web/modules/custom/joinup_core/joinup_core.module
```diff
@@ -274,6 +278,19 @@ function joinup_core_entity_base_field_info(EntityTypeInterface $entity_type) {
       ->setComputed(TRUE);
   }
 
+  // Add the 'subscription_bundles' property to the 'og_membership' entity.
+  if (($entity_type->id() === 'og_membership')) {
+    $fields['subscription_bundles'] = BaseFieldDefinition::create('map')
```
brummbar: As written in the ticket, this will create a serialised field with no easy way of filtering by query.

web/profiles/joinup/joinup.behat.inc
```diff
@@ -2236,4 +2236,57 @@ JS;
...
+    // This can also be accessed with:
+    // $og_membership->get('subscription_bundles')->getValue()[0][$type];
+    $og_bundles = $og_membership->get('subscription_bundles')->getValue();
+    $og_bundles = array_map(function ($row) {
+      return $row['entity_type'] . ':' . $row['bundle'];
+    }, $og_bundles);
+
+    sort($bundles);
+    sort($og_bundles);
+    PHPUnit_Framework_Assert::assertEquals($bundles, $og_bundles);
```
brummbar: Add the `\` in front of `PHPUnit_Framework_Assert`.

build.project.xml
```diff
@@ -105,6 +105,13 @@
 
     <!-- Execute Drupal updates. -->
     <target name="execute-updates">
+        <drush
+                command="entity:updates"
+                assume="yes"
+                root="${website.drupal.dir}"
+                bin="${drush.bin}"
+                verbose="${drush.verbose}">
+        </drush>
```
pfrenssen: This should be removed. The entity updates are only intended to be run during development and are too dangerous to run in production environments. They should be replaced by regular updates or post-updates.

web/modules/custom/joinup_core/src/Plugin/Field/FieldType/EntityBundleItem.php
```diff
@@ -0,0 +1,76 @@
...
+ * This is a pseudo reference field that references bundles by storing the
+ * entity type and the bundle machine name.
+ *
+ * @FieldType(
+ *   id = "entity_bundle",
+ *   label = @Translation("Entity bundle"),
+ *   description = @Translation("A simple field referencing to bundles")
+ * )
+ */
+class EntityBundleItem extends FieldItemBase {
```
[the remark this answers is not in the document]
pfrenssen: `EntityDuad` 😁

build.project.xml
idimopoulos: Aye aye captain.

web/modules/custom/joinup_core/src/Plugin/Field/FieldType/EntityBundleItem.php
```diff
@@ -0,0 +1,76 @@
+<?php
+
+namespace Drupal\joinup_subscription\Plugin\Field\FieldType;
```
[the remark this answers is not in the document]
idimopoulos: Holy molly.. That is a real miss..

pfrenssen:
So I commented this because the full entity updates were being performed:

> This should be removed. The entity updates are only intended to be run during development and are too dangerous to run in production environments. They should be replaced by regular updates or post-updates.

Instead of fixing it the full entity-updates are now done inside an update hook instead of with a drush command. Unfortunately just moving the code around doesn't fix it, and now we cannot deploy the update on production because the update hook fails.

PR #1064, opened 2018-01-25
4269: Test ADMSv1 to v2 updates

claudiu-cristea:


web/modules/custom/rdf_etl/src/Plugin/EtlAdms2ConvertPassInterface.php
```diff
@@ -0,0 +1,68 @@
...
+   */
+  public function convert(array $data): void;
+
+  /**
+   * Performs testing assertions.
+   *
+   * @param \Drupal\KernelTests\KernelTestBase $test
+   *   The testing class.
+   */
+  public function performAssertions(KernelTestBase $test): void;
```
idimopoulos: Can you explain why this is here? This takes a KernelTestBase class and below there is a getTestngRdfData while these are defined as conversion plugins. How come this is not part of a test or so?

web/modules/custom/rdf_etl/src/Plugin/EtlAdms2ConvertPassPluginBase.php
```diff
@@ -0,0 +1,270 @@
...
+   * @param string|null $objects
+   *   (optional) A serialized list of objects (delimited by space). If passed,
+   *   the results will be limited to this list objects.
+   *
+   * @return array[]
+   *   The query results as array.
+   */
+  protected function getTriplesFromGraph(string $graph_uri, ?string $subject = NULL, ?string $predicate = NULL, ?string $objects = NULL): array {
+    $filter = [];
+    $filter[] = $subject ? "VALUES ?subject { <$subject> } ." : '';
```
idimopoulos: The trailing dot here can be moved to the implode function as 
`implode(" .\n", $filter);`

web/modules/custom/rdf_etl/src/Plugin/EtlAdms2ConvertPassPluginBase.php
```diff
@@ -0,0 +1,270 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getTestingRdfData(): ?string {
+    return NULL;
+  }
+
+  /**
+   * Gets all the triples from a given graph, ordered by subject.
```
idimopoulos: Keep in mind that there is a limit to the fetched triples which I think defaults to a thousand or ten thousands. 
You should perform checks with more than this amount of triples to ensure consistency (or at least that the query indeed fetches all results).

This number is set in the `virtuoso.ini`'s parameter `ResultSetMaxRows`. Maybe a test with that limit set to a small number can also avoid memory consuming tests.

web/modules/custom/rdf_etl/tests/src/Kernel/ConvertToAdms2Test.php
```diff
@@ -0,0 +1,141 @@
...
+      }
+    }
+
+    if ($rdf_data = $this->prepareRdfData($rdf_data)) {
+      $graph_store = new GraphStore($connection_uri);
+      $graph = new Graph();
+      $graph->parse($rdf_data);
+      $out = $graph_store->replace($graph, $graph_uri);
+      if (!$out->isSuccessful()) {
+        throw new \Exception("Cannot import RDF data from plugin '$plugin_id' in graph '$graph_uri'.");
```
idimopoulos: This `$plugin_id` variable seems to be outside the `foreach` loop so it will take the last value in the array regardless of which plugin's data actually fail, no?

web/modules/custom/rdf_etl/src/Plugin/EtlAdms2ConvertPassInterface.php
claudiu-cristea: See my explanation in the issue ticket.

web/modules/custom/rdf_etl/src/Plugin/EtlAdms2ConvertPassPluginBase.php
claudiu-cristea: Done.

web/modules/custom/rdf_etl/src/Plugin/EtlAdms2ConvertPassPluginBase.php
claudiu-cristea: I prefered to make this query safe by splitting the results in chunks with `LIMIT` and `OFFSET` clauses. I went with a chunk of 10000 items which seems for me reasonable. The default value  of `ResultSetMaxRows` is one million and I cannot imagine a system being configured with less than 10000 triples.

PR #1069, opened 2018-01-29
ISAICP-4208 - Data pipeline step: Upload

sandervd:


claudiu-cristea:
Problems related to the general architecture of `rdf_etl` (such QA should have been raised in #1058):

- I went 'out of bounds' with the pipeline and I wasn't able to use anymore the pipeline because the sequence is stored in the state. So, even a cache rebuild won’t help. It keeps sayin':
  ```Exception: Cannot get out of bound element from pipeline definition list. in Drupal\rdf_etl\PipelineStepDefinitionList->current() (line 47)```
- The data persisted across steps is stored in state vars. That means 2 users will overlap one on each other. We discussed 2 possible solutions here:
  1. Use `user.private_tempstore` instead of state vars and make graph per-user `http://etl-sink/{user}`. Implement a garbage-collection policy that will crear unfinished pipelines. We can even popup a status message (warning) that alerts the user "You have an unfinished RDF file import process. Click _here_ to finish the process or _reset/clear the import_".
  1. Use the Lock API to allow only one import operation at a time. TBD what happens if the work is abandoned and the lock is ON, how to release the lock?
-  What if the user takes a lunch break after a step and the whole process waits for user input in a step X with UI? Supposing that he even close and restart the computer. When he goes back to `/admin/content/etl` shouldn't he be redirected directly to the step X?

web/modules/custom/rdf_etl/src/Plugin/EtlProcessStep/ManualUploadStep.php
```diff
@@ -16,13 +22,54 @@
...
+    parent::__construct($configuration, $plugin_id, $plugin_definition);
+  }
 
   /**
    * {@inheritdoc}
    */
   public function execute(array $data): void {
-    // TODO: Implement execute() method.
+    /** @var \EasyRdf\Http\Response $response */
+    $response = $this->graphStore->replace($data['graph'], self::SINK_GRAPH);
```
claudiu-cristea: As we've discussed on chat, we need to store the sink graph value in the plugin default configuration so that we're able to swap it in tests. That would mean the contents of `\Drupal\rdf_etl\ProcessStepBase::defaultConfiguration` will be:
```php
public function defaultConfiguration(): array {
  return [
    'sink_graph' => static::SINK_GRAPH,
  ];
}
```
`const SINK_GRAPH = 'http://etl-sink/';` declaration should be moved in the plugin interface and this line will become:
```php
$response = $this->graphStore->replace($data['graph'], $this->getConfiguration()['sink_graph']);
```

In this way we're able to create a process step plugin instance that uses a different graph:
```php
$config = ['sink_graph' => 'http://example.com/testing_sink_graph'];
$step = \Drupal::service('plugin.manager.etl_process_step')
  ->createInstance('convert_to_adms2', $config);
```

PR #1074, opened 2018-01-30
4212: ADMS validation step

claudiu-cristea:


composer.json
```diff
@@ -43,7 +43,7 @@
         "drupal/piwik": "~1.0",
         "drupal/piwik_reporting_api": "^1.0-beta2",
         "drupal/r4032login": "dev-1.x#391f10d69b9b8c466a2fddac2a6b99f1289e7781",
-        "drupal/rdf_entity": "dev-8.x-1.x",
+        "drupal/rdf_entity": "dev-graph-store-trait",
```
idimopoulos: I do not see a PR on the upstream repo or any ticket related to that branch. How are we going to keep that here?

web/modules/custom/adms_validator/src/AdmsValidationResult.php
```diff
@@ -0,0 +1,145 @@
...
+  /**
+   * Fix rules 100, 101, 102, 103.
+   *
+   * @param string $graph_uri
+   *   The graph URI.
+   * @param \Drupal\rdf_entity\Database\Driver\sparql\Connection $sparql
+   *   The SPARQL endpoint.
+   *
+   * @todo Remove this hack in ISAICP-4296.
+   * @see https://webgate.ec.europa.eu/CITnet/jira/browse/ISAICP-4296
```
idimopoulos: I see the upstream issue included in this ticket is closed. What is the outcome of this?

web/modules/custom/rdf_etl/tests/src/Kernel/AdmsValidationStepTest.php
```diff
@@ -0,0 +1,106 @@
...
+   * @dataProvider providerTestAdmsValidationStepPlugin
+   */
+  public function testAdmsValidationStepPlugin(string $rdf_file, bool $valid): void {
+    /** @var \Drupal\rdf_etl\Plugin\EtlProcessStepManager $manager */
+    $manager = \Drupal::service('plugin.manager.etl_process_step');
+    $data = ['sink_graph' => static::TEST_GRAPH];
+    $plugin = $manager->createInstance('adms_validation', $data);
+
+    $graph = new Graph();
+    $graph->parseFile(__DIR__ . "/../../fixtures/$rdf_file");
```
idimopoulos: Let's let drupal handle the path. We can use
```
$filepath = drupal_get_path('module', 'rdf_etl') . "/tests/fixtures/{$rdf_file}";
$graph->parseFile($filepath);
```
instead.

web/modules/custom/rdf_etl/tests/src/Kernel/AdmsValidationStepTest.php
```diff
@@ -0,0 +1,106 @@
...
+    $this->createGraphStore()->replace($graph, static::TEST_GRAPH);
+
+    // Execute the validation step.
+    $plugin->execute($data);
+
+    if ($valid) {
```
idimopoulos: Please, change this to $expected_valid;

composer.json
claudiu-cristea: Yepp, I forgot to add the PR. Here it is: https://github.com/ec-europa/rdf_entity/pull/22

web/modules/custom/adms_validator/src/AdmsValidationResult.php
claudiu-cristea: The upstream issue was opened based on a false assumption, so I closed that issue. Now, we moved this problem in ISAICP-4296 and this issue is no more blocked on ISAICP-4296 (because of this hack). So the QA of ISAICP-4296 and its upstream issue is not anymore in the scope of this QA.

web/modules/custom/rdf_etl/tests/src/Kernel/AdmsValidationStepTest.php
claudiu-cristea: When you are referring a file within the same extension (module, theme, profile) you should use the path relative to `__DIR__` instead of `drupal_get_path()`. Learn here why https://www.drupal.org/project/drupal/issues/2351919

PR #1077, opened 2018-02-01
ISAICP-4294: collapse custom page attachments

brummbar:


web/themes/joinup/templates/field/field--fieldset--show-more.html.twig
```diff
@@ -0,0 +1,44 @@
+{#
+/**
+ * @file
+ * Override.
```
idimopoulos: A restricted description from someone that pays quite a bit of attention. Update this please :)

brummbar: Ouch :)

PR #1081, opened 2018-02-07
Isaicp 4287: Change the rdf type uri for collections, solutions, releases and distributions

idimopoulos:


web/modules/custom/joinup_core/joinup_core.install
```diff
@@ -60,3 +62,94 @@ function joinup_core_requirements($phase) {
 function joinup_core_update_8100(&$sandbox) {
   \Drupal::service('module_installer')->install(['config_sync']);
 }
+
+/**
+ * Update the type of collections, solutions, releases and distributions.
+ *
+ * Two things will happen here.
```
claudiu-cristea: This block of text should be removed from the docblock. The reason is that the docblock description is outputted in in the updates list (see the screenshot) and it might confuse the user/admin. Instead you can move it as inline comment inside the function. Also, I think the 2 "things" deserve to be numbered or bulleted.
1__php

web/modules/custom/joinup_core/joinup_core.install
```diff
@@ -60,3 +62,94 @@ function joinup_core_requirements($phase) {
...
+ * The first is that all related field mappings will be updated.
+ * The second is that all entities in the database will be updated.
+ */
+function joinup_core_update_8102() {
+  $entity_type_id = 'rdf_entity';
+  $query = Database::getConnection('default', 'sparql_default');
```
claudiu-cristea: Let's use a correct naming for this var. Query doesn't seem to describe its purpose. `$connection`, `$sparql_connection` or just `$sparql`?

web/modules/custom/joinup_core/joinup_core.install
```diff
@@ -60,3 +62,94 @@ function joinup_core_requirements($phase) {
...
+      'new_uri' => 'http://www.w3.org/ns/dcat#Distribution',
+    ],
+  ];
+
+  $bundle_predicate = 'http://www.w3.org/1999/02/22-rdf-syntax-ns#type';
+  foreach ($bundles as $bundle => $properties) {
+    $rdf_entity_mapping = RdfEntityMapping::loadByName($entity_type_id, $bundle);
+
+    // Update the mappings.
+    $rdf_entity_mapping->setRdfType($properties['new_uri']);
```
claudiu-cristea: Nit: Just for fanciness, this could be written more compact with chaining `$rdf_entity_mapping->setRdfType($properties['new_uri'])->save();`.

web/modules/custom/joinup_core/joinup_core.install
```diff
@@ -60,3 +62,92 @@ function joinup_core_requirements($phase) {
...
+    $rdf_entity_mapping->save();
+
+    // Update the entities.
+    $graphs = $rdf_entity_mapping->getGraphs();
+    foreach ($graphs as $graph_uri) {
+      // Keep counters in order to verify the results.
```
claudiu-cristea: Well, are we not trusting SPARQL queries or we are adding tests inside the updating code? This is weird. Are we doing the same when we're running an UPDATE SQL statement against a MySQL database? No, we are not doing with MySQL, so why doing this with SPARQL? Normally this would be covered by an _update path test_ (see `\Drupal\FunctionalTests\Update\UpdatePathTestBase`) but as we discussed, we are not providing such tests. So, we test every time manually the results of an update. Looks like we're mistrusting the below `WITH ... DELETE {...} INSERT {...} WHERE {...}` query. Could you explain this count check?

web/modules/custom/joinup_core/joinup_core.install
idimopoulos: Opps.. I thought only the first line was presented there.

web/modules/custom/joinup_core/joinup_core.install
idimopoulos: No we normally do not test the update sql queries. My only concern here was that we were mangling with the database directly so I thought that we should have a simple check. But that's fine. Without the test, the code simplifies a lot.

PR #1082, opened 2018-02-07
ISAICP-4305: Install field manually on hook_update

idimopoulos:


pfrenssen:
Looking good!

I added a line to clear the field type cache before adding the new field since I was getting a `PluginNotFoundException` when trying to run the update with a backup of the production database. We are using a brand new field type that doesn't exist in the cache yet.

PR #1084, opened 2018-02-08
4307: Don't try to move a missed file

claudiu-cristea:


pfrenssen:
Looking good, thanks!

PR #1085, opened 2018-02-08
ISAICP-4308: Fix rounding error causing update to loop endlessly.

pfrenssen:


pfrenssen:
I verified the splitting that @claudiu-cristea did, it is the right approach and it looks good. I also tried it out locally with the production database.

PR #1097, opened 2018-02-22
ISAICP-4330: Do not allow the update path to break if a membership re…

idimopoulos:
…fers to a deleted group.

### DIFF
--- a/web/modules/custom/joinup_subscription/joinup_subscription.post_update.php
+++ b/web/modules/custom/joinup_subscription/joinup_subscription.post_update.php
@@ -29,9 +29,10 @@ function joinup_subscription_post_update_subscribe_to_community_content(&$sandbo
     ->execute();
 
   foreach (OgMembership::loadMultiple($mids) as $membership) {
-    // Avoid breaking the update path for leftover memberships.
+    // Organic Groups can be configured to delete memberships asynchronously in
+    // a cron job or a manually started batch process. This means orphaned
+    // memberships can be present in the database. Ignore them.
     if (empty($membership->getGroup())) {
-      trigger_error(sprintf("Group with id %s is missing.", $membership->getGroupId()), E_USER_WARNING);
       continue;
     }
     $membership->set('subscription_bundles', [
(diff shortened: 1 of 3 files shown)


web/modules/custom/joinup_subscription/joinup_subscription.post_update.php
```diff
@@ -29,6 +29,11 @@ function joinup_subscription_post_update_subscribe_to_community_content(&$sandbo
     ->execute();
 
   foreach (OgMembership::loadMultiple($mids) as $membership) {
+    // Avoid breaking the update path for leftover memberships.
+    if (empty($membership->getGroup())) {
+      trigger_error(sprintf("Group with id %s is missing.", $membership->getGroupId()), E_USER_WARNING);
+      continue;
+    }
```
pfrenssen: I don't understand this, isn't it allowed to delete groups in the site? I think this is by design, we are not cleaning up reverse references on purpose. Why are we throwing warnings for this?

Also, since this fixes the update path, shouldn't this be against master?

web/modules/custom/joinup_subscription/joinup_subscription.post_update.php
```diff
@@ -29,6 +29,11 @@ function joinup_subscription_post_update_subscribe_to_community_content(&$sandbo
     ->execute();
 
   foreach (OgMembership::loadMultiple($mids) as $membership) {
+    // Avoid breaking the update path for leftover memberships.
```
pfrenssen: This comment makes it seem that this is a workaround, but it is not. Let's instead explain why we really do this, maybe something like this:

```
    // Organic Groups can be configured to delete memberships asynchronously in 
    // a cron job or a manually started batch process. This means orphaned
    // memberships can be present in the database. Ignore them.
```

web/modules/custom/joinup_subscription/joinup_subscription.post_update.php
idimopoulos: Maybe it is better to create a PR against master.
As for the warning, it is not to declare something as miss behaving. It is merely to capture the exceptions in the output during the upgrade so that we can keep track of it.

We are already doing something similar to ensure the update path provides us enough output, e.g. the files missing and the ctt release description text copy.

pfrenssen:
Whoops changing the base pulled in hundreds of commits, that's not what I expected.

I will address the two small remarks, then rebase on master manually.

pfrenssen:
OK fixed remarks and rebased.

@idimopoulos do you want to have a final look?

idimopoulos:
Changes are fine :)

PR #1098, opened 2018-02-22
HOTFIX-ISAICP-4328: Re-enable subpathauto, views_bulk_operations, email_registration

claudiu-cristea:


web/modules/custom/joinup_core/joinup_core.install
```diff
@@ -60,3 +60,18 @@ function joinup_core_requirements($phase) {
 function joinup_core_update_8100(&$sandbox) {
   \Drupal::service('module_installer')->install(['config_sync']);
 }
+
+/**
+ * Re-enable the subpathauto, views_bulk_operations and email_registration.
+ */
+function joinup_core_update_8101() {
+  $core_extensions = \Drupal::configFactory()->getEditable('core.extension');
```
brummbar: I would add a line to clear the config cache, to be sure to pull the stored configuration. We are discussing that stale caches might have caused our current configuration.

claudiu-cristea: Yes, let's do that. Even, IMO, `getEditable()` should not read cached entries because you cannot edit and save an object retrieved from stale cache.

pfrenssen:
This has been converted into a script to execute with `drush php:script` so this is no longer needed.

PR #1101, opened 2018-02-27
ISAICP-4323 against master: Purge rdf backend is not working as expected

idimopoulos:


idimopoulos:
Closing in favor of #1094

PR #1102, opened 2018-02-27
4336: Drupal ^8.5.0

claudiu-cristea:


composer.json
```diff
@@ -47,7 +47,7 @@
         "drupal/rdf_entity": "dev-8.x-1.x",
         "drupal/redirect": "^1.0",
         "drupal/redis": "^1.0@RC",
-        "drupal/search_api": "~1.5",
+        "drupal/search_api": "1.6.0",
```
pfrenssen: This doesn't need to be locked any more.

claudiu-cristea: We're not in QA yes. Don't hurry ;)

pfrenssen: Woops sorry, I thought now that it is green it is ready for QA :) go ahead

composer.json
```diff
@@ -221,7 +219,8 @@
                 "Allow also YouTube embed URLs. Improve the YT regexp @see https://www.drupal.org/node/2899093": "https://www.drupal.org/files/issues/2899093-8.patch"
             },
             "drupal/views_data_export": {
-                "Display schema should extend rest export @see https://www.drupal.org/project/views_data_export/issues/2896104": "https://www.drupal.org/files/issues/2896104-2-fix_display_schema.patch"
+                "Display schema should extend rest export @see https://www.drupal.org/project/views_data_export/issues/2896104": "https://www.drupal.org/files/issues/2896104-2-fix_display_schema.patch",
+                "Download link returns 406 with Drupal ^8.5.0 @see https://www.drupal.org/project/views_data_export/issues/2951185": "https://www.drupal.org/files/issues/2951185-2.patch"
```
idimopoulos: This issue is a temporary workaround that adds the ?_format=csv query parameter to the link by default.
If the link is altered manually though, a critical error will still occur but by the module. So not an exception, but the page shows the text "Critical error" in the normal theme.

web/.ht.router.php
```diff
@@ -0,0 +1,65 @@
+<?php
```
idimopoulos: I think we should add this to the .gitignore. We don't need this file in our project.

composer.json
idimopoulos: That if you go to a page that the csv link is listed somewhere in the page, it is generated with the `?_format=csv` parameter. But you can always try to enter the url manually in the address bar without the parameter if you know it.

claudiu-cristea: Then you "don't know it" :) Because, now, the URL should contain the query param too.

web/.ht.router.php
claudiu-cristea: According to https://github.com/drupal-composer/drupal-scaffold this is a "scaffold file" and

> It is assumed that the scaffold files will be committed to the repository, to ensure that the correct files are used on the CI server (see Limitation, above). After running composer install for the first time commit the scaffold files to your repository.

Not needing that file it's not a good reason. We have also other (scaffold) files, like `web.config`, that we are not using but we are still keep them to ensure the integrity. I would not start to check file by file and see that we are using or not. I prefer that we have a complete Drupal codebase.

idimopoulos: Indeed you are right on that! Didn't think it that way! I thought they were there by mistake :P (not the web.config since it is a server file but the rest)

PR #1103, opened 2018-02-27
ISAICP-4338: Extra && operator in update_production.sh script

brummbar:


pfrenssen:
Looking good! This `&&` was left behind when we removed the line that enables config_readonly recently.

PR #1112, opened 2018-03-05
ISAICP-4348: Exclude rule 15 from the ADMS validation

idimopoulos:


web/modules/custom/adms_validator/src/AdmsValidationResult.php
```diff
@@ -28,10 +28,11 @@ class AdmsValidationResult {
    * @see https://webgate.ec.europa.eu/CITnet/jira/browse/ISAICP-4296
    */
   public function __construct(Result $result, $graph_uri, Connection $sparql) {
+    $skipped_rules = [15, 100, 101, 102, 103];
     foreach ($result as $error) {
       // @todo Remove this hack in ISAICP-4296.
       // @see https://webgate.ec.europa.eu/CITnet/jira/browse/ISAICP-4296
-      if (!in_array($error->Rule_ID->getValue(), range(100, 103))) {
+      if (!in_array($error->Rule_ID->getValue(), $skipped_rules)) {
```
claudiu-cristea: Well, I don't it's enough to skip the rule, we should also implement our own validation. See `$this->applyRules100To103()`

PR #1116, opened 2018-03-08
Isaicp 4353: Bulk update expired items.

idimopoulos:


web/modules/custom/joinup_core/src/EventSubscriber/RefreshCachedFieldsEventSubscriber.php
```diff
@@ -86,89 +87,98 @@ public function __construct(EntityTypeManagerInterface $entityTypeManager, TimeI
...
+    $piwik_config = $this->configFactory->get('piwik.settings');
+    $piwik_reporting_api_config = $this->configFactory->get('piwik_reporting_api.settings');
+    $site_id = $piwik_config->get('site_id');
+    $token_auth = $piwik_reporting_api_config->get('token_auth');
+
+    $sub_params = [
+      'module' => 'API',
+      'idSite' => $site_id,
+      'token_auth' => $token_auth,
+    ];
```
pfrenssen: All this stuff should not be needed, it is handled automatically by the Piwik Reporting API module.

web/modules/custom/joinup_core/joinup_core.post_update.php
```diff
@@ -74,3 +74,14 @@ function joinup_core_post_update_move_contact_form_attachments() {
 function joinup_core_post_update_install_smart_trim() {
   \Drupal::service('module_installer')->install(['smart_trim']);
 }
+
+/**
+ * Sets the PIWIK batch size to 50 items and item limit to 1000.
+ */
+function joinup_core_post_update_piwik_bulk_size() {
+  // This will allow to mainly always timeout, otherwise keep processing.
+  \Drupal::configFactory()->getEditable('cached_computed_field.settings')
```
pfrenssen: It is not needed to update config using a post-update hook, just update the exported config. It will be automatically synced on deployment.

web/modules/custom/joinup_core/src/EventSubscriber/RefreshCachedFieldsEventSubscriber.php
```diff
@@ -86,89 +87,98 @@ public function __construct(EntityTypeManagerInterface $entityTypeManager, TimeI
...
     }
-    // Only refresh the field if it has actually expired. It might have been
-    // updated already since it has been added to the processing queue.
-    if (!$this->fieldNeedsRefresh($event)) {
+
+    try {
+      $response = $query->execute()->getResponse();
+    }
+    catch (\Exception $e) {
+      $this->loggerFactory->get('joinup_core')->error($e->getMessage());
```
pfrenssen: Great, log the error and move on, best approach in this case. :+1:

web/modules/custom/joinup_core/src/EventSubscriber/RefreshCachedFieldsEventSubscriber.php
```diff
@@ -86,89 +87,98 @@ public function __construct(EntityTypeManagerInterface $entityTypeManager, TimeI
...
+      if (!$entity = $this->getExpiredEntity($item)) {
+        continue;
+      }
+
+      $query_params = $this->getSubqueryParams($entity, $sub_params);
+      $query->setParameter('urls[' . $index . ']', http_build_query($query_params));
     }
-    // Only refresh the field if it has actually expired. It might have been
-    // updated already since it has been added to the processing queue.
-    if (!$this->fieldNeedsRefresh($event)) {
```
pfrenssen: This line is lost, let's keep it so we don't do unnecessary requests.

web/modules/custom/joinup_core/src/EventSubscriber/RefreshCachedFieldsEventSubscriber.php
```diff
@@ -86,89 +87,98 @@ public function __construct(EntityTypeManagerInterface $entityTypeManager, TimeI
...
-   * Returns the action count the given entity received in the given period.
+   * Fetches the expired entity.
+   *
+   * @param \Drupal\cached_computed_field\ExpiredItemInterface $item
+   *   The expired item object.
+   *
+   * @return \Drupal\Core\Entity\EntityInterface|null
+   *   The expired entity object or NULL if none is found.
+   */
+  protected function getExpiredEntity(ExpiredItemInterface $item) {
```
pfrenssen: This duplicates `::getEntity()`.

web/modules/custom/joinup_core/src/EventSubscriber/RefreshCachedFieldsEventSubscriber.php
```diff
@@ -355,4 +335,37 @@ protected function getPiwikSettings($bundle) {
...
+   *
+   * @param \Drupal\Core\Entity\EntityInterface $entity
+   *   The entity with the expired field.
+   * @param array $parameters
+   *   A list of extra parameters to pass to the array of parameters.
+   *
+   * @return array
+   *   A list of parameters.
+   */
+  protected function getSubqueryParams(EntityInterface $entity, array $parameters = []): array {
```
pfrenssen: Hmm what is this? Do you have any documentation for it? It looks like this is a feature from Piwik, meaning that it should better belong in the library or contrib module instead of in custom code. At least then this code is reusable.

web/profiles/joinup/config/install/cached_computed_field.settings.yml
```diff
@@ -0,0 +1,3 @@
+batch_size: 100
+item_limit: 200
```
pfrenssen: Let's remove the `item_limit`.

web/modules/custom/joinup_core/src/EventSubscriber/RefreshCachedFieldsEventSubscriber.php
```diff
@@ -86,89 +87,98 @@ public function __construct(EntityTypeManagerInterface $entityTypeManager, TimeI
   /**
    * {@inheritdoc}
    */
-  public function refreshExpiredField(RefreshExpiredFieldEventInterface $event) {
-    // Only react to fields that store the download count or visit count.
-    if (!in_array($event->getFieldName(), [static::FIELD_NAME_DOWNLOAD_COUNT, static::FIELD_NAME_VISIT_COUNT])) {
-      return;
+  public function refreshExpiredFields(RefreshExpiredFieldsEventInterface $event) {
+    $items = $event->getExpiredItems()->getItems();
```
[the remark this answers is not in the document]
idimopoulos: It could. I think this was based on the previous `::getExpiredItem()` method name. It is not an issue of Joinup though and should be fixed upstream.
*Reasons to rename:* Item being completely consistent with the return type.
*Reasons to not:* Try to imagine that the `Items` part of the `getExpiredItems` refers to `Items` which is a vague term for someone that is not familiar with the code and knows that it is a Collection Item. Thus, the developer needs to check what kind of return objects are returned. If it was `getExpiredEntities` then it would probably be in need of renaming. But `getExpiredItems` are referring to "Items" which has no direct explanation. So a glimpse into the return type is anyway needed.

web/modules/custom/joinup_core/src/EventSubscriber/RefreshCachedFieldsEventSubscriber.php
idimopoulos: These 3 properties are put into the subqueries. Each sub query needs these information separately as they are handled independently. The reporting api module is only handling the parent query in which I do not set it.

web/modules/custom/joinup_core/joinup_core.post_update.php
idimopoulos: It wasn't working when I tried it... I removed it.

web/modules/custom/joinup_core/src/EventSubscriber/RefreshCachedFieldsEventSubscriber.php
idimopoulos: Ok, I re-added it, though I do not think that this is a case in Joinup. But I cannot predict the future.

web/modules/custom/joinup_core/src/EventSubscriber/RefreshCachedFieldsEventSubscriber.php
idimopoulos: True story is true.!

idimopoulos:
I am closing this in favour of https://github.com/ec-europa/joinup-dev/pull/1130

PR #1118, opened 2018-03-15
ISAICP-4197: Remove obsolete workaround.

pfrenssen:
A long time ago CPHP did not support Composer plugins, so we couldn't
use composer-patches unless we removed the packages that needed patching
and then reinstalled all dependencies ourselves. This has long been
fixed so we don't need this workaround any more.


idimopoulos:
I am merging though I find this functionality quite helpful.

PR #1121, opened 2018-03-16
Isaicp 4386

pfrenssen:


pfrenssen:
Closing after discussion with team members, our procedure is to not test moderator level functionality using the admin user. We don't need this.

PR #1130, opened 2018-03-26
Isaicp 4353: Piwik bulk operations - Joinup side

idimopoulos:


idimopoulos:
This is a new pull request that is against the parent ticket, not develop. It is mainly the same though. As soon as this is accepted, the parent ticket can be merged too.
QA remarks can be found at https://github.com/ec-europa/joinup-dev/pull/1116.

composer.json
```diff
@@ -42,7 +42,7 @@
         "drupal/persistent_login": "^1.0-alpha4",
         "drupal/phingdrushtask": "dev-7.x-2.x#f4b020f09da3c174fce40f2bd9a9617689f57124",
         "drupal/piwik": "~1.0",
-        "drupal/piwik_reporting_api": "^1.0-beta2",
+        "drupal/piwik_reporting_api": "dev-1.x",
```
sandervd: In the upstream, I would  try to refactor PiwikQueryFactory:
As it is a factory, all methods should be static to communicate that the factory itself is stateless.

web/modules/custom/joinup_core/src/EventSubscriber/RefreshCachedFieldsEventSubscriber.php
```diff
@@ -355,4 +317,48 @@ protected function getPiwikSettings($bundle) {
...
+    $url_parameter = $this->getUrlParameter($entity);
+
+    $sub_query = $this->piwikQueryFactory->getQuery($method);
+    $date_range = [
+      // If the period is 0 we should get all results since launch.
+      $period > 0 ? (new DateTimePlus("$period days ago"))->format('Y-m-d') : $this->configFactory->get('joinup_core.piwik_settings')->get('launch_date'),
+      (new DateTimePlus())->format('Y-m-d'),
+    ];
+
+    $parameters['period'] = 'year';
```
sandervd: When using the year range, this results in requesting the total of 'views' for the requested years.
This could lead to a weird sliding window, depending on the point in the year.
A solution could be to always request the previous and the current year.
That way at least the proportions won't be too much off.

web/modules/custom/joinup_core/src/EventSubscriber/RefreshCachedFieldsEventSubscriber.php
```diff
@@ -86,89 +86,81 @@ public function __construct(EntityTypeManagerInterface $entityTypeManager, TimeI
...
-      $count = 0;
-      foreach ($response as $year => $result) {
-        $item = $result[0];
-        if (!empty($item->$type)) {
-          $count = $count + (int) $item->$type;
-        }
-      }
-      return $count;
+    if (empty($url)) {
+      throw new JoinupCoreUndefinedUrlException();
```
sandervd: Please add a comment explaining when this can occur.

web/modules/custom/joinup_core/src/EventSubscriber/RefreshCachedFieldsEventSubscriber.php
```diff
@@ -86,89 +86,81 @@ public function __construct(EntityTypeManagerInterface $entityTypeManager, TimeI
...
     }
   }
 
   /**
-   * Returns the action count the given entity received in the given period.
+   * Gets the correct URL parameter for the query.
    *
    * @param \Drupal\Core\Entity\ContentEntityInterface $entity
-   *   The entity to check.
+   *   The entity from which to prune the URL.
```
sandervd: Don't know if prune is the right word here.

web/modules/custom/joinup_core/src/EventSubscriber/RefreshCachedFieldsEventSubscriber.php
idimopoulos: I will remove this entirely since it is meant to catch distributions without a file mainly because in any other case, the url will be present.
I will have the entity updated already with a count of 0 if that case is reached in order to have it excluded from the next update effort.

composer.json
idimopoulos: Do you think we should open an upstream ticket there?

PR #1132, opened 2018-03-27
Isaicp 4367: Prevent moderators to assign the Administrator role.

idimopoulos:


idimopoulos:
Merging after accepted in UAT today.

PR #1134, opened 2018-03-27
4285: Step to delete triples unsupported by Joinup

claudiu-cristea:


web/modules/custom/joinup_federation/src/Plugin/pipeline/Step/RemoveUnsupportedData.php
```diff
@@ -0,0 +1,65 @@
...
+      }
+    }
+    $rdf_entity_bundle_uris = SparqlArg::serializeUris($rdf_entity_bundle_uris);
+
+    $query = <<<Ouery
+DELETE FROM <$graph_uri> {
+  ?subject ?predicate ?object
+}
+WHERE {
+  ?subject ?predicate ?object
```
sandervd: Would this produce the same result?
DELETE FROM <$graph_uri> {
  ?subject ?predicate ?object
}
WHERE {
  ?subject ?predicate ?object .
  ?subject a ?entity_type .
  FILTER ( ?entity_type NOT IN ( $rdf_entity_bundle_uris ) ) .
}

PR #1136, opened 2018-03-29
Update Drupal core to 8.5.1

brummbar:


claudiu-cristea:
I included this patch in #1139 in order to prepare the new release which should contain also this.

PR #1137, opened 2018-03-30
4412: Abstract the pipeline logic in a contrib

claudiu-cristea:


web/modules/custom/pipeline/src/Controller/PipelineExecutionController.php
```diff
@@ -0,0 +1,126 @@
...
+
+  /**
+   * The pipeline orchestrator service.
+   *
+   * @var \Drupal\pipeline\PipelineOrchestratorInterface
+   */
+  protected $orchestrator;
+
+  /**
+   * The persistent state of the importer.
```
idimopoulos: This is not joinup specific any more so this is not an importer. Maybe "The state manager service" simply.

web/modules/custom/pipeline/src/Controller/PipelineExecutionController.php
```diff
@@ -0,0 +1,126 @@
...
+   */
+  protected $stateManager;
+
+  /**
+   * Constructs a new PipelineExecutionController object.
+   *
+   * @param \Drupal\pipeline\PipelineOrchestratorInterface $orchestrator
+   *   The pipeline orchestrator service.
+   * @param \Drupal\pipeline\PipelineStateManagerInterface $state_manager
+   *   The persistent state of the importer.
```
idimopoulos: Same as above, this needs to be updated in a generic non joinup definition.

web/modules/custom/pipeline/src/PipelineOrchestratorInterface.php
```diff
@@ -0,0 +1,26 @@
...
+
+namespace Drupal\pipeline;
+
+/**
+ * Provides an interface for the pipeline orchestrator.
+ */
+interface PipelineOrchestratorInterface {
+
+  /**
+   * Executes the orchestrator.
```
idimopoulos: This is very vague and I have no clue if I am right but it seems to me that an orchestrator (as mentioned in the orchestrator class) is merely a tracker for the step and the state while a pipeline is a series of steps, so in this term, the pipeline is the one executed, not the orchestrator.

web/modules/custom/pipeline/src/PipelinePermissions.php
```diff
@@ -0,0 +1,65 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function create(ContainerInterface $container) {
+    return new static($container->get('plugin.manager.pipeline_pipeline'));
+  }
+
+  /**
+   * Returns an array of node type permissions.
```
idimopoulos: Returns an array of pipeline related permissions.

web/modules/custom/pipeline/src/Plugin/PipelineStepInterface.php
```diff
@@ -0,0 +1,68 @@
...
+   *   An array of data to be passed to the execute method.
+   *
+   * @return null|array
+   *   If no errors were encountered during the step execution, nothing should
+   *   be returned. Return the error message as a render array.
+   */
+  public function execute(array &$data);
+
+  /**
+   * Gives a chances to plugins to perform some last tasks before executing.
```
idimopoulos: Gives a chance*

web/modules/custom/pipeline/src/Plugin/PipelineStepInterface.php
```diff
@@ -0,0 +1,68 @@
...
+   * @return \Drupal\pipeline\Plugin\PipelinePipelineInterface
+   *   The pipeline plugin instance.
+   */
+  public function getPipeline();
+
+  /**
+   * Allows this step to react after the pipeline execution.
+   *
+   * This method should be used to perform cleanup for objects, storage entries
+   * created by this step but required for the rest of the pipeline.
```
idimopoulos: Can you rephrase the `storage entries created by this step but required for the rest of the pipeline` please? It does not make much sense with the rest of the sentence. The sentence calls to clean up objects created in this step so probably you wanted to use `unless required for the rest of the pipeline`?

web/modules/custom/pipeline/src/Plugin/PipelineStepInterface.php
```diff
@@ -0,0 +1,68 @@
...
+   */
+  public function onPipelineSuccess();
+
+  /**
+   * Allows this step to react when the pipeline execution end with error.
+   *
+   * This method should be used to perform cleanup for objects, storage entries
```
idimopoulos: Same as above.

web/modules/custom/joinup_federation/tests/modules/joinup_federation_test/src/Plugin/pipeline/Pipeline/JoinupFederationTestingPipeline.php
```diff
@@ -0,0 +1,19 @@
...
+
+use Drupal\joinup_federation\JoinupFederationPipelinePluginBase;
+
+/**
+ * Provides a pipline testing plugin.
+ *
+ * @PipelinePipeline(
+ *   id = "joinup_federation_testing_pipeline",
+ *   label = @Translation("Joinup federation testing pipeline"),
+ *   steps = {
```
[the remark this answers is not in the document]
claudiu-cristea: True. As long as we don't perform pipeline operations, like `::next()`, `::current()`, etc. this will work. Otherwise an empty step list will throw an exception in `\Drupal\pipeline\Plugin\PipelinePipelinePluginBase::getStepList()`.

web/modules/custom/pipeline/src/Plugin/PipelineStepInterface.php
claudiu-cristea: I think the short description "tells the whole story". I improved that a bit.

PR #1138, opened 2018-04-03
Isaicp 4209: Provide the rdf-provenance entities and related API.

idimopoulos:


web/modules/custom/joinup_federation/src/Plugin/pipeline/Step/AttachProvenanceData.php
```diff
@@ -0,0 +1,116 @@
...
+}
+QUERY;
+
+    $sink = $this->getSinkGraphUri();
+    $query = str_replace('@sink_graph', $sink, $query);
+    $results = $this->sparql->query($query);
+    $activities = [];
+    foreach ($results as $result) {
+      $uri = $result->subject->getUri();
+      $activities[$uri] = $this->entityTypeManger->getStorage('rdf_entity')->create([
```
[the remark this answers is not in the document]
brummbar: @claudiu-cristea : please don't introduce personal preferences in the code style. The team has discussed multiple times that the "long" approach is valid. This is not a valid rejection point.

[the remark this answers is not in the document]
brummbar: Indeed. There is no recommended way. This is as correct as the short version, and actually this is the preferred way in our team. So no, this should not be changed to the short form. 
You already opened an issue in d.o regarding this (https://www.drupal.org/project/drupal/issues/2945539) and @pfrenssen already had a reply in there with the reason behind the whole team likes the full syntax more.

idimopoulos:
I am renaming the PR. I initially thought that this PR was related to the step that attaches the provenance entities to their respective solutions but as we discussed this morning, it is merely to provide the entities. I will remove all etl-related functionality from the ticket.

web/modules/custom/rdf_entity_provenance/rdf_entity_provenance.module
```diff
@@ -0,0 +1,63 @@
...
+  $fields['provenance_associated_with'] = BaseFieldDefinition::create('string')
+    ->setLabel(t('Provenance - Source'))
+    ->setDescription(t('The source that this entity was derived from.'))
+    ->setComputed(FALSE)
+    ->setCardinality(1)
+    ->setRequired(FALSE);
+
+  // A link towards the created local entity.
+  // @see: https://www.w3.org/TR/2013/REC-prov-o-20130430/#generated.
+  $fields['provenance_entity'] = BaseFieldDefinition::create('string')
```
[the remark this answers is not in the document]
idimopoulos: As we discussed this morning, this field will hold a URI related to the entity this provenance record is about. The thing though is that we will be holding records for entities we do not host/federate. Just to keep track of them. Thus, we cannot use a reference field for this unfortunately.

web/modules/custom/rdf_entity_provenance/src/ProvenanceHelper.php
```diff
@@ -0,0 +1,92 @@
...
+
+namespace Drupal\rdf_entity_provenance;
+
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\rdf_entity\RdfInterface;
+
+/**
+ * Provides helper methods to fetch, check and update provenance data.
+ */
+class ProvenanceHelper implements ProvenanceHelperInterface {
```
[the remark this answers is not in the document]
idimopoulos: I think it shortens up a lot the loadByProperties method and we are going to need that a lot. If it is not necessary, we can remove it.

web/modules/custom/rdf_entity_provenance/rdf_entity_provenance.module
```diff
@@ -0,0 +1,66 @@
...
+  // We are keeping a record for each entity available in a repository, whether
+  // we import it or not - in order to keep a track of the federation status.
+  // Thus, the field cannot be an entity reference.
+  // @see: https://www.w3.org/TR/2013/REC-prov-o-20130430/#generated.
+  $fields['provenance_entity'] = BaseFieldDefinition::create('string')
+    ->setLabel(t('Provenance - Entity'))
+    ->setDescription(t('The entity that was generated by this activity.'))
+    ->setComputed(FALSE)
+    ->setCardinality(1)
+    ->setRequired(FALSE);
```
claudiu-cristea: I think we cannot imagine a provenance entity missing, at least, this field.

web/modules/custom/rdf_entity_provenance/rdf_entity_provenance.module
```diff
@@ -0,0 +1,66 @@
...
+
+  // This field is not part of the provenance schema but will be used to
+  // blacklist entities so that they are not imported.
+  $fields['provenance_enabled'] = BaseFieldDefinition::create('boolean')
+    ->setLabel(t('Provenance - Enabled'))
+    ->setDescription(t('If set to false, the entity will not be federated.'))
+    ->setComputed(FALSE)
+    ->setDefaultValue(TRUE)
+    ->setCardinality(1)
+    ->setRequired(FALSE);
```
claudiu-cristea: This cannot be NULL, only TRUE or FALSE so it's required, no?

web/modules/custom/rdf_entity_provenance/src/ProvenanceHelper.php
```diff
@@ -0,0 +1,92 @@
...
+   * @param \Drupal\rdf_entity\RdfInterface $rdf_entity
+   *   The rdf entity.
+   *
+   * @return \Drupal\rdf_entity\RdfInterface
+   *   The provenance activity.
+   */
+  protected function createProvenanceActivity(RdfInterface $rdf_entity): RdfInterface {
+    /** @var \Drupal\rdf_entity\RdfInterface $activity */
+    $activity = $this->getStorage()->create([
+      'bundle' => 'provenance_activity',
```
claudiu-cristea: s/bundle/rid ?

web/modules/custom/rdf_entity_provenance/src/ProvenanceHelperInterface.php
```diff
@@ -0,0 +1,52 @@
...
+   * external source and thus has been federated.
+   *
+   * @param \Drupal\rdf_entity\RdfInterface $rdf_entity
+   *   The rdf entity.
+   *
+   * @return bool
+   *   Whether the entity has provenance record associated and thus, is
+   *   federated.
+   */
+  public function isFederated(RdfInterface $rdf_entity): bool;
```
claudiu-cristea: I'm a little bit reluctant about the added value of this method. As the implementation is saying, it's a simple "load by referred entity" which is accomplished by `::getProvenanceActivity()`

web/modules/custom/rdf_entity_provenance/src/ProvenanceHelper.php
```diff
@@ -0,0 +1,92 @@
...
+    return $activity;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function loadProvenanceActivity(RdfInterface $rdf_entity): ?RdfInterface {
+    /** @var \Drupal\rdf_entity\RdfInterface $activity */
+    $activity = $this->getStorage()->loadByProperties([
+      'bundle' => 'provenance_activity',
```
claudiu-cristea: s/bundle/rid ?

web/modules/custom/rdf_entity_provenance/src/ProvenanceHelperInterface.php
```diff
@@ -0,0 +1,52 @@
...
+  /**
+   * Loads or creates a provenance record for the passed rdf entity.
+   *
+   * @param \Drupal\rdf_entity\RdfInterface $rdf_entity
+   *   The rdf entity.
+   *
+   * @return \Drupal\rdf_entity\RdfInterface|null
+   *   The provenance activity related to the rdf entity passed.
+   */
+  public function getProvenanceActivity(RdfInterface $rdf_entity): RdfInterface;
```
claudiu-cristea: What if I want to get the provenance for a blacklisted entity? How do I pass the `$rdf_entity`? I guess we have to pass the ID (uri) of the entity to cover also assets not imported on our system (AKA not having Drupal entities). Also the naming should reflect, I think, the fact that we are passing the referred entity, like `getProvenanceByReferredEntity()`?

web/modules/custom/rdf_entity_provenance/src/ProvenanceHelperInterface.php
```diff
@@ -0,0 +1,52 @@
...
+   * Retrieves the provenance activity related to an rdf entity.
+   *
+   * @param \Drupal\rdf_entity\RdfInterface $rdf_entity
+   *   The rdf entity.
+   *
+   * @return \Drupal\rdf_entity\RdfInterface|null
+   *   The provenance activity related to the rdf entity passed, or null if no
+   *   record exists.
+   */
+  public function loadProvenanceActivity(RdfInterface $rdf_entity): ?RdfInterface;
```
claudiu-cristea: Same for `@param` type and naming.

web/profiles/joinup/joinup.info.yml
```diff
@@ -70,6 +70,7 @@ dependencies:
   - r4032login
   - rdf_draft
   - rdf_entity
+  - rdf_entity_provenance
```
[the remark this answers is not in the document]
claudiu-cristea: This was not addressed.

web/modules/custom/rdf_entity_provenance/src/ProvenanceHelper.php
```diff
@@ -0,0 +1,92 @@
...
+    return $activity;
+  }
+
+  /**
+   * Retrieves the rdf_entity storage.
+   *
+   * @return \Drupal\Core\Entity\EntityStorageInterface
+   *   The entity storage.
+   */
+  protected function getStorage() {
```
claudiu-cristea: It's OK that we don't statically cache the storage as it cannot be used twice in the same request.

web/modules/custom/rdf_entity_provenance/rdf_entity_provenance.module
```diff
@@ -0,0 +1,66 @@
...
+    ->setComputed(FALSE)
+    ->setCardinality(1)
+    ->setRequired(FALSE);
+
+  // A link towards the entity that the provenance record stores metadata about.
+  // We are keeping a record for each entity available in a repository, whether
+  // we import it or not - in order to keep a track of the federation status.
+  // Thus, the field cannot be an entity reference.
+  // @see: https://www.w3.org/TR/2013/REC-prov-o-20130430/#generated.
+  $fields['provenance_entity'] = BaseFieldDefinition::create('string')
```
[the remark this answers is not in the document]
claudiu-cristea: Also we have to add a constraint to this field and ensure its unicity. We cannot have 2 provenance entities referring the same entity.

web/modules/custom/rdf_entity_provenance/src/ProvenanceHelper.php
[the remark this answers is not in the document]
claudiu-cristea: OK, let's keep it but see comments I made on each method.

web/modules/custom/rdf_entity_provenance/src/ProvenanceHelper.php
claudiu-cristea: And also, `:: loadByProperties()` returns an array. We should use `reset()` or `array_shift()` to pickup the first item and ensure a NULL if the array is empty.

web/modules/custom/rdf_entity_provenance/rdf_entity_provenance.module
```diff
@@ -0,0 +1,66 @@
...
+  // @see: https://www.w3.org/TR/2013/REC-prov-o-20130430/#startedAtTime
+  $fields['provenance_started'] = BaseFieldDefinition::create('timestamp')
+    ->setLabel(t('Provenance - Created'))
+    ->setDescription(t('The time that the provenance entity was created.'))
+    ->setComputed(FALSE)
+    ->setCardinality(1)
+    ->setRequired(FALSE);
+
+  // The source of the entity is a link to the Agent. However, an Agent property
+  // is not strictly needed so this will be a link towards the source of the
```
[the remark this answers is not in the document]
idimopoulos: The thing here is that the Agent is a good way to track the source of information. However, in our current implementation, the source is not  part of our pipeline epic ticket.
As we discussed yesterday during the standup (invalid information for github), for tracking new entities, it would be nice to have the source tracked. This field is meant to achieve just that.

The provenance entities, if you follow the link in the ticket and the attached images, can come in many ways and patterns. We picked the simplest one and even more we stripped off properties that are not necessary. The provenance entities after all are only for our internal use.
Sander's suggestion from yesterday, to hardcode a uri in the pipeline (e.g. the uri of the ctt repo) seems like a solution that can help us with that.

web/modules/custom/rdf_entity_provenance/rdf_entity_provenance.module
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * Implements hook_entity_bundle_field_info().
+ */
+function rdf_entity_provenance_entity_bundle_field_info(EntityTypeInterface $entity_type, $bundle, array $base_field_definitions) {
+  if ($entity_type->id() !== 'rdf_entity' || $bundle !== 'provenance_activity') {
+    return [];
+  }
+
+  $fields = [];
```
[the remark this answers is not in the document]
idimopoulos: I know :( but the fields `started`, `entity` and `enabled` can easily be confused as properties with fields like `created`, any referenced entity, and `status` (or similar). I wanted to keep them easily identifiable.

web/modules/custom/rdf_entity_provenance/config/install/rdf_entity.rdfentity.provenance_activity.yml
```diff
@@ -0,0 +1,10 @@
+langcode: en
+status: true
+dependencies: {  }
+name: 'Provenance activity'
+rid: provenance_activity
```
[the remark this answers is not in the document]
idimopoulos: The 'activity' keyword comes from the provenance pattern. The activity describes the procedure that imports the entity and the other 2 provenance entities of our schema are the Agent and the Entity. The provenance keyword as well is for the same reason as the fields above.

web/modules/custom/rdf_entity_provenance/rdf_entity_provenance.module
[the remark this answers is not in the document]
idimopoulos: This is a thin line. It is convenient to have an entity reference but in my mind, it does not fit well to have an entity reference when it is not meant to refer an entity. A referral to an entity is the matter only when we host the entity. 
We did the same thing with the file field url. In the end, it is a url, whether locally it exists as an entity or not, does not matter for the type of the field.
Also, is this ok for when saving the entity? As far as I know, when you try to point to a non existing entity through a reference field, there is a cleanup happening to the array of values that depends on whether the referred entity exists..

web/modules/custom/rdf_entity_provenance/rdf_entity_provenance.module
```diff
@@ -0,0 +1,66 @@
...
+  // The time that the activity started to federate the remote entity. This is
+  // different from the created date as it can be updated if the entity is
+  // federated again.
+  // @see: https://www.w3.org/TR/2013/REC-prov-o-20130430/#startedAtTime
+  $fields['provenance_started'] = BaseFieldDefinition::create('timestamp')
+    ->setLabel(t('Provenance - Created'))
+    ->setDescription(t('The time that the provenance entity was created.'))
+    ->setComputed(FALSE)
+    ->setCardinality(1)
+    ->setRequired(FALSE);
```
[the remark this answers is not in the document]
claudiu-cristea: Just removing the method has no effect. If you inspect:
```
\Drupal::service('entity_field.manager)
  ->getFieldDefinitions('rdf_entity', 'provenance_activity')['provenance_started']
  ->isRequired();
```
you'll get `FALSE`.

So, we need to explicitly set `TRUE`: `->setRequired(TRUE)`

web/modules/custom/rdf_entity_provenance/rdf_entity_provenance.module
claudiu-cristea: Same as for `provenance_started`.

web/modules/custom/rdf_entity_provenance/rdf_entity_provenance.module
claudiu-cristea: Same as for `provenance_started`.

web/modules/custom/rdf_entity_provenance/rdf_entity_provenance.module
```diff
@@ -0,0 +1,63 @@
...
+    return [];
+  }
+
+  $fields = [];
+
+  // The time that the activity started to federate the remote entity. This is
+  // different from the created date as it can be updated if the entity is
+  // federated again.
+  // @see: https://www.w3.org/TR/2013/REC-prov-o-20130430/#startedAtTime
+  $fields['provenance_started'] = BaseFieldDefinition::create('timestamp')
```
claudiu-cristea: I wonder why we don't map the `changed` base field to `http://www.w3.org/ns/prov#startedAtTime` and get rid of this field. The advantage is that, we get the default value automatically. For example, with the actual code, before saving an entity created with `ProvenanceHelper::createProvenanceActivity()` we need to assign explicitly the `provenance_started`. But `changed` does it automatically, so when saving we'll get the correct time without worrying.

idimopoulos: Because normally, the provenance entity has both `provenance_started` and `provenance_ended`. The timestamp here should take the time that the import started, not the one that the entity was saved. The time difference is of course not of vital value nor will diverge a lot, but for example, if we were to import thousands of entities, we would fetch them, we would parse them and we would store them. In this case, a time difference exists between the time that the fetching started (provenance_started) until the conversion/validation and finally the saving of the entity.

PR #1141, opened 2018-04-09
ISAICP-4436: Solution counter on header vs solution counter on facet.

idimopoulos:


idimopoulos:
@claudiu-cristea sorry, I forced pushed since the previous solution was not correct and there is no point to pollute the repo with reverts when it is simply two commits. Your comment's reference point was lost as a result. In any case, you were right. The issue was indeed completely elsewhere.

PR #1142, opened 2018-04-09
ISAICP-4435: add a rel nofollow attribute to the Advanced Search link

brummbar:


tests/features/collection/homepage.feature
```diff
@@ -151,6 +151,12 @@ Feature: Collection homepage
       | Rohirrim make extraordinary deal                  |
       | Breaking: Gandalf supposedly plans his retirement |
 
+  Scenario: Search engines and link crawlers should not follow advanced search link.
+    Given I go to the homepage of the "Middle earth daily" collection
+    Then search engines should be discouraged to follow the link "Advanced search"
```
claudiu-cristea: This custom step definition looks too specialised to be reusable. We should avoid creating new step definitions when we can accomplish the same goal with existing definitions. I think `Then I should see an "a:contains('Advanced search')[rel=nofollow]" element` should work and this step definition is part of the Mink context so it's pretty standard. Then we can remove the custom step definition.

sandervd: I'm all in favor of simplifying, but at the other hand there is the convention to use as much as possible business language in the tests. Although the existing step tests exactly the same, the custom step makes it much more obvious what is actually being tested.

claudiu-cristea: I see the "business language" point. However, we have to balance somehow between the 2. Just creating a step definition for each business case would create huge custom step definition repositories where most of the definitions would be used in only one place. I'm for reusing as much as we can. Less code to maintain. But I'm not opposing the actual solution too. @sandervd it's your decision.

claudiu-cristea: OK, let's keep the new step definitions.

PR #1155, opened 2018-04-19
ISAICP-4452: Apply Drupal 8.5.2 security update

pfrenssen:
This updates the distribution to the latest 8.5.2 release of Drupal core.

As part of this update Composer needed to be locked on the previous release, since the latest 1.6.4 release has a critical issue that causes packages to go missing from `composer.lock`, ref: https://github.com/composer/composer/issues/7268.


claudiu-cristea:
Now we need a follow up to do the same with `develop`.

PR #1157, opened 2018-04-19
4423: Tallin declaration submission permissions

claudiu-cristea:


tests/features/communities/tallinn/permissions.feature
```diff
@@ -0,0 +1,65 @@
...
+    # A user outside of Tallinn collection is not able to edit any report.
+    Given I am logged in as gheorghe
+    When I visit the tallinn_report content "Romania" edit screen
+    Then I should get a 403 HTTP response
+    When I visit the tallinn_report content "France" edit screen
+    Then I should get a 403 HTTP response
+
+    # A moderator is able to change any report's author.
+    Given I am logged in as chef
+    When I visit the tallinn_report content "Romania" edit screen
```
idimopoulos: :D :D I would also have used "Greece" most probably if I were to write that test :P

tests/features/communities/tallinn/permissions.feature
```diff
@@ -0,0 +1,71 @@
...
+      | Kind of Tallinn | dominique |
+    And tallinn_report content:
+      | title          | author    | collection      |
+      | Romania Report | gheorghe  | Kind of Tallinn |
+      | France Report  | dominique | Kind of Tallinn |
+
+    # A user outside of Tallinn collection is not able to edit any report.
+    Given I am logged in as gheorghe
+    When I visit the tallinn_report content "Romania Report" edit screen
+    Then I should get a 403 HTTP response
```
idimopoulos: Is there a specific requirement that the user has to be a member to access this? I don't think we have that restriction on Tallinn and I also don't think that the "reporters" will be joinup common users at all. At least not necessarily.

web/profiles/joinup/joinup.profile
```diff
@@ -220,6 +220,10 @@ function joinup_form_node_form_alter(&$form, FormStateInterface $form_state, $fo
   $form['revision']['#access'] = FALSE;
   $form['meta']['#access'] = FALSE;
 
+  /** @var \Drupal\Core\Entity\EntityFormInterface $form_object */
+  $form_object = $form_state->getFormObject();
+  $form['uid']['#access'] = \Drupal::currentUser()->hasPermission('administer nodes') && ($form_object->getEntity()->bundle() === 'tallinn_report');
```
idimopoulos: This can be easily transferred to the hook in the tallinn module that you already created above. This will also allow us to maintain the specific community functionality all tied with the tallinn module.

tests/features/communities/tallinn/permissions.feature
claudiu-cristea: Probably regular users can see the reports. This could be tested (and implemented if case).

PR #1163, opened 2018-04-25
4213: 3-way merge

claudiu-cristea:


web/modules/custom/joinup_federation/src/Plugin/pipeline/Step/ThreeWayMerge.php
```diff
@@ -0,0 +1,269 @@
...
+        $needs_save = TRUE;
+
+        // Delete the incoming entity from the staging graph.
+        $incoming_entity->skip_notification = TRUE;
+        $incoming_entity->delete();
+      }
+
+      if ($needs_save) {
+        $local_entity->skip_notification = TRUE;
+        $local_entity->save();
```
idimopoulos: Are we sure that we want to save the entities here? We have the ticket about filtering the solutions to import and we should at least add a `@todo` that the provenance entities should be handled along.

Though I still consider whether this should be handled in the form submit of the entities to import filtering.

claudiu-cristea: The provenance entities will be created/updated one step before this step, in user filtering step. At that point the entities are already in the `staging` graph. What are we doing here is only to move them from `staging` to `default` for new entities and copy over for existing entities. I think this was what has been discussed yesterday with Sander.

web/modules/custom/joinup_federation/tests/src/Kernel/ThreeWayMergeStepTest.php
```diff
@@ -0,0 +1,152 @@
...
+
+    // Check that the step ran without any error.
+    $this->assertNull($result);
+
+    /** @var \Drupal\rdf_entity\RdfInterface $solution */
+    $solution = Rdf::load('http://asset');
+
+    // Check that an existing entity values are overridden.
+    $this->assertEquals('Asset', $solution->label());
+    $this->assertEquals('This is an Asset.', $solution->get('field_is_description')->value);
```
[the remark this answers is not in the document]
claudiu-cristea: I added the `field_status` in the test. See the last commit.

PR #1166, opened 2018-04-25
Isaicp 4111

pfrenssen:


pfrenssen:
What was done in this PR:

* Notification message was updated and test coverage added.
* The original notification was developed before we were using Message Notify, and was still sent using the core mail manager and the message content was managed through the admin UI. This is converted now to a Message.
* The message requires a number of message arguments that are also used in other messages (e.g. `@actor:*`, `@group:*` etc). I put these in reusable static methods in a helper class `MessageArgumentGenerator`. I cleaned up all duplicate code for the arguments I used. This resulted in the removal of ~150 duplicated lines.
* The consolidation of the message arguments in `MessageArgumentGenerator` uncovered that we had two duplicate arguments for the group title: `@group:label` and `@group:title`. I unified this to `@group:title` since this is consistent with `@entity:title`.
* Updated PHP 7 typing that I encountered in the files that I touched.
* Documented why in Joinup our group administrator roles have the `is_admin` flag disabled. This set me on a wrong path, I mistakenly thought we were dealing with a bug which I spent a day trying to fix before realizing this was intentional and working as designed: https://github.com/Gizra/og/pull/370
* I needed to reuse a bunch of code from `InvitationMessageHelper`. This went through a couple of iterations, but I settled on creating a simple service with some helper methods for creating messages related to any kind of entity: `EntityMessageHelper`. This was initially a base class but this was a bad idea, and I ended up creating a standalone service which can be injected into any more specialized service like `InvitationMessageHelper`.
* Removed obsolete code regarding mail tokens that we were using in the distant past.

web/modules/custom/joinup_user/config/install/message.template.og_membership_role_change.yml
```diff
@@ -0,0 +1,19 @@
...
+dependencies: {  }
+template: og_membership_role_change
+label: 'Group membership role change'
+description: 'Message sent when a user receives a new role inside a collection or solution.'
+text:
+  -
+    value: '<p>Your role has been changed to @membership:roles</p>'
+    format: full_html
+  -
+    value: '<p>@actor:full_name has changed your role in @group:bundle "@group:title" to @membership:roles.</p><p>If you think this action is not clear or not due, please contact Joinup Support.</p>'
```
idimopoulos: According to the ticket, the last part, 'please contact Joinup Support', should contain a link to the support page (contact form). This link should be the 'Joinup Support' phrase.

web/modules/custom/joinup_notification/src/EntityMessageHelperInterface.php
```diff
@@ -0,0 +1,96 @@
...
+   * @param int $limit
+   *   The maximum number of entities to return. Defaults to 10 messages.
+   * @param string $order
+   *   Whether to sort by ascending or descending creation date. Defaults to
+   *   descending order.
+   *
+   * @return \Drupal\Core\Entity\EntityInterface[]
+   *   The corresponding messages.
+   */
+  public function getMessages(EntityInterface $entity, string $template, string $field_name, array $values = [], int $limit = 10, string $order = self::SORT_DESC): array;
```
idimopoulos: The `$order` here is never changed in our codebase and we only use the `DESC` order to sort the messages. I don't think that it is needed as a parameter which also makes the constants uneeded, especially the `SORT_ASC` which is never used. I know it is normal that we have it because we have the `SORT_DESC` one too, but these are never used so the parameter is not needed at all.

web/modules/custom/joinup_notification/src/EntityMessageHelper.php
```diff
@@ -0,0 +1,131 @@
...
+ * To send these messages, use the JoinupMessageDelivery service.
+ *
+ * There are different use cases for sending messages, this is only intended for
+ * basic creation and retrieval of messages. In case your entity requires custom
+ * business logic then feel free to inject this service and leverage it. See for
+ * example the InvitationMessageHelper which implements default arguments and
+ * resends a single message for each invitation instead of creating a new
+ * message every time.
+ *
+ * Because of the wild variations in business logic regarding messages this was
```
idimopoulos: I felt a small pat on my shoulder like someone was understanding me the moment I read the adjective 'wild' that described the word 'variations' of the notification messages' business logic.

composer.json
```diff
@@ -184,6 +184,9 @@
             "drupal/message_digest": {
                 "Aggregated patch for issues #2285947 #2881263 #2933016 #2933896 #2933931 #2934111 @see https://www.drupal.org/project/issues/message_digest": "resources/patch/message_digest_aggregated.patch"
             },
+            "drupal/og": {
+                "OgRole::isAdmin() should return TRUE for administrators": "https://patch-diff.githubusercontent.com/raw/Gizra/og/pull/370.patch"
```
idimopoulos: As you comment in the case where this is used, we cannot use the `::isAdmin` method to ensure that a role is an administrator role. This patch is an improvement, but not needed.

web/modules/custom/joinup_notification/src/MessageArgumentGenerator.php
```diff
@@ -0,0 +1,167 @@
...
+   *   Optional user object for the actor. Defaults to the current user.
+   *
+   * @return array
+   *   An associative array of actor data, with the following keys:
+   *   - '@actor:field_user_family_name': The last name of the actor.
+   *   - '@actor:field_user_first_name': The first name of the actor.
+   *   - '@actor:full_name': The full name of the actor.
+   *   - '@actor:role': Will sometimes contain the word 'moderator'.
+   *
+   * @throws \Drupal\Core\TypedData\Exception\MissingDataException
```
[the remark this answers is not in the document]
pfrenssen: It is thrown in the `->first()` calls when retrieving the first and last name, see `ListInterface::first()`.

web/modules/custom/joinup_notification/src/EntityMessageHelperInterface.php
pfrenssen: But then how are you going to retrieve a list of messages in ascending order if you need it?

This is a generic service to retrieve any kind of message related to any kind of entity, it is not limited to the use case of this particular issue, or I would have called it the `OgRoleChangeNotificationHelperInterface`.

web/modules/custom/joinup_notification/src/EntityMessageHelper.php
pfrenssen: Yeah this was written after many hours of fruitless searching for a possible API that fits all use cases. Such a thing we will not see in our lifetimes.

PR #1173, opened 2018-04-30
ISAICP-4466: Set expiration timers on cached computed fields to a daily level

idimopoulos:


web/modules/custom/joinup_core/joinup_core.post_update.php
```diff
@@ -90,3 +90,30 @@ function joinup_core_post_update_remove_action_transfer_solution_ownership() {
 function joinup_core_post_update_install_tallinn() {
   \Drupal::service('module_installer')->install(['tallinn']);
 }
+
+/**
+ * Update the counter refresh times and refresh the snapshot [ISAICP-4466].
+ */
+function joinup_core_post_update_update_cached_timers() {
```
pfrenssen: Do we still need this now the upstream bug in Config Sync is fixed? This seems to be a workaround for that exact problem.

idimopoulos: I have tested this. I removed the manual snapshot refresh in the post update as the whole thing is working without it. However, the bug that was fixing this, was already resulting in the configuration within that post update being already outdated and wrongly detected. We cannot have it fixed without having a separate release unless we do it manually in a post update. 

After this though, we will not need to update config through a post update again, unless needed for the hook.

pfrenssen: I don't understand the implications of your comment, my understanding is that this command in our deployment script `update-production.sh` will make sure the config will be updated:

```
./vendor/bin/drush updatedb --yes &&
./vendor/bin/drush cs-update --discard-overrides --yes &&
./vendor/bin/drush cache-rebuild --yes
```

So this will first run the database updates (including post-updates, no longer needed in this case), then apply the config changes with the `--discard-overrides` option, meaning that anything that might be incorrect in the database will be ignored and overwritten.

Do you believe this will not work for some reason? You mention that the configuration was outdated and wrongly detected, but isn't that just because the post-updates run before the config is applied?

The only scenario I can imagine where this wouldn't work is if Config Sync might think that the config value has not changed, and won't attempt to update it. This decision is based on the information in the snapshot, so it should work in theory. Of course there might be a bug preventing it from working correctly.

idimopoulos: Ok so think of this as: 
Active storage (config table) as A.
Snapshot storage (config_sync_snapshot_extension table) as S
and files (configuration in config install folders) as F.
And let's assume that some value somewhere is now "1".

So the normal flow is
* Active storage has the value 1, snapshot has the value 1 and files have the value one. So:
A:1, S:1, F:1.
* We do a change so we have A:1, S:1, F:2.
* We run `cs-update`
* The config sync module checks for changes by comparing S and F (not A and F).
* The F:2 is detected.
* The A is updated with the new value. So we have A:2, S:1, F:2.
* A new snapshot is created by updating the S compared to the F so we have A:2, S:2, F:2.

What happened now though, is that 
* Active storage has the value 1, snapshot has the value 1 and files have the value one. So:
A:1, S:1, F:1.
 On the above, 1, was the correct value, but we had to update it for a bit. So we updated the value to 2 so we had A:2, S:1, F:2. This can happen with a normal update or with a direct update through a hook. No matter what, the value in S remains as 1.
* The snapshot failed to update S properly due to the bug so S remained 1.
* Now we need to revert the value and we changed F to 1 so we had A:2, S:1, F:1. 
* Since the config sync compares the snapshot to the file system, the updated value is not detected. For the same reason, if we have the S updated, which will be compared to F, the result will still be A:2, S:1, F:1. So no updates will be detected after that too.

pfrenssen: OK so would this then be solved if we run a command to drop the snapshot before running the config sync?

web/modules/custom/joinup_core/joinup_core.post_update.php
```diff
@@ -111,9 +111,4 @@ function joinup_core_post_update_update_cached_timers() {
     $editable->setData($data);
     $editable->save();
   }
-
-  // Refresh the snapshot properly.
-  $config_sync_snapshotter = \Drupal::service('config_sync.snapshotter');
-  $config_sync_snapshotter->deleteSnapshot();
-  $config_sync_snapshotter->refreshSnapshot();
```
pfrenssen: I would still remove the entire post update function, and especially the part that sets the config. The code to refresh the snapshot could stay, but even that seems unnecessary. This problem will sort itself out soon enough, if not on the first deployment, then on the second.

pfrenssen:
Thanks!

PR #1179, opened 2018-05-08
ISAICP-4461: show the abstract at the top of the collection overview page

brummbar:


web/modules/custom/collection/config/install/rdf_entity.rdfentity.collection.yml
```diff
@@ -1,11 +1,10 @@
...
 dependencies: {  }
-third_party_settings: {  }
 name: Collection
 rid: collection
 description: 'An Asset Repository stores descriptions of Assets and Asset Distributions.'
 label_singular: collection
 label_plural: collections
 label_count:
-  no_count_capitalize: "Collection\x03Collections"
+  no_count_capitalize: !!binary Q29sbGVjdGlvbgNDb2xsZWN0aW9ucw==
```
claudiu-cristea: I understand how this has landed here. However, I prefer the former version which is, at least, human readable. But the whole file change makes no sense. Even the config export is touching a lot of files we should keep in PR only the relevant changes so that the QA is not polluted by noise.

web/modules/custom/collection/config/install/rdf_entity.mapping.rdf_entity.collection.yml
```diff
@@ -5,11 +5,13 @@ dependencies:
     - rdf_entity.graph.default
     - rdf_entity.graph.draft
     - rdf_entity.rdfentity.collection
-third_party_settings: {  }
 id: rdf_entity.collection
 entity_type_id: rdf_entity
 bundle: collection
 rdf_type: 'http://www.w3.org/ns/adms#AssetRepository'
+graph:
```
claudiu-cristea: The whole file change makes no sense and is out-of-scope. It only adds noise to the PR. Please revert.

web/modules/custom/collection/collection.module
```diff
@@ -372,6 +379,30 @@ function collection_rdf_entity_view(array &$build, EntityInterface $entity, Enti
...
+ */
+function collection_rdf_entity_view_alter(array &$build, EntityInterface $entity, EntityViewDisplayInterface $display) {
+  /** @var \Drupal\rdf_entity\RdfInterface $entity */
+  if ($entity->bundle() !== 'collection' || $display->getMode() !== 'default') {
+    return;
+  }
+
+  // Hide the description when the abstract is available.
+  if (Element::children($build['field_ar_abstract'])) {
+    $build['field_ar_description']['#access'] = FALSE;
```
claudiu-cristea: Is there a good reason why we are not disabling the description via the entity view display? I commented this line, then I disabled the field from the entity display and I achieved the same goal. Also, I'm repeating my belief here (and I will do it again and again): We should switch, as much as we can, from maintaining code to maintaining config.

web/modules/custom/collection/collection.module
```diff
@@ -320,17 +321,23 @@ function collection_entity_extra_field_info() {
     'visible' => FALSE,
   ];
 
+  $display['read_more'] = [
```
claudiu-cristea: Just a side note: Instead of hardcoding we should go more and more with "DS way". I think such fields can be easily configured in DS instead of coding them. But this is just my thinking for the future, not for this specific PR.

web/modules/custom/collection/collection.module
brummbar: This is done because part of the scope of the ticket is to show the description only when the abstract is not available. So we cannot use configuration here.

PR #1185, opened 2018-05-14
Isaicp-4377: Add the rest of the conversion passes.

idimopoulos:


web/modules/custom/joinup_federation/src/Plugin/Adms2ConvertPass/Pass41.php
```diff
@@ -0,0 +1,100 @@
...
+ *
+ * URI: dct:issued
+ * Type: N.A.
+ * Action: Updated
+ * Description:
+ * - The cardinality changed to single valued. Only the first value will be
+ * kept.
+ *
+ * @Adms2ConvertPass(
+ *   id = "pass_41",
```
claudiu-cristea: We decided to take the "conservative" approach. Meaning that we can make any assumptions is such cases. If the incoming value have the cardinality of 1, then will pass the later ADMS-AP validation. But if the cardinality is >=2 then we cannot assume that we pickup the first, last, etc. Conclusion:
* Either remove this plugin/file.
* Or remove the annotation so that the plugin manager will not pick it up, add a `@todo` explaining the reason.

web/modules/custom/joinup_federation/src/Plugin/Adms2ConvertPass/Pass45.php
```diff
@@ -0,0 +1,101 @@
...
+ *
+ * URI: dct:publisher
+ * Type: N.A.
+ * Action: Updated
+ * Description:
+ * - The cardinality changed to single valued. Only the first value will be
+ * kept.
+ *
+ * @Adms2ConvertPass(
+ *   id = "pass_45",
```
claudiu-cristea: Same as for step 41.

web/modules/custom/joinup_federation/tests/src/Kernel/ConvertToAdms2StepTest.php
```diff
@@ -19,30 +19,25 @@ class ConvertToAdms2StepTest extends StepTestBase {
   public function test() {
     /** @var \Drupal\joinup_federation\JoinupFederationAdms2ConvertPassPluginManager $plugin_manager */
     $plugin_manager = $this->container->get('plugin.manager.joinup_federation_adms2_convert_pass');
+    $sink_graph = static::getTestingGraphs()['sink'];
+    $data = ['sink_graph' => $sink_graph];
 
-    $rdf_data = '';
```
claudiu-cristea: Very nice change! 👍

web/modules/custom/joinup_federation/src/Plugin/Adms2ConvertPass/Pass37.php
```diff
@@ -30,28 +32,60 @@ class Pass37 extends JoinupFederationAdms2ConvertPassPluginBase {
...
+      WHERE {
+        ?subject a ?type .
+        ?subject <http://www.w3.org/ns/dcat#downloadURL> ?download_url .
+        FILTER NOT EXISTS { ?subject <http://www.w3.org/ns/dcat#accessURL> ?download_url } .
+        VALUES ?type { <http://www.w3.org/ns/adms#AssetDistribution> <http://www.w3.org/ns/dcat#Distribution> } .
+      }
+QUERY;
+
+    // The delete query is separated so that values that are not converted to
+    // access url are also cleaned up.
```
sandervd: Nice approach!

web/modules/custom/joinup_federation/src/Plugin/Adms2ConvertPass/Pass41.php
```diff
@@ -0,0 +1,99 @@
...
+   */
+  public function convert(array $data): void {
+    $results = $this->getTriplesFromGraph($data['sink_graph'],
+      NULL,
+      'http://www.w3.org/ns/dcat#issued'
+    );
+
+    $entities = [];
+    // In order to be fair, and since the notion of deltas does not exist in
+    // SPARQL, we are only going to keep the first instance found for the given
```
sandervd: Well, you can still model this (see adms:nextVersion/previousVersion), is a bit awkward indeed...

web/modules/custom/joinup_federation/src/Plugin/Adms2ConvertPass/Pass41.php
```diff
@@ -0,0 +1,99 @@
...
+      if (isset($entities[$result['subject']])) {
+        continue;
+      }
+      $entities[$result['subject']] = $result['object'];
+    }
+
+    foreach ($entities as $subject => $object) {
+      $query = <<<QUERY
+WITH <{$data['sink_graph']}>
+DELETE { <$subject> <http://www.w3.org/ns/dcat#issued> ?issued }
```
sandervd: For the issued it is probably fine to delete, for others it is probably safer to rely on the validator.

web/modules/custom/joinup_federation/src/Plugin/Adms2ConvertPass/Pass45.php
```diff
@@ -0,0 +1,102 @@
...
+        continue;
+      }
+      $entities[$result['subject']] = $result['object'];
+    }
+
+    foreach ($entities as $subject => $object) {
+      $query = <<<QUERY
+WITH <{$data['sink_graph']}>
+DELETE { <$subject> <http://www.w3.org/ns/dcat#publisher> ?publisher }
+INSERT { <$subject> <http://www.w3.org/ns/dcat#publisher> <$object> }
```
sandervd: I'm a bit conflicted here, but as we can always re-federate, it is probably fine to discard the other publishers.

web/modules/custom/joinup_federation/src/Plugin/Adms2ConvertPass/Pass41.php
idimopoulos: I have discussed with @claudiu this class and we are going to remove it as we are using the conservative approach.

web/modules/custom/joinup_federation/src/Plugin/Adms2ConvertPass/Pass45.php
idimopoulos: This is also going to be removed. Sorry for not having done that yet.

PR #1199, opened 2018-05-23
Isaicp 4500: Fix predicates for solution_type and test_resource_type fields

idimopoulos:


web/modules/custom/joinup_core/joinup_core.install
```diff
@@ -68,3 +68,29 @@ function joinup_core_update_8101() {
...
+QUERY;
+  $query->query($test_resource_type_query);
+  $solution_type_query = <<<QUERY
+DELETE { GRAPH ?g { ?subject <http://purl.org/dc/terms/DCMIType> ?object } }
+INSERT { GRAPH ?g { ?subject <http://purl.org/dc/terms/type> ?object } }
+USING <http://joinup.eu/solution/draft>
+USING <http://joinup.eu/solution/published>
+WHERE { GRAPH ?g { ?subject <http://purl.org/dc/terms/DCMIType> ?object } }
+QUERY;
+  $query->query($solution_type_query);
```
sandervd: It seems that the query is a bit too 'greedy', and updates triples outside of the selected graphs:
after the update the following query returns results:
SELECT count(?subject)
FROM NAMED <http://joinup.eu/licence/published>
WHERE {
  GRAPH <http://joinup.eu/licence/published> {
    ?subject <http://joinup.eu/test_resource_type> ?object 
  }
}

Issue #1196, opened 2018-05-22
Taxonomy issue


cassm199:
Hi,

We've deployed JoinUp on our Ubuntu Server environment, however when trying to create a new collection, the policy domain does not return any values, however the Policy Domain taxonomy contains a number of values. With regards to this particular taxonomy, upon trying to edit or deleting any one of the terms, the below error is returned:

The website encountered an unexpected error. Please try again later.</br></br><em class="placeholder">InvalidArgumentException</em>: Graph &#039;draft&#039; doesn&#039;t exist for entity type &#039;taxonomy_term&#039;. in <em class="placeholder">Drupal\rdf_entity\Entity\RdfEntitySparqlStorage-&gt;Drupal\rdf_entity\Entity\{closure}()</em> (line <em class="placeholder">1264</em> of <em class="placeholder">modules/contrib/rdf_entity/src/Entity/RdfEntitySparqlStorage.php</em>). 

These are the specs:

Drupal: 8.5.3
Ubuntu :16
PHP: 7.1
Database: MySQL

sandervd:
Hi @cassm199,

Indeed, the policy domain taxonomy has been populated during the migration of Joinup from Drupal 6 to Drupal 8. Could you try the following:
- Extract the policy domain file into resources/fixtures
[policy_domain.rdf.zip](https://github.com/ec-europa/joinup-dev/files/2026898/policy_domain.rdf.zip)
- Run the phing target 'import-rdf-fixtures'
./vendor/bin/phing import-rdf-fixtures

This should populate the policy domain taxonomy with the terms of Joinup.

cassm199:
Hi @sandervd 

Thanks for your prompt reply. I've tried your suggestion however to no avail. The Policy Domains terms were still not visible. I've outputted the results from the phing import-rdf-fixtures command and no errors were thrown.

`Buildfile: /var/www/joinup-dev/build.xml
   [import] Unable to find build file: /var/www/joinup-dev/build.local.xml... skipped
     [echo] Loading AWS additional tasks.
     [echo] Turn environment variable into a phing property
     [echo] Loading infra additional tasks.
 [property] Loading /var/www/joinup-dev/build.properties.local
 [property] Loading /var/www/joinup-dev/build.properties
 [property] Loading /var/www/joinup-dev/build.properties.dist
     [echo] Loading Behat task.
     [echo] Loading After Import RDF fixtures task.
[...]

PR #1202, opened 2018-05-28
4521: Exception on editing Policy Domain terms

claudiu-cristea:


composer.json
```diff
@@ -44,7 +44,7 @@
         "drupal/piwik": "~1.0",
         "drupal/piwik_reporting_api": "dev-1.x",
         "drupal/r4032login": "dev-1.x#391f10d69b9b8c466a2fddac2a6b99f1289e7781",
-        "drupal/rdf_entity": "^1.0.0-alpha5",
+        "drupal/rdf_entity": "dev-draft_for_taxonomies",
```
idimopoulos: Now that we are not using the 8.x-1.x version in our code, until we have a new alpha, we should not any more require another branch in composer, because even when the upstream patch is accepted, it will not get into the codebase. We will need to keep using the 1.0.0-alpha5 and use the PR as a patch.

claudiu-cristea:
Resolves #1196

PR #1195, opened 2018-05-22
ISAICP-4498: Export initial test view for the collection administrato…

pfrenssen:
…r report.

PR #1203, opened 2018-05-28
ISAICP-4498

pfrenssen:


pfrenssen:
An earlier implementation for this was created in #1195 but it was abandoned after it was decided to go with a custom controller rather than a view.

PR #1208, opened 2018-05-30
Isaicp 4499: Fix the owner type predicate

idimopoulos:


web/modules/custom/owner/owner.install
```diff
@@ -0,0 +1,23 @@
+<?php
+
+/**
+ * @file
+ * Install, uninstall, schema and requirements hook for the owner module.
+ */
+
+/**
+ * Implements hook_update_N().
```
brummbar: We should drop this like in the other hooks, so the update list doesn't contain this unneeded information.

web/modules/custom/owner/owner.install
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Fix the owner type predicate.
+ */
+function owner_update_8100() {
+  /** @var \Drupal\Core\Database\Connection $query */
+  $query = \Drupal::service('sparql_endpoint');
```
brummbar: I saw some other remarks where it was asked to use a better variable for $query, like $sparql connection or similar. Maybe we could uniform more?

web/modules/custom/owner/owner.install
```diff
@@ -0,0 +1,23 @@
...
+  $owner_predicate_query = <<<QUERY
+DELETE { GRAPH ?g { ?subject <http://joinup.eu/voc/owner-type> ?object } }
+INSERT { GRAPH ?g { ?subject <http://purl.org/dc/terms/type> ?object } }
+USING <http://joinup.eu/owner/published>
```
brummbar: I know that Joinup only adds the published graph, but I was thinking that if somebody has extended Joinup and added more graphs, our update hook wouldn't fix those. Should we retrieve the graphs from the system just to be sure? Just asking.

PR #1215, opened 2018-06-04
Isaicp 4392: Change the 'Cancel account' button label on the user cancel form

idimopoulos:


web/profiles/joinup/src/Form/UserCancelForm.php
```diff
@@ -65,6 +65,9 @@ public function buildForm(array $form, FormStateInterface $form_state) {
     $collections = $this->relationManager->getCollectionsWhereSoleOwner($this->entity);
 
     if (!empty($collections)) {
+      // No access to the 'Cancel' button should be given if the user is the
+      // sole owner of a collection.
+      $form['actions']['submit']['#access'] = FALSE;
```
pfrenssen: I agree with this but shouldn't we solve this differently? I would suggest to show a greyed out version of the button with a tooltip "This user cannot be deleted since it is the owner of a collection".

PR #1225, opened 2018-06-07
Isaicp 4366: Update the share/unshare form

idimopoulos:


pfrenssen:
Code looks good! No remarks, and yes I tried to find something :)

pfrenssen:
I have a case during manual testing where the Unshare link is not appearing in the contextual links menu of the shared tile. It turns out to be caused by a core bug: https://www.drupal.org/project/drupal/issues/2773591

I am not going to block this on that difficult problem, this looks good to go!

PR #1228, opened 2018-06-08
Isaicp 4393: Do not allow all owners of a group to be deleted.

idimopoulos:


tests/features/bootstrap/FeatureContext.php
```diff
@@ -39,6 +39,7 @@ class FeatureContext extends RawDrupalContext implements SnippetAcceptingContext
    * Define ASCII values for key presses.
    */
   const KEY_LEFT = 37;
+
```
pfrenssen: If it is important to have a blank line here then a new docblock should be added for the second line, and the docblock for the first line should be rewritten.

tests/features/bootstrap/FeatureContext.php
```diff
@@ -425,7 +426,10 @@ public function assertFieldOptionSelected($option, $select) {
    */
   public function assertFieldRadioSelected($radio, $field) {
     // Find the grouping fieldset that contains the radios field.
-    $fieldset = $this->getSession()->getPage()->find('named', ['fieldset', $field]);
+    $fieldset = $this->getSession()->getPage()->find('named', [
```
pfrenssen: Please revert this change, it makes it less readable.

tests/features/bootstrap/FeatureContext.php
```diff
@@ -615,7 +619,10 @@ public function assertContextualLinkInRegionPresent($text, $region) {
     $links = $this->findContextualLinkPaths($this->getRegion($region));
 
     if (!isset($links[$text])) {
-      throw new \Exception(t('Contextual link %link expected but not found in the region %region', ['%link' => $text, '%region' => $region]));
+      throw new \Exception(t('Contextual link %link expected but not found in the region %region', [
```
pfrenssen: Please revert this change, it makes it less readable.

tests/features/bootstrap/FeatureContext.php
```diff
@@ -636,7 +643,10 @@ public function assertContextualLinkInRegionNotPresent($text, $region) {
     $links = $this->findContextualLinkPaths($this->getRegion($region));
 
     if (isset($links[$text])) {
-      throw new \Exception(t('Unexpected contextual link %link found in the region %region', ['%link' => $text, '%region' => $region]));
+      throw new \Exception(t('Unexpected contextual link %link found in the region %region', [
```
pfrenssen: Please revert this change, it makes it less readable.

tests/features/bootstrap/FeatureContext.php
```diff
@@ -844,7 +854,10 @@ public function fillFieldWithValues($field, $values) {
     $values = $this->explodeCommaSeparatedStepArgument($values);
 
     /** @var \Behat\Mink\Element\NodeElement[] $items */
-    $items = $this->getSession()->getPage()->findAll('named', array('field', $field));
+    $items = $this->getSession()->getPage()->findAll('named', array(
```
pfrenssen: Please revert this change, it makes it less readable.

tests/features/bootstrap/FeatureContext.php
```diff
@@ -995,8 +1008,7 @@ public function assertTexts(TableNode $table) {
     foreach ($lines as $line) {
       try {
         $this->assertSession()->pageTextContains($line);
-      }
-      catch (ResponseTextException $e) {
+      } catch (ResponseTextException $e) {
```
pfrenssen: This is no longer adhering to coding standards.

tests/features/bootstrap/FeatureContext.php
```diff
@@ -1021,8 +1033,7 @@ public function assertNoTexts(TableNode $table) {
     foreach ($lines as $line) {
       try {
         $this->assertSession()->pageTextNotContains($line);
-      }
-      catch (ResponseTextException $e) {
+      } catch (ResponseTextException $e) {
```
pfrenssen: This is no longer adhering to coding standards.

tests/features/bootstrap/FeatureContext.php
```diff
@@ -1110,6 +1121,40 @@ public function assertPageNotCached() {
...
+
+  /**
+   * Attempts to uncheck a checkbox in a table row containing a given text.
+   *
+   * @param string $text
+   *   Text in the row.
+   *
+   * @Given I deselect/uncheck the :row_text row
+   */
+  public function assertDeselectRow($text) {
```
pfrenssen: Can I has type hints for the argument and return type?

tests/features/bootstrap/FeatureContext.php
```diff
@@ -1110,6 +1121,40 @@ public function assertPageNotCached() {
...
+   *   The checkbox element.
+   *
+   * @throws \Exception
+   *   If the page contains no rows, no row contains the text or the row
+   *   contains no checkbox.
+   *
+   * @return \Behat\Mink\Element\NodeElement|null
+   *   The checkbox element.
+   */
+  protected function getRowCheckboxByText($text) {
```
pfrenssen: Here also provide the type hints.

tests/features/bootstrap/FeatureContext.php
```diff
@@ -1110,6 +1121,40 @@ public function assertPageNotCached() {
...
+   *   Text in the row.
+   *
+   * @return \Behat\Mink\Element\NodeElement|null
+   *   The checkbox element.
+   *
+   * @throws \Exception
+   *   If the page contains no rows, no row contains the text or the row
+   *   contains no checkbox.
+   *
+   * @return \Behat\Mink\Element\NodeElement|null
```
pfrenssen: Since this method should be throwing exceptions when any of the required conditions are not met, it should not return `NULL` under any circumstances.

tests/features/joinup_core/owner_deletion.feature
```diff
@@ -0,0 +1,63 @@
+@api
+Feature: As a site owner
+  In order to always have an owner assigned to a group
```
pfrenssen: I think it would maybe be clearer if you say

> In order to avoid orphaned groups

because this doesn't explain why it is important that a group needs an owner

tests/features/joinup_core/owner_deletion.feature
```diff
@@ -0,0 +1,63 @@
+@api
```
pfrenssen: Wouldn't it be better to add this in the existing feature that is dealing with deleting owners (`owner.moderation.feature`), or at least in the same folder?

tests/features/joinup_core/owner_deletion.feature
```diff
@@ -0,0 +1,63 @@
...
+    Given I deselect the "Group owner 1" row
+    When I press "Apply to selected items"
+    Then I should see "Delete the selected membership(s) was applied to 1 item."
+
+    # Action is properly interrupted even if other memberships are about to be deleted.
+    Given I select the "Group owner 1" row
+    And I select the "Group member" row
+    And I select "Delete the selected membership(s)" from "Action"
+    When I press "Apply to selected items"
+    Then I should see the error message "You cannot delete all owners of the <type>."
```
pfrenssen: Can we improve this message in some way? In the standard use case of Joinup it is only possible to have 1 owner. We only have a very small number of collections with multiple owners, and this is only because they were migrated from the old platform like this.

So this means that in the 80% use case this message will be shown when the 1 single owner is deleted, and in that sense it is strange to see "You cannot delete all owners".

web/modules/custom/joinup_core/joinup_core.module
```diff
@@ -1051,4 +1051,73 @@ function joinup_core_form_views_form_og_members_overview_default_alter(&$form, F
...
+ *
+ * @param string $bulk_form_key
+ *   The bulk form key representing the entity's id, language.
+ *
+ * @see \Drupal\views\Plugin\views\field\BulkForm::loadEntityFromBulkFormKey
+ *
+ * @return string
+ *   The entity's id.
+ */
+function joinup_core_extract_bulk_form_entity_id($bulk_form_key) {
```
pfrenssen: Type hints are welcome.

tests/features/bootstrap/FeatureContext.php
idimopoulos: For this file, I did not notice that it was not scanned by phpcs. I will revert all non-functional changes. I guess I clicked an auto formatter and these were changed..

tests/features/joinup_core/owner_deletion.feature
idimopoulos: The 'owner' folder is about rdf entity type 'owner's. It is the corresponding entity to the contact_information. Not the actual user. It is a bit confusing..

tests/features/joinup_core/owner_deletion.feature
idimopoulos: Yeah, but do you have any alternative? We could have something like `You cannot proceed as that would leave the group orphaned` but that kinda involves terms that the user might be unaware of.

tests/features/bootstrap/FeatureContext.php
pfrenssen: I've noticed PHPStorm doing this even without pressing any buttons.

A good tip is to use `git add -p` when adding files to the staging area, this allows you do to a mini-review of the changes that are being added. It's really helpful for catching these auto-formatted lines.

pfrenssen:
It's looking good now. There were a few questions, I answered these and provided the small fixes for them myself. Let's wait for the tests.

pfrenssen:
This has caused 31 random failures in the last 3 days. Reverted.

PR #1229, opened 2018-06-08
ISAICP-4531: Provide a post update hook to force an update on the cached computed field timers

idimopoulos:


idimopoulos:
The approach is changed to a follow up, which includes a fix in the conifg_sync module.

PR #1235, opened 2018-06-12
4482: Provide field defaults for imported entities

claudiu-cristea:


claudiu-cristea:
Just a technical note regarding changing the field `collection` from the `solution bundle`: I had to do that because of tests. In kernel tests we don't enable the `solution.module` because its complex dependencies. Thus, in order to make `$solution->collection` work, I had to copy/paste large parts of the `solution.module` into the testing module. To avoid that, I preferred to fix the field to behave like a true computed field, simplifying the whole logic from `solution.module`.

PR #1237, opened 2018-06-13
4574: Provenance activity entities should not be indexed

claudiu-cristea:


web/modules/custom/joinup_search/joinup_search.behat.inc
```diff
@@ -55,23 +55,47 @@ class SearchSubContext extends DrupalSubContextBase implements DrupalSubContextI
...
+   *
+   * @param string $type
+   *   The text of the content tab.
+   *
+   * @throws \Exception
+   *   Thrown when the tab is not found in the page.
+   *
+   * @Given the :type content tab is not displayed
+   */
+  public function assertContentTabNotDisplayed(string $type) {
```
idimopoulos: I think this step is the same as ::assertContentTypeFacetItemsNotPresent of the same class (the other supports multiple labels) and is not needed. 
The same applies for it's usage on the steps.

idimopoulos: If you think we can keep it though for a single label check, I am fine with it.

claudiu-cristea: I was thinking that it's the pair of `assertContentTabDisplayed ()`. Usually we have the "exists" and "not exists" pair. I missed the other one. No strong opinion.

idimopoulos:
I will merge this request myself if green. The last commit is discussed with @claudiu-cristea. No further QA is needed.

PR #1238, opened 2018-06-13
DON'T  MERGE: Testing...

claudiu-cristea:


web/modules/custom/asset_distribution/asset_distribution.module
```diff
@@ -63,6 +64,20 @@ function asset_distribution_form_rdf_entity_asset_distribution_form_alter(&$form
...
+/**
+ * Submit handler for rdf_entity_asset_distribution_form.
+ *
+ * Redirects to the overview of the parent entity instead of the entity itself.
+ *
+ * @see asset_distribution_form_rdf_entity_asset_distribution_form_alter()
+ */
+function asset_distribution_form_rdf_entity_asset_distribution_form_submit(&$form, FormStateInterface $form_state) {
+  $parent = Rdf::load($form_state->get('context_rdf_entity_id'));
+  $form_state->setRedirectUrl($parent->toUrl());
```
idimopoulos: The main reason that the add_asset_distribution tests were failing here is the AddContentBlock.
The construction of the `Add <content/rdf type>` buttons is implemented by passing the route context as a parameter to construct the URLs.
By removing this submission handler, the `\Drupal\rdf_entity\Form\RdfForm::buildForm` was kicking in, redirecting to the distribution overview instead of the solution/release overview. 
On the distribution page, the solution is winning over the release parent because the primary actions concern the solution. This means, that while being on the overview of the distribution, the `Add distribution` menu item, concerns the group, not the parent of the current overview, thus, failing on the title constraint, as there *can* be a distribution with the same name as with one of the distributions of a release.

tests/features/asset_distribution/add_asset_distribution.feature
```diff
@@ -198,5 +198,6 @@ Feature: Add distribution through the UI
       | Quality Yard | Boost Software License | Solution random x name |
 
     When I go to the homepage of the "Solution random x name" solution
-    Then I should not see the text "WTFPL"
-    And I should not see the text "Boost Software License"
+    # Distributions are still shown in the solution page and so the License text will be visible in the page.
```
idimopoulos: I tested this manually by adding a standalone distribution on the solution and I also queried the production server to verify that this is the current status quo of this functionality.
To query the production server, on production's SPARQL endpoint, you can run the following query to get a list of solutions that have at least one standalone distribution:
```
select DISTINCT(?entity)
where {
  ?entity a <http://www.w3.org/ns/adms#Asset> .
  ?entity <http://www.w3.org/ns/dcat#distribution> ?distr .
  FILTER NOT EXISTS { ?entity <http://purl.org/dc/terms/isVersionOf> ?parent }
}
```
Adding a standalone distribution through the UI also makes this distribution appear in the overview page.
I can assume that the previous implementation of the `solution` parameter was either not clearing the cache or was not creating all appropriate connections (like having the solution also as parent), thus, making the distributions not available on the overview. 

I updated this test to check for the text to the header since this is a regression test after deciding to remove the licenses from the header region of the solution (previously they were shown in the little box where the download count also appears).

web/modules/custom/asset_distribution/asset_distribution.behat.inc
```diff
@@ -167,37 +165,17 @@ class AssetDistributionSubContext extends DrupalSubContextBase implements Drupal
...
-        $parent_entity = self::getRdfEntityByLabel($value);
-
-        // If we know the parent then we can transparently populate the group
-        // reference to the solution.
-        if (!array_key_exists('solution', $aliased_values)) {
-          $solution_label = NULL;
-          if ($parent_entity->bundle() === 'solution') {
-            $solution_label = $parent_entity->label();
-          }
-          elseif ($solution_entity_id = $parent_entity->field_isr_is_version_of->value) {
```
idimopoulos: I found the core of the relation issues. It is the `->value` part of this line. It will never retrieve the value as the `field_isr_is_version_of` is an entity reference, and the magic method to retrieve the value is `->target_id`... We have been having this in our code for a while. It has given us a false positive in some tests. I think it is a very good thing to have finally gotten rid of it. 
Your relations work fine up to now.

web/modules/custom/asset_release/src/Controller/AssetReleaseController.php
```diff
@@ -122,29 +121,7 @@ public function overview(RdfInterface $rdf_entity) {
...
-    foreach ($releases as $release) {
-      foreach ($release->get('field_isr_distribution')->getValue() as $distribution_reference) {
-        $release_distribution_ids[] = $distribution_reference['target_id'];
-      }
-    }
-
-    $query = $this->queryFactory->get('rdf_entity');
-    $query
-      ->condition('rid', 'asset_distribution')
-      ->condition(OgGroupAudienceHelperInterface::DEFAULT_FIELD, $rdf_entity->id());
```
idimopoulos: Please, check https://github.com/ec-europa/joinup-dev/pull/1238/files#diff-c379272bb9dad959c9312d24c58f63f2L182 for the reason why this line was not working properly but was appearing to work properly on our tests.

I am pretty sure that manual check will not work when attempting to test this though I don't want to spend the time to check it.

The steps to reproduce would be to:
* Create a solution.
* Create a release of that solution on a draft state.
* Create a distribution of that release.
* Log out, because a facilitator/moderator has view access to the draft releases.
* Go to the "Download releases" page.

*Expected behaviour:* The distribution is hidden due to the parent release being hidden.
*Actual behaviour:* The distribution is displayed as a standalone distribution.
*Reason:* If you pay close attention to the code above, the releases are loaded but removes from the list of the `$releases` variable all those that the user does not have view access. Then, standalone distributions are attempted to be retrieved by loading all distributions belonging to the solution group apart from those that belong to a release. The later list is built excluding distributions belonging to the `$releases` variable above which though now does not include the draft releases. That means that all distributions belonging to draft releases are loaded normally and marked as standalone. This was the 3rd failure of this PR.

PR #1240, opened 2018-06-14
Isaicp 4579: Extend the adms test files to cover as most cases as possible

idimopoulos:


idimopoulos:
Implementing this ticket caused the main branch to fail since the full file brought up some issues with the importer that were there from before. As this ticket is implemented after though, it is blocked on the original pipeline until it works with it since the tests should be green for this ticket to be merged in.

PR #1244, opened 2018-06-18
Clarify the method that retrieves the parent's eLibrary creation option

pfrenssen:
I encountered this method during my work on ISAICP-4498 and it confused my brains. The documentation did not help to unconfuse. So let's fix.


pfrenssen:
Thanks! I fixed and rebased so it is only 1 commit for this small fix.

PR #1246, opened 2018-06-19
4576: Imported distributions are not connected to their parent

claudiu-cristea:


web/modules/custom/asset_distribution/src/DistributionParentFieldItemList.php
```diff
@@ -0,0 +1,92 @@
...
+   * {@inheritdoc}
+   */
+  public function postSave($update): bool {
+    // Set the parent only for new distributions.
+    /** @var \Drupal\rdf_entity\RdfInterface $parent */
+    if (!$update && !$this->isEmpty() && ($parent = $this->list[0]->entity)) {
+      $parent->skip_notification = TRUE;
+      $field_name = $parent->bundle() === 'solution' ? 'field_is_distribution' : 'field_isr_distribution';
+      $parent->get($field_name)->appendItem(['target_id' => $this->getEntity()->id()]);
+      $parent->save();
```
idimopoulos: There is an issue that I run against during building the import demo
In the above postSave, the `$parent = $this->list[0]->entity` will attempt to retrieve the parent entity and assign the distribution and then save.
However, at that point, since the import sequence is not necessarily queued (e.g. first import all solutions, then all distributions etc) but we are importing as we go - and I am not sure if we can queue it since we have many circular references, the solution might be loaded through the staging graph, then be attempted to be saved with the new reference, then `solution_rdf_entity_presave` kicks in and attempts to find a state, but since the entity is not not imported yet, no state is found and an exception will be thrown.

Now there are multiple things going on here:
* First, while using an entity graph, the entity loaded at this point is not _ready_ to be imported since default values have not been assigned. However, since the entity exists in the staging graph only, it _could_ be considered as validated and enforce the state as such.
* Second, solutions that come from the staging graph, are supposed to already have the connection set as they come from an rdf file. So the solution loaded at this point, should already have the reference set. Under this idea, we could 
** exclude updating the parent if it comes from the staging graph or the reference already exists. This is the approach I took in the demo that you can find at https://github.com/ec-europa/joinup-dev/pull/1243/commits/9873acff652128dd6d16f07fcdacef9b42baaa64#diff-9cce0801a759303c3bb2378a7e759020R60 .
The problem with my approach is that it handles a _staging_ graph explicitly and locks it as a graph that solutions are not uploaded. 
** not update the parent entity if the reference is already set. Though this would seem the most _proper_ approach at this point, there is no check in our validator that a distribution is referenced by the parent that we have set as a parent, thus, if the rdf file is wrong, the exception will be thrown again. So, I did not choose this solution because in case of an error, the procedure would crash instead of failing silently.

In conclusion, it is indeed very hard to handle all circular references generically but this needs to at least temporarily be handled as it will block a proper import.
Maybe a careful queueing of the to-be-imported entities could save us from this trouble but we have to counter this anyway.

web/modules/custom/asset_distribution/src/DistributionParentFieldItemList.php
```diff
@@ -0,0 +1,92 @@
...
+    }
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function postSave($update): bool {
+    // Set the parent only for new distributions.
+    /** @var \Drupal\rdf_entity\RdfInterface $parent */
+    if (!$update && !$this->isEmpty() && ($parent = $this->list[0]->entity)) {
```
claudiu-cristea: Probably this line should be:

```php
if (!$update && !empty($this->list) && ($parent = $this->list[0]->entity)) {
```

PR #1253, opened 2018-06-20
Hotfix: Isaicp 4589: Allow facilitators to receive a notification as well on a membership request

idimopoulos:


web/modules/custom/joinup_notification/src/EventSubscriber/OgMembershipSubscriber.php
```diff
@@ -66,7 +66,7 @@ public function onCreate(NotificationEvent $event) {
         'rdf_entity-collection-administrator' => [
           $template_id,
         ],
-        'rdf_entity-solution-administrator' => [
+        'rdf_entity-collection-facilitator' => [
```
pfrenssen: This looks like it is removing the possibility of sending the notifications to solution administrators. I'm not sure if this will affect anything. It is not possible for users to request membership of a solution, but facilitators can add new members at will in order to expand the facilitator team.

Let's see what the tests say.

idimopoulos:
```
$user_data_array = [
      'og_roles' => [
        'rdf_entity-collection-administrator' => [
          $template_id,
        ],
        'rdf_entity-solution-administrator' => [
          $template_id,
        ],
      ],
    ];
```

However, there are no membership requests for solutions so the second index is probably wrong and should be `rdf_entity-collection-facilitator`, thus, making this ticket indeed a bug. There was no spamming question for this issue.

PR #1263, opened 2018-06-22
ISAICP-4474: Enrich some message templates with a disclaimer.

idimopoulos:


tests/features/collection/notification.collection.feature
```diff
@@ -219,6 +253,3 @@ Feature: Notification test for the collection transitions.
       | recipient | NC member2                                                                  |
       | subject   | The collection NC to delete was deleted.                                    |
       | body      | The collection "NC to delete", of which you are a member, has been deleted. |
-
```
brummbar: How come this was removed from here and moved on top? Just out of curiosity.

tests/features/asset_release/asset_release.moderation.feature
```diff
@@ -97,9 +97,12 @@ Feature: Asset release moderation
     And I should see the following warning messages:
       | You are viewing the published version. To view the latest draft version, click here. |
     And the following email should have been sent:
```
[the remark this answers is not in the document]
idimopoulos: Indeed. I have removed all of the instances where only the recipient and subject are checked. Thanks for noticing :)

tests/features/collection/notification.collection.feature
idimopoulos: This does not actually make any difference when the tests are green. However, when I was changing the emails, I had to re run the tests multiple times. After fixing the email during the manual proposal of the collection, other cases were breaking. However, the step is not an after scenario step so it was never being run. Then I had to clean the collection manually to re run it.

Like this, the collection is cleaned up right after the relevant lines are ran

brummbar: Oh yes perfect, it's actually an improvement then. 👍

PR #1274, opened 2018-07-03
Isaicp 4393: Try again to ensure that not all owners of a group can be deleted.

idimopoulos:


tests/features/bootstrap/FeatureContext.php
```diff
@@ -1109,7 +1110,41 @@ public function assertPageNotCached() {
    *
    * @Given I select/check the :row_text row
    */
-  public function assertSelectRow($text) {
+  public function assertSelectRow(string $text): void {
```
pfrenssen: If we are touching this then we should take the opportunity to update the method name. Methods that are named `assertSomething()` are intended for checking a certain condition. They should throw an exception if the condition is not met.

I know this is also checking something, but there is a difference between checking a condition and checking a checkbox :)

tests/features/bootstrap/FeatureContext.php
```diff
@@ -1109,7 +1110,41 @@ public function assertPageNotCached() {
...
+
+  /**
+   * Attempts to uncheck a checkbox in a table row containing a given text.
+   *
+   * @param string $text
+   *   Text in the row.
+   *
+   * @Given I deselect/uncheck the :row_text row
+   */
+  public function assertDeselectRow(string $text): void {
```
pfrenssen: This is not an assertion, so let's rename the method to something that makes more sense like `deselectNamedRow()`.

tests/features/joinup_core/owner_deletion.feature
```diff
@@ -0,0 +1,56 @@
...
+    And I am logged in as "Site moderator"
+    And I go to the homepage of the "An owned group" <type>
+    And I click "Members"
+    And I select the "Group owner 1" row
+    And I select the "Group owner 2" row
+
+    # Try to delete all owners at once.
+    And I select "Delete the selected membership(s)" from "Action"
+    When I press "Apply to selected items"
+    Then I should see the error message "You cannot delete the owner of a <type>."
```
pfrenssen: A random failure had been affecting this part of the test.

PR #1277, opened 2018-07-04
4623: Deal with imported data broken references

claudiu-cristea:
This should be reworked to run in batch, if #1275 goes first.

### DIFF
--- a/tests/fixtures/files/valid_adms.rdf
+++ b/tests/fixtures/files/valid_adms.rdf
@@ -48,7 +48,11 @@
         <dc:hasVersion rdf:resource="http://example.com/release/1"/>
         <dc:hasVersion rdf:resource="http://example.com/release/2"/>
         <dct:relation rdf:resource="http://example.com/solution/3"/>
-        <dct:type rdf:resource="http://vocabulary/term"/>
+        <dct:relation rdf:resource="http://example.com/solution/2"/>
+        <dct:type rdf:resource="http://vocabulary/term/1"/>
+        <dct:language rdf:resource="http://example.com/language/invalid/1"/>
+        <dct:language rdf:resource="http://example.com/language/invalid/2"/>
+        <dct:language rdf:resource="http://publications.europa.eu/resource/authority/language/RON"/>
     </rdf:Description>
 
     <!-- A solution with only a standalone distribution. -->
@@ -184,7 +188,7 @@
     </rdf:Description>
 
     <!-- Custom term of a custom vocabulary to fill in the solution type. -->
-    <rdf:Description rdf:about="http://vocabulary/term">
+    <rdf:Description rdf:about="http://vocabulary/term/1">
         <rdf:type rdf:resource="http://www.w3.org/2004/02/skos/core#Concept"/>
         <skos:label>Term</skos:label>
         <skos:inScheme rdf:resource="http://vocabulary"/>
--- a/tests/fixtures/files/valid_adms.rdf
+++ b/tests/fixtures/files/valid_adms.rdf
@@ -48,6 +48,7 @@
         <dc:hasVersion rdf:resource="http://example.com/release/1"/>
         <dc:hasVersion rdf:resource="http://example.com/release/2"/>
         <dct:relation rdf:resource="http://example.com/solution/3"/>
+        <dct:type rdf:resource="http://vocabulary/term"/>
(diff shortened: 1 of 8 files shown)


web/modules/custom/joinup_federation/src/Plugin/pipeline/Step/BrokenReferences.php
```diff
@@ -0,0 +1,177 @@
...
+      $changed = FALSE;
+      $reference_fields = $this->getAdmsSchemaEntityReferenceFields($entity->bundle(), ['rdf_entity', 'taxonomy_term']);
+
+      foreach ($reference_fields as $field_name => $target_entity_type_id) {
+        /** @var \Drupal\Core\Field\EntityReferenceFieldItemListInterface $field */
+        $field = $entity->get($field_name);
+
+        // Remove references to entities that were blacklisted by the user.
+        if ($blacklist && ($target_entity_type_id === 'rdf_entity')) {
+          $changed |= $this->removeBlacklistedReferences($field, $blacklist);
```
idimopoulos: I actually had to look this up. Is there a reason we are asking for an integer to be returned here? In any case, the `$changed` variable is initialized as `FALSE`. Either initialize it as `0` or use plain `=` signs in this line.

web/modules/custom/joinup_federation/src/Plugin/pipeline/Step/BrokenReferences.php
```diff
@@ -0,0 +1,177 @@
...
+    if (!$field->isEmpty()) {
+      $deltas_to_remove = [];
+      foreach ($field as $delta => $field_item) {
+        if (isset($blacklist[$field_item->target_id])) {
+          $deltas_to_remove[] = $delta;
+        }
+      }
+      if ($deltas_to_remove) {
+        foreach ($deltas_to_remove as $delta_to_remove) {
+          $field->removeItem($delta_to_remove);
```
idimopoulos: This is not going to work properly if more than one deltas are found for removal. That is because the deltas array is build above and then they are removed here. But if, lets say and array of 2 items exists (index 0 and index 1), and both of them are to be removed, then the removeItem method will first remove index 0, but then only index 0 will remain as the other items are shifted.
I noticed it breaks the functionality when more items are to be removed.

web/modules/custom/joinup_federation/src/Plugin/pipeline/Step/BrokenReferences.php
```diff
@@ -0,0 +1,177 @@
...
+      $referenced_terms = $field->referencedEntities();
+      $deltas_to_remove = [];
+      foreach ($field as $delta => $field_item) {
+        if (!isset($referenced_terms[$delta])) {
+          $deltas_to_remove[] = $delta;
+        }
+      }
+      if ($deltas_to_remove) {
+        foreach ($deltas_to_remove as $delta_to_remove) {
+          $field->removeItem($delta_to_remove);
```
idimopoulos: Same as above.

tests/fixtures/files/valid_adms.rdf
```diff
@@ -47,6 +47,8 @@
         <dcat:landingPage rdf:resource="http://example.com"/>
         <dc:hasVersion rdf:resource="http://example.com/release/1"/>
         <dc:hasVersion rdf:resource="http://example.com/release/2"/>
+        <dct:relation rdf:resource="http://example.com/solution/3"/>
```
idimopoulos: Please, provide more than invalid entries for each field that can have invalid entries.

claudiu-cristea: @idimopoulos for each taxonomy term reference field?

web/modules/custom/joinup_federation/src/Plugin/pipeline/Step/BrokenReferences.php
claudiu-cristea: We don't have the same operator for booleans, something that should do `$changed = $changed || ....`.

web/modules/custom/joinup_federation/src/Plugin/pipeline/Step/BrokenReferences.php
claudiu-cristea: This is a good catch. I tested only with one item

PR #1288, opened 2018-07-12
Isaicp 4637: Provide a solution for the virtuoso issue.

idimopoulos:


web/modules/custom/joinup_sparql/src/Database/Driver/sparql/Connection.php
```diff
@@ -0,0 +1,97 @@
...
+/**
+ * @addtogroup database
+ * @{
+ */
+class Connection extends BaseConnection implements ConnectionInterface {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function query(string $query): Result {
```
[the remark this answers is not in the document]
claudiu-cristea: >  Are you ok with that (counting the 5 seconds in the performance)?

I think that reflects the truth. But I don't have a strong opinion. The solution I proposed is not against adding the
```
if (!empty($this->logger)) {
  $this->logger->clear('webprofiler');
}
```
block of code. I think it could be added just before `sleep()`, if you think. I'll let you decide :)

web/modules/custom/joinup_sparql/src/Database/Driver/sparql/Connection.php
```diff
@@ -18,76 +19,34 @@ class Connection extends BaseConnection implements ConnectionInterface {
...
-        throw new SparqlQueryException('Execution of query failed: ' . $query);
-      }
-      if (!empty($this->logger)) {
-        $this->logger->clear('webprofiler');
-      }
+    catch (SparqlQueryException $e) {
+      // During a Virtuoso checkpoint, the server locks down, causing HTTP
+      // requests on the SPARQL endpoint to fail with a 404 response. We wait a
+      // reasonable amount of time and then we retry one more time.
+      // @see: http://docs.openlinksw.com/virtuoso/checkpoint/
```
claudiu-cristea: No ":" after `@see`.

web/modules/custom/joinup_sparql/src/Database/Driver/sparql/Connection.php
```diff
@@ -18,76 +19,34 @@ class Connection extends BaseConnection implements ConnectionInterface {
...
```
claudiu-cristea: Same ^^^

PR #1289, opened 2018-07-12
4558: Cannot redirect from a private file path

claudiu-cristea:
Upstream patch https://www.drupal.org/project/redirect/issues/2985288 should be reviewed too.


claudiu-cristea:
Upstream patch https://www.drupal.org/project/redirect/issues/2985288 should be reviewed too.

PR #1291, opened 2018-07-13
4607: Tallinn dashboard endpoint

claudiu-cristea:


web/modules/custom/embed_block/src/Plugin/Filter/EmbedBlockFilter.php
```diff
@@ -0,0 +1,129 @@
...
+            $block_content = $this->renderer->render($build);
+          }
+          // If the user cannot access the block, means that the block exists
+          // but should not be rendered, so we still have to replace the
+          // placeholder with an empty string.
+          else {
+            $block_content = '';
+          }
+          // Cache metadata applies regardless if the user can access the block.
+          $response
```
claudiu-cristea: This statement (4 lines) could be replaced by: `$response->addCacheableDependency($block_plugin);`. I figured out late :)

web/modules/custom/joinup_communities/tallinn/src/Controller/DashboardController.php
```diff
@@ -0,0 +1,207 @@
...
+   */
+  public function __construct(EntityFieldManagerInterface $entity_field_manager, DashboardAccessInterface $dashboard_access) {
+    $this->entityFieldManager = $entity_field_manager;
+    $this->dashboardAccess = $dashboard_access;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function create(ContainerInterface $container) {
```
pfrenssen: The return type of the factory method can be specified as `: self`.

web/modules/custom/joinup_communities/tallinn/src/Controller/DashboardController.php
```diff
@@ -0,0 +1,207 @@
...
+   * @param \Symfony\Component\HttpFoundation\Request $request
+   *   The HTTP request Symfony instance.
+   *
+   * @return \Drupal\Core\Cache\CacheableResponseInterface
+   *   The cached Json response.
+   *
+   * @throws \LogicException
+   *   If the group entity is missing.
+   */
+  public function getData(Request $request): CacheableResponseInterface {
```
pfrenssen: This `$request` argument is not used.

web/modules/custom/joinup_communities/tallinn/src/Form/DashboardPermissionsForm.php
```diff
@@ -0,0 +1,115 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function submitForm(array &$form, FormStateInterface $form_state): void {
+    $this->toggleConfigReadonly(FALSE);
+    $this->configFactory->getEditable('tallinn.settings')
+      ->set('dashboard.access_type', $form_state->getValue('access_type'))
+      ->save();
+    $this->toggleConfigReadonly(TRUE);
```
pfrenssen: I think this is an inventive solution to give some real time control about the site configuration to moderators, but unfortunately this is not sufficient. On the next deployment the original config as defined in `tallinn.settings.yml` will be reapplied and the access will be restricted again.

If this would be something that needs to be toggled repeatedly and with great urgency then we could come up with a way to exclude certain config keys from being overwritten during deploy. For example we could use the config_filter module to exclude the `tallinn.settings.dashboard.access_type` key.

But I think this is overkill, this will only be toggled one single time, and we know in advance when this will happen so we can schedule a release on that day.

So maybe this whole form is unneeded?

web/modules/custom/joinup_communities/tallinn/src/Plugin/Block/DashboardBlock.php
```diff
@@ -0,0 +1,319 @@
...
+    return $this->dashboardAccess->access($account);
+  }
+
+  /**
+   * Builds the block render array.
+   *
+   * @return array
+   *   Render array
+   */
+  protected function buildRenderArray(): array {
```
pfrenssen: There is a lot of very intricate building of HTML in this block, but I can't see it when I am running the tests. There is also no test that checks the content of this block as far as I can see. Is this out of scope for the ticket?

web/modules/custom/joinup_communities/tallinn/tallinn.post_update.php
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * Install the 'embed_block' module and add the dashboard block in page content.
+ */
+function tallinn_post_update_add_block() {
+  \Drupal::service('module_installer')->install(['embed_block']);
+
+  // The 'Implementation monitoring' page node ID equals 701254.
+  /** @var \Drupal\node\NodeInterface $custom_page */
+  $custom_page = Node::load(701254);
```
pfrenssen: Can't we use a UUID for this? That is guaranteed to be unique, and then we can also create this custom page on new installs.

Currently when doing a clean install, or when running the Behat tests, this page is not available. This explains why I cannot see the block containing the report in the Tallinn collection.

tests/features/communities/tallinn/dashboard.feature
```diff
@@ -0,0 +1,96 @@
...
+      | error messages                                     |
+      | Access denied. You must sign in to view this page. |
+    And I go to "/admin/config/content/tallinn"
+    Then I should see the following error message:
+      | error messages                                     |
+      | Access denied. You must sign in to view this page. |
+
+    Given I am logged in as Dinesh
+    When I go to "/tallinn-dashboard"
+    Then the response status code should be 403
```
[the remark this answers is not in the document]
claudiu-cristea: Yes, I would prefer to keep it here as it make more sense to this functionality.

web/modules/custom/joinup_communities/tallinn/src/Plugin/Block/DashboardBlock.php
claudiu-cristea: This HTML markup is coming from the INNO team, see https://codepen.io/weslito/pen/ajObKq?editors=0010. I only abstracted it as a Drupal render array.

web/modules/custom/joinup_communities/tallinn/src/Form/DashboardPermissionsForm.php
claudiu-cristea: Finally, I brought on the table an old idea on how to trick config readonly. I switched to a state entry. Yes, I know, smells a lot as `variable_[get|set]()` :)

web/modules/custom/joinup_communities/tallinn/tallinn.post_update.php
claudiu-cristea: I removed the content creation on module install. This was wrong decision on my side. @idimopoulos was right that we should provide the content in test. The solution now is:
* For automated tests (Behat) the content is provided by tagging with `@tallinn`.
* For manual testing, you should simply enable the `demo_content` module. I added some sample data there that works well for manual testing.

tests/features/communities/tallinn/dashboard.feature
```diff
@@ -0,0 +1,96 @@
...
+    And I go to "/admin/config/content/tallinn"
+    Then the response status code should be 403
+
+    Given I am logged in as Gilfoyle
+    When I go to "/tallinn-dashboard"
+    Then the response status code should be 200
```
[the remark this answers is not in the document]
claudiu-cristea: Related to:

> I'm here getting an error The Tallinn collection entity is missing when I am testing manually.

I removed the content creation on module install. This was wrong decision on my side. @idimopoulos was right that we should provide the content in test, not on install. The solution now consist in:

* For automated tests (Behat) the content is provided by tagging with `@tallinn`.
* For manual testing, you should simply enable the `demo_content` module. I added some sample data there that works well for manual testing.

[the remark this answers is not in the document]
pfrenssen: OK putting it in the `demo_content` module is a good idea. This is a special case though, the functionality literally depends on the content being there, this is not user generated. So there should be a way to have this present without running any tests and pausing them half way through the execution.

I'll mention in the `README` for the module that the `demo_content` module should be enabled for this to work.

web/modules/custom/joinup_communities/tallinn/src/Form/DashboardPermissionsForm.php
pfrenssen: I had a discussion with Marco and the functional team about this and they explained that they do want this as a toggle in the UI since they don't want to risk the publication of the dashboard being delayed because of a potential deployment issue.

This is a good temporary solution but we should switch to config anyway once the dashboard is public, because state is intended to be temporary. I created a followup to deal with this in September: https://webgate.ec.europa.eu/CITnet/jira/browse/ISAICP-4662

pfrenssen:
There are two commits that were later reverted, let's rebase the branch so these are no longer there. There's no good reason to revert things that were not yet merged into develop or master.

pfrenssen:
Rebased and force pushed.

pfrenssen:
I finished the code review, thanks @claudiu-cristea for using very logical and clean commits, this made reviewing a breeze! I will now do another in-depth manual test.

PR #1295, opened 2018-07-19
Isaicp 4636: Remove default links from the table of contents.

idimopoulos:


idimopoulos:
The ticket will be tested in https://github.com/ec-europa/joinup-dev/pull/1286 and the scope of this PR is included now in the other one.

PR #1299, opened 2018-07-20
Isaicp 4541: Fix the members counter.

idimopoulos:


web/modules/custom/collection/collection.module
```diff
@@ -415,12 +415,15 @@ function collection_rdf_entity_view_alter(array &$build, EntityInterface $entity
  *   Amount of members.
  */
 function collection_member_count(RdfInterface $entity) {
-  return \Drupal::entityQuery('og_membership')
+  $query = \Drupal::database()->select('og_membership', 'om');
+  $query->leftJoin('users_field_data', 'ufd', 'om.uid = ufd.uid');
```
brummbar: Shouldn't we use an `innerJoin` here? If there is no user, the count should be gone too.

web/modules/custom/collection/collection.module
```diff
@@ -415,12 +415,15 @@ function collection_rdf_entity_view_alter(array &$build, EntityInterface $entity
...
+  $query->fields('om', ['uid'])
+    ->condition('ufd.status', 1)
     ->condition('entity_type', 'rdf_entity')
     ->condition('entity_id', $entity->id())
-    ->condition('state', OgMembershipInterface::STATE_ACTIVE)
-    ->count()
-    ->execute();
+    ->condition('state', OgMembershipInterface::STATE_ACTIVE);
+  $results = $query->execute();
+  return empty($results) ? 0 : count($results->fetchAllAssoc('uid'));
```
brummbar: I think we can use `->countQuery()`?

brummbar:
Can you also add a test to see that, when a user is approved, the counter changes?

tests/features/collection/homepage.feature
```diff
@@ -6,11 +6,11 @@ Feature: Collection homepage
 
   Background:
     Given users:
-      | Username |
-      | Frodo    |
-      | Boromir  |
-      | Legoloas |
-      | Gimli    |
+      | Username | Status |
+      | Frodo    | 1      |
```
pfrenssen: I would love to see human readable values here, like `active` and `blocked` instead of these weird numbers. The product owner and functional contractors don't know what it means if a status is set to `1`.

brummbar: Yes I looked into this and saw that we didn't do this yet in any test that uses the Status column. I didn't want to reject for this, but we can take the chance now that we have to fix some other remarks.

brummbar: Yes it got into my mind twice but I didn't want to reject only for this, as I saw that sadly we used this syntax in other places too. But since now the ticket needs more work, let's take the chance to fix it.

web/modules/custom/collection/collection.module
```diff
@@ -362,6 +362,7 @@ function collection_rdf_entity_view(array &$build, EntityInterface $entity, Enti
     $tag = $entity->getEntityTypeId() . ':' . $entity->id();
     // @see collection_og_membership_presave()
     $tags = Cache::buildTags('joinup-membership', [$tag]);
+    $tags = Cache::mergeTags($tags, ['user_list', 'og_membership_list']);
```
brummbar: This causes this cache entry to be invalidated whenever a membership, in any group, is updated. We need to narrow down the invalidation. What is really missing is the invalidation of the cache tag `joinup-membership:group_type:group_id` in a `hook_og_membership_delete()` hook, the same approach done in `collection_og_membership_presave()`.

tests/features/collection/homepage.feature
brummbar: As @pfrenssen suggested, let's take the chance now to replace 1 with `active` and 0 with `blocked` to give business meaning to those numbers.

tests/features/collection/homepage.feature
```diff
@@ -48,15 +48,8 @@ Feature: Collection homepage
...
-    # Test caching of the metrics: Members.
-    # Gimli is not a member yet.
-    When I am logged in as Gimli
-    And I go to the homepage of the "Middle earth daily" collection
-    And I press the "Join this collection" button
-    And I go to the homepage of the "Middle earth daily" collection
-    Then I see the text "4 Members" in the "Header" region
 
-    # see ISAICP-3599
+    # @see ISAICP-3599
```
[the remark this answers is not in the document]
idimopoulos: Then `@@@@@@@@@@SEEEEEEEEE` is also valid??? :D

web/modules/custom/collection/collection.module
idimopoulos: It is not only on delete. When a membership is approved it should also be invalidated.

web/modules/custom/joinup_user/joinup_user.behat.inc
```diff
@@ -121,4 +124,25 @@ class JoinupUserSubContext extends DrupalSubContextBase implements DrupalSubCont
...
+   * @param \Drupal\DrupalExtension\Hook\Scope\BeforeUserCreateScope $scope
+   *   An object containing the entity properties and fields that are to be used
+   *   for creating the user as properties on the object.
+   *
+   * @throws \Exception
+   *   Thrown when some of the values are not one of the expected ones.
+   *
+   * @BeforeUserCreate
+   */
+  public static function massageFieldsBeforeUserCreate(BeforeUserCreateScope $scope) {
```
brummbar: It took me a while but I found where we convert the other field labels for the user.
It's in `\Drupal\joinup\Traits\UserTrait::createUser()`, which is invoked from `\Drupal\joinup\Context\DrupalContext::createUsers()`. We can move this there so all the related code is together.

web/modules/custom/collection/collection.module
```diff
@@ -563,8 +564,14 @@ function collection_search_api_solr_query_alter(Query $solarium_query, QueryInte
...
 function collection_og_membership_presave(OgMembershipInterface $entity) {
-  $tags = [$entity->getGroupEntityType() . ':' . $entity->getGroupId()];
-  Cache::invalidateTags(Cache::buildTags('joinup-membership', $tags));
+  joinup_core_og_membership_invalidate_group_tags($entity);
+}
+
+/**
+ * Implements hook_ENTITY_TYPE_predelete().
+ */
+function collection_og_membership_predelete(OgMembershipInterface $entity) {
```
brummbar: Since these functions are about og_membership in general, should we move it together with the `joinup_core_user_update()` function that you just wrote? I didn't think about this before, your new code exposed it! 😝

idimopoulos: I was thinking of it as well and I was struggling a bit. The reason it is in collections is that it is collections that have the leave button and the members count. However, I think that it is a good option as well to keep everything cache related gathered in joinup_core.

web/modules/custom/joinup_user/joinup_user.behat.inc
idimopoulos: I am reverting actually. We already have tests failing because of this.

brummbar: Yes we should refactor the whole thing as we shouldn't put traits in traits. Ok for now.

PR #1303, opened 2018-07-26
Isaicp 4611: Implement TOC outline

idimopoulos:


web/modules/custom/custom_page/src/Plugin/Block/TableOfContentsOutline.php
```diff
@@ -0,0 +1,368 @@
...
+  /**
+   * Returns the menu tree of the current og menu.
+   *
+   * @return \Drupal\Core\Menu\MenuLinkTreeElement[]
+   *   The og menu tree.
+   */
+  protected function getMenuTree(): array {
+    if (empty($this->menuTree)) {
+      $og_menu_id = $this->getOgMenuName();
+      $this->menuTree = $this->menuLinkTree->load($og_menu_id, new MenuTreeParameters());
```
brummbar: I have added a test that shows that when a menu item links to an unpublished page, it still shows in the menu. We need to run access checks on this.
We also can use the `MenuTreeParameters` to load only the active items. For both you can look at `web/core/modules/system/src/Controller/SystemController.php:106`.
Added a test also for this.

web/modules/custom/custom_page/src/Plugin/Block/TableOfContentsOutline.php
```diff
@@ -0,0 +1,368 @@
...
+      $og_menu_id = $this->getOgMenuName();
+      $tree = $this->menuLinkTree->load($og_menu_id, new MenuTreeParameters());
+      $this->flatOutlineTree($tree);
+
+      foreach ($this->flattenedMenu as $menu_link_id => $menu_item) {
+        $title = $menu_item->getTitle();
+        /** @var \Drupal\Core\Url $url */
+        $route_name = $menu_item->getRouteName();
+
+        if ($title == 'Overview' && $route_name === 'entity.rdf_entity.canonical') {
```
brummbar: Maybe we can skip loading these items at all in the `getMenuTree()` method using the tree parameters?
Or at least avoid to have these 3 strings in 3 different places.

web/modules/custom/custom_page/src/Plugin/Block/TableOfContentsOutline.php
```diff
@@ -0,0 +1,368 @@
...
+      ];
+    }
+
+    $build = [
+      '#theme' => 'links',
+      '#links' => $links,
+    ];
+
+    $og_menu_id = $this->getOgMenuName();
+    $build['#cache']['tags'][] = 'config:system.menu.ogmenu-' . $og_menu_id;
```
brummbar: Dunno why, but if I change the menu structure the cache is not invalidated. Added a test for it.

web/themes/joinup/templates/block/block--toc-outline.twig
```diff
@@ -0,0 +1,31 @@
+{#
+/**
+ * @file
+ * Theme override to display a block.
```
brummbar: Let's update this to explain why we override the template, like we did for the other template.

web/themes/joinup/templates/block/block--toc-outline.twig
```diff
@@ -0,0 +1,31 @@
...
+ * - title_prefix: Additional output populated by modules, intended to be
+ *   displayed in front of the main title tag that appears in the template.
+ * - title_suffix: Additional output populated by modules, intended to be
+ *   displayed after the main title tag that appears in the template.
+ *
+ * @see template_preprocess_block()
+ */
+#}
+
+  {% block content %}
```
brummbar: I think we can just use `{% include 'block.html.twig' %} here.

web/modules/custom/custom_page/src/Plugin/Block/TableOfContentsOutline.php
```diff
@@ -0,0 +1,368 @@
...
+   *
+   * To get the correct element we will get the previous element from the
+   * flattened list which is naturally the next element in the list.
+   *
+   * @return \Drupal\Core\Menu\MenuLinkInterface|null
+   *   The previous menu link element or null if no element is found.
+   */
+  protected function getPrevElement(): ?MenuLinkInterface {
+    $flattened_menu = $this->getFlattenedMenu();
+    $active_link_id = $this->getActiveLink()->getPluginId();
```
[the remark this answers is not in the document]
idimopoulos: I will put the check in the beginning of the build method.

PR #1309, opened 2018-08-06
ISAICP-4658: Sequence of WYSIWYG input filters.

idimopoulos:
…iv elements are not duplicated.

### DIFF
--- a/web/modules/custom/joinup_user/joinup_user.install
+++ b/web/modules/custom/joinup_user/joinup_user.install
@@ -85,10 +85,3 @@ function joinup_user_setup_default_avatar() {
 
   $field_user_photo->set('settings.default_image', $default_image)->save(TRUE);
 }
-
-/**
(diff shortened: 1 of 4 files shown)


idimopoulos:
In the end it is indeed an issue with JoinupVideo but I do not think we should consider this a bug, rather than a mutual conflict.
What happens is that the video_embed_field video button creates a json markup which is later translated into an iframe during the rendering as processed text.
JoinupVideo was created to allow direct input of iframes into the text area and then checks whether the src property of the iframe matches one of the providers.
JoinupVideo reuses the functionality provided by video_embed_field module, including the render array and the classes.
The problem comes with then the video_embed_field translates a json array into an iframe, the JoinupVideo then checks if it is valid and re translates it. That causes a duplicated div container with the class set for responsive video which, in its turn, gives an extra blank space on the video bottom.

The solution is to not use this filter after video_embed_field WYSIWYG filter.

PR #1310, opened 2018-08-06
ISAICP-4686 Webtools Geocoding integration

pfrenssen:


composer.json
```diff
@@ -180,6 +182,9 @@
             "drupal/facets": {
                 "Facet Summary for Facet Items with a set dependency. @see https://www.drupal.org/node/2873480": "resources/patch/facet_summary_for_facet-2873480-18.patch"
             },
+            "drupal/filecache": {
+                "Provide an option to persist the cache. @see https://github.com/pfrenssen/filecache/pull/2": "https://patch-diff.githubusercontent.com/raw/pfrenssen/filecache/pull/2.patch"
```
claudiu-cristea: This concept of persisting cache over an cache clear is at least odd. I've never seen something similar. Cache is something that can and must be wiped out. Cache that survives its own bin "clear cache" command it's so non-API, IMHO. For me looks like an over-engineering but the problem is at the conceptual level: I've never seen a cache backend that, after clearing, the cache, is still persisting the cache items. I understand the reason why this has been added: _We need the cache items to serve as HTTP request replacements on tests_. But, that could have been accomplished with a very simple developer task:

1. Make the real request;
1. Manually copy/move the fixture files from the cache directory backend to the Behat (or PHPUnit) fixture  directory;
1. Commit them in the Git repo.

Is there something else that would require this level of complexity? And, again, not the complexity that scares me but I'm seriously questioning the concept of cache items being persisted over a cache clear.

composer.json
```diff
@@ -255,6 +260,16 @@
             "type": "vcs",
             "url": "https://github.com/Gizra/og.git"
         },
+        {
+            "_comment": "This directly references the repository. This can be removed once a release is made and the repository is announced on Packagist.",
+            "type": "vcs",
+            "url": "https://github.com/openeuropa/oe_webtools_location.git"
```
claudiu-cristea: I see it's registered by Packagist: https://packagist.org/packages/openeuropa/oe_webtools_location

composer.json
```diff
@@ -255,6 +260,16 @@
...
+        },
+        {
+            "_comment": "This directly references the repository. This can be removed once a release is made and the repository is announced on Packagist.",
+            "type": "vcs",
+            "url": "https://github.com/openeuropa/webtools-geocoding-provider.git"
```
claudiu-cristea: Looks registered already: https://packagist.org/packages/openeuropa/webtools-geocoding-provider

web/modules/custom/joinup_event/config/install/field.field.node.event.field_event_coordinates.yml
```diff
@@ -13,15 +13,15 @@ third_party_settings:
     weight: 0
     hidden: false
     disabled: false
+    providers:
+      - webtools_geocoding
```
claudiu-cristea: Just an observation and a subject of meditation: As `webtools_geocoding` is part of the field instance 3rd-party-settings, means that this field instance depends on `webtools_geocoding` config entity. As an effect, it should be listed in `field.field.node.event.field_event_coordinates.yml` under `dependencies.config`. We can do it manually, by editing the YAML file but, unfortunately, there's no API to do it in the `geocoder_field` (which is the module that adds the 3rd-party-setting). I've been discussed this, on Slack, with Alex Pott and he confirmed that there's no API mechanism that can hook into the "calculate dependencies" process and add/remove dependencies. I also suggested a future API expansion by adding a 4th parameter to `::setThirdPartySetting($module, $key, $value, $dependencies)`. In our case this would allow adding also dependencies along with the 3rd-party-settings. These are only thoughts around a configuration dependency inconsistency. But at geocoder level we should handle this in https://www.drupal.org/project/geocoder/issues/3016710.

build.project.xml
```diff
@@ -467,6 +506,6 @@
     <target
         name="install-dev"
         description="Install the website and set up the development environment."
-        depends="setup-virtuoso-permissions, purge-solr-backend, purge-virtuoso-backend, clean-redis-settings, install, ensure-search-api-solr-site-hash, setup-redis, import-rdf-fixtures, setup-dev, configure-apache-solr-drupal, configure-matomo-drupal, create-demo-users" />
+        depends="setup-virtuoso-permissions, purge-solr-backend, purge-virtuoso-backend, clean-redis-settings, disable-webtools-geocoding-cache, install, ensure-search-api-solr-site-hash, setup-redis, import-rdf-fixtures, setup-dev, configure-apache-solr-drupal, configure-matomo-drupal, create-demo-users, enable-webtools-geocoding-cache" />
```
claudiu-cristea: I don't think persisting the cache should be the default setup when installing. I did some manual tests and I ended up by Git status reporting a lot of new files. In th every rare cases when developer wants to create some Webtools cached items, he will manually enable the feature. However, I don't believe that we should go with this "persist cache switch" -- see bellow.

composer.json
pfrenssen: This cache persisting functionality also existed in the D7 version of the module, and there it was even the default strategy. I think use cases similar to ours (providing a file based cache of data provided by third parties) is one of the main use cases for even using this module.

See here for the D7 implementation of this persisting cache: https://cgit.drupalcode.org/filecache/tree/filecache.inc#n317

web/modules/custom/joinup_event/config/install/field.field.node.event.field_event_coordinates.yml
pfrenssen: > I've been discussed this, on Slack, with Alex Pott and he confirmed that there's no API mechanism that can hook into the "calculate dependencies" process and add/remove dependencies. 

I had a look in core at the third party settings implementation and couldn't find a way to set the config dependencies from there. Thanks for confirming this with Alex. Shall we make a core issue about this? This should be part of the API.

build.project.xml
pfrenssen: OK we can add this to the test pipeline.

composer.json
claudiu-cristea: That is wrong. At least, we can live with that in D7, but in D8 we have honour the cache backend contract. And the contract says:

```php
  /**
   * Deletes all cache items in a bin.
   */
  public function deleteAll();
```

Period.

But with file cache module, with this setting, when we call this method we end up with a non-empty bin. Now imagine an arbitrary module that uses this module and he needs this settings ('permanent') for given functionality. He will enable the setting in settings.php, but in other place of his application will use the same cache backend expecting normal results on a cache clear, specifically an empty bin. It's a confusing behavior. 

I see in D7 they also have there a 3rd option: persisting an amount of time (UNIX timestamp). In D8 this would overlap the `max-age` concept, which would create more confusion.

However, if you insist, I can merge this PR as it is but I cannot endorse the file cache upstream patch, which conceptually for me is not to be accepted. So, this can go with the patched module and, at some point, if anyone else will RTBC that issue, we can remove the patch from composer.

pfrenssen: OK I understand your point and agree. I still think this is a valuable feature for this particular module, in fact this was the reason I chose it, because it offers a persisting disk cache.

However the API contract breach should be communicated clearly to the users of the module. I will add a warning to the README and will also discuss it with Ognyan (the module maintainer) to get his input.

web/modules/custom/joinup_event/config/install/field.field.node.event.field_event_coordinates.yml
claudiu-cristea: I added https://www.drupal.org/project/drupal/issues/3016895

build.project.xml
pfrenssen: I have tried removing it but the results are not ideal. Without this every time a test is started it will abort immediately and inform that the Phing target needs to be executed first. Unfortunately this needs to be active for every test run otherwise we cannot detect if any requests were made during the test.

Wouldn't it be better to leave this in? Writing tests is a big part of the development work and I think it would become tiresome to execute this manually after every build.

composer.json
pfrenssen: I have resolved this by adding the relevant information to the README. I have also contacted the module maintainer to get his blessing.

PR #1317, opened 2018-08-09
4661: Spam-Tackling: Avoid the "Name=Surname" accounts

claudiu-cristea:


tests/features/user/register.feature
```diff
@@ -60,3 +60,27 @@ Feature: User registration
       | subject   | Your Joinup account was created successfully.                                                                                   |
       | body      | The Joinup Support Team created your account on Joinup. Please log in through the following link in order to set your password. |
     And the account for SuperUser should be active
+
+  Scenario: A user account whose first and last names are the identical is
+    deleted after creation but the user is receiving all notification as if the
+    account would have been created.
+
+    Given I am an anonymous user
+    And I go to "/user/register"
```
brummbar: Let's go through the interface
```
And I am on the homepage
And I click "Create new account"
```

tests/features/user/register.feature
```diff
@@ -60,3 +60,27 @@ Feature: User registration
...
+    Then I should see the success message "Thank you for applying for an account. Your account is currently pending approval by the site administrator."
+    And the following email should have been sent:
+      | recipient_mail | spam@example.com                                                                                                                                                                                                                                 |
+      | subject        | Your Joinup account is pending approval.                                                                                                                                                                                                         |
+      | body           | Thank you for registering at Joinup. Your application for an account is currently pending approval. Once it has been approved, you will receive another email containing information about how to sign in, set your password, and other details. |
+    # Creating a new account with same username and E-mail just to prove that,
+    # the in previous attempt, the user has been deleted.
+    Given user:
```
brummbar: Let's create a step similar to `\FeatureContext::assertUserExistence()` that tests that a user is not there.
So we can keep our business language and say `And I should not have a "spam" user`.

web/modules/custom/joinup_user/joinup_user.module
```diff
@@ -22,6 +22,13 @@ use Drupal\user\UserInterface;
  * Implements hook_ENTITY_TYPE_update().
  *
  * Send a notification whenever an admin updates a user profile.
+ *
+ * @todo This is a very complicated, hard to follow and hard to maintain
+ * implementation. Basically what we need here is to move the whole logic in the
+ * form API layer and avoid using all the confusing workarounds with the storing
+ * the form ID in a service. Then we can drop also the form alter and the
+ * 'joinup_user.active_form_submit' service. Fix this in ISAICP-4698.
+ * @see https://webgate.ec.europa.eu/CITnet/jira/browse/ISAICP-4698
```
brummbar: 👍

web/modules/custom/joinup_user/joinup_user.module
```diff
@@ -304,6 +311,34 @@ function joinup_user_form_alter(&$form, FormStateInterface $form_state, $form_id
...
+function joinup_user_submit_spam_guard(array $form, FormStateInterface $form_state) {
+  $user_id = $form_state->getFormObject()->getEntity()->id();
+  // We reload the user entity because the one stored in the form state is not
+  // entirely formed, and it's possible to triggers errors when other modules
+  // are participating to the entity deletion.
+  $user = User::load($user_id);
+  // If the user's family and fist names are identical, this is 99.99% spam.
+  if ($user->get('field_user_family_name')->value === $user->get('field_user_first_name')->value) {
+    $user->skip_notification = TRUE;
+    $user->delete();
```
brummbar: Neat! 👍

web/modules/custom/joinup_user/joinup_user.module
```diff
@@ -304,6 +311,34 @@ function joinup_user_form_alter(&$form, FormStateInterface $form_state, $form_id
...
+/**
+ * Represents an additional submit callback for the user registration form.
+ *
+ * @param array $form
+ *   The form render array.
+ * @param \Drupal\Core\Form\FormStateInterface $form_state
+ *   The current form state.
+ */
+function joinup_user_submit_spam_guard(array $form, FormStateInterface $form_state) {
+  $user_id = $form_state->getFormObject()->getEntity()->id();
```
brummbar: I couldn't test practically, but can't we use the `$form_state->get('user')` (see `web/core/modules/user/src/RegisterForm.php:96`) ?

tests/features/user/register.feature
claudiu-cristea: I don't see the value this change could bring. However, I'm not against changing it.

web/modules/custom/joinup_user/joinup_user.module
claudiu-cristea: Well, initially I did `$user = $form_state->getFormObject()->getEntity()`, then on `$user->delete()` but I received the exception:

```
Error: Call to a member function get() on null in pathauto_entity_delete() (line 108 of modules/contrib/pathauto/pathauto.module).
```
Same happens with `$form_state->get('user')`. The reason was that `$entity->get('path')` is `NULL` in this stage because it seems that `path` fiels is not yet populated as the entity building in not over yet. But an entity fill load gives a full formed entity.

claudiu-cristea:
Fixed the QA remarks except tha last one which should stay as it is.

PR #1318, opened 2018-08-10
4368: Send a notification to contact form submitter

claudiu-cristea:


web/modules/custom/contact_form/config/install/field.field.message.contact_form_submission.field_contact_copy.yml
```diff
@@ -0,0 +1,22 @@
...
+dependencies:
+  config:
+    - field.storage.message.field_contact_copy
+    - message.template.contact_form_submission
+id: message.contact_form_submission.field_contact_copy
+field_name: field_contact_copy
+entity_type: message
+bundle: contact_form_submission
+label: 'Send yourself a copy'
+description: 'Send yourself a copy of this message.'
```
pfrenssen: It's not necessary to have both the label and the description since they both say basically the same thing. Just having the label is sufficient. If you think "Send yourself a copy" is not clear enough we can change the label to "Send yourself a copy of this message".

web/modules/custom/contact_form/contact_form.module
```diff
@@ -71,7 +72,16 @@ function contact_form_form_message_contact_form_submission_submit(&$form, FormSt
...
   $event_dispatcher->dispatch(ContactFormEvents::CONTACT_FORM_EVENT, $event);
-  drupal_set_message(t('Your message has been submitted. Thank you for your feedback.'));
+
+  if ($event->isSuccessful()) {
+    \Drupal::messenger()->addStatus(t('Your message has been submitted. Thank you for your feedback.'));
+  }
+  else {
+    \Drupal::messenger()->addError(t('The message submission failed.'));
+    \Drupal::logger('contact')->error('Cannot send contact form message.', [
+      'message' => $event->getEntity()->toArray(),
```
pfrenssen: Good idea to log an error when the message cannot be sent.

web/modules/custom/contact_form/contact_form.module
```diff
@@ -117,3 +127,44 @@ function contact_form_mail_alter(&$message) {
...
+    $subject = PlainTextOutput::renderFromHtml($subject);
+    $message['body'][] = t('Subject: %subject', ['%subject' => $subject]);
+    $message['body'][] = t('Content:');
+    $message['body'][] = '<blockquote>';
+    $message['body'][] = check_markup($entity->field_contact_message->value, $entity->field_contact_message->format);
+    $message['body'][] = '</blockquote>';
+    $message['body'][] = t('Thank you for your interest, we will revert back to you as soon as possible.');
+  }
+  else {
+    $message['body'][] = t('We will revert back to you as soon as possible.');
```
pfrenssen: The expression "revert back to you" is not correct English, this should be something like "We will reply as soon as possible". I will confirm this with the functional team who is responsible for delivering the user facing strings.

PR #1321, opened 2018-08-13
4706: Fight the deadlock bug

claudiu-cristea:


pfrenssen:
I proposed a PR to OG that will clean up orphans after each Behat scenario rather than at the very end of the test suite. This should hopefully fix the timeouts for good.

See https://github.com/Gizra/og/pull/424

pfrenssen:
BTW the error message in the `PDOException` is misleading, this is not a true deadlock but the lengthy transactions are exceeding the lock timeouts set in MySQL.

pfrenssen:
The exception occurred again. It seems cleaning up the orphans after each scenario doesn't solve the timeouts.

I noticed something interesting: every time we have this failure it is caused by the OgMembership entity with ID 172. Let's do some digging which entity this is and in which test this is created. This might help us to narrow down the problem.

Launched a test that shows some debugging data about this entity: https://app.continuousphp.com/git-hub/ec-europa/joinup-dev/build/97cddfe6-be2a-4fbf-b9c0-c8e6b84852c2/logs/9dd082ac-0aad-4637-8ac4-036ff87f19b1

pfrenssen:
Patch that adds debugging data had a bug: `Fatal error: Call to a member function getEntityTypeId() on null `. Fixed it and launched a new test: https://app.continuousphp.com/git-hub/ec-europa/joinup-dev/build/b7f3e4d0-b8fc-4483-9a73-aa8441689a41

pfrenssen:
No error, launched new test: https://app.continuousphp.com/git-hub/ec-europa/joinup-dev/build/9aac4662-c217-440d-aafa-b349e00f3396

pfrenssen:
```
Error: Call to a member function label() on null in /home/cphp/data/container/4b112bb0-3ef9-4421-bda8-3a54293df39d/workspace/web/modules/contrib/og/src/OgDeleteOrphansBase.php:143
```

pfrenssen:
Starting tests:

- [ ] https://app.continuousphp.com/git-hub/ec-europa/joinup-dev/build/6a69c8ed-1c58-41fe-82fa-51fc3ee179b3/logs/88bd72b2-d9d4-4bce-b387-90cbe4589e48 - Failed because of #1326 
- [ ] https://app.continuousphp.com/git-hub/ec-europa/joinup-dev/build/b315cae4-20b5-49a6-bba1-4fb115ea4015

PR #1335, opened 2018-08-23
ISAICP-4730: enable custom pages to show content shared in the collection

brummbar:


web/modules/custom/custom_page/config/install/field.field.node.custom_page.field_cp_content_listing.yml
```diff
@@ -24,6 +24,7 @@ default_value:
       enabled: 0
       query_presets: ''
       limit: 12
+      show_shared: 0
```
idimopoulos: I think this is causing the tests to fail. The `show_shared` property of the value field is dynamic. You declare it dynamically into the custom widget that extends the default widget and you already set the default value there (`'#default_value' => $default_values['show_shared'] ?? FALSE`).

This little piece here however, assumes that the show_shared **column** is part of the definition of the `field.value.search_api_field` defined in search_api_field.schema.yml.

I am not sure if you can simply override the mapping to add an extra property. But I am not that expert anyway.

In any case, removing this should fix the issues.

PR #1342, opened 2018-08-29
Isaicp 4092

pfrenssen:


pfrenssen:
When working on this I noticed that the error handling in the update script was no longer working as intended. The original approach was to chain all commands using `&&` which will cause the process to abort on the first error, and then the exit code was examined and an error message was output if needed.

Unfortunately the chained sequence of commands was broken in 3 places, caused by new commands being added to the script. This completely defeats the error handling. This was not caught in QA since our team consists of PHP experts, and not everybody knows all the details of BASH scripting.

I've reworked this to trap errors instead. This is not the best possible solution since if an error would occur early on in the procedure this will not halt it, but will allow subsequent steps to continue. But I think this is a safer way to deal with it without becoming overly verbose. At least developers with a basic knowledge of BASH scripting will then be able to work on the script without the risk that the entire error handling will be broken.

scripts/update_production.sh
```diff
@@ -6,31 +6,54 @@
...
+if [ ${STATUS} -ne 0 ]; then
+  echo "The following line is missing from web/sites/default/settings.php:"
+  echo '$settings['\''config_readonly'\''] = !file_exists(getcwd() . '\''/../disable-config-readonly'\'');'
+  exit ${STATUS}
+fi
+
+echo "Disabling config_readonly."
+touch disable-config-readonly
+
 ./vendor/bin/drush updatedb --yes &&
```
claudiu-cristea: Well, I have a problem with this approach. I find it better to exit with an error after each command failure. If `updatedb` is failing, for instance, I don't want to continue.  It just happened to me 2 days ago -- a build failed in `updb` but it took me 10 minutes to read the console log starting from bottom, till I found the cause. But probably this out-of-scope? I would like to hear your opinion.

pfrenssen: Yes this has been restored in this PR. All commands that are time intensive are chained together again, and I have introduced an error trap that will cause any error to cause the build to fail.

pfrenssen:
Addressed the remarks. I also made sure the script checks both `settings.php` and `settings.local.php` for the setting, since it is in fact preferable for this change to be put in `settings.local.php`.

PR #1346, opened 2018-09-03
ISAICP-4740: Re-use the logic from the link formatter to the uri field of tallinn formatter.

idimopoulos:


web/modules/custom/joinup_communities/tallinn/src/Plugin/Field/FieldFormatter/TallinnEntryFormatter.php
```diff
@@ -57,10 +102,17 @@ public function viewElements(FieldItemListInterface $items, $langcode) {
...
 
     if (!empty($value['uri'])) {
+      $link_title = $value['uri'];
+
+      // Trim the link text to the desired length.
+      if (!empty($settings['trim_length'])) {
+        $link_title = Unicode::truncate($value['uri'], $settings['trim_length'], FALSE, TRUE);
+      }
+
       $element['#uri'] = [
```
claudiu-cristea: UX, Accessibility: Now that the URL will be trimmed, the user could not see the whole URL if for some reasons he's interested. Let's add the `title` attribute where we se the full URL so that when he hovers over the link he can read the full address.

idimopoulos:
For the remarks,
1. I have created a follow up issue for the front end theming on mobiles. It seems that the issue is not only happening on tallinn entries.
2. I noticed that but the value we are setting is the default so I thought that I don't need to update it. However, since number 3 is referring to its schema, it is a good point to add it.
3. I have added a small schema configuration.
4. Done.

PR #1348, opened 2018-09-03
Isaicp 4716: Fix URL pointing to contact form, on emails.

idimopoulos:


web/modules/custom/joinup_notification/src/EventSubscriber/NotificationSubscriberBase.php
```diff
@@ -310,7 +310,7 @@ protected function generateArguments(EntityInterface $entity) {
         $actor_first_name . ' ' . $actor_family_name :
         $actor->get('full_name')->value;
     }
-    $arguments['@site:contact_url'] = Url::fromRoute('contact_form.contact_page')->toUriString();
+    $arguments['@site:contact_url'] = Url::fromRoute('contact_form.contact_page', [], ['absolute' => TRUE])->toString();
```
claudiu-cristea: The solution in 100% correct. Just for the future, there's a more elegant way to do the same thing:
```php
Url::fromRoute('contact_form.contact_page')->setAbsolute()->toString();
```

PR #1364, opened 2018-09-20
ISAICP-4771: Change hook_ENTITY_TYPE_load to hook_entity_storage_load

idimopoulos:


web/modules/custom/solution/solution.module
```diff
@@ -317,9 +317,13 @@ function solution_entity_extra_field_info(): array {
 }
 
 /**
- * Implements hook_ENTITY_TYPE_load().
+ * Implements hook_entity_storage_load().
  */
-function solution_rdf_entity_load($entities) {
+function solution_entity_storage_load(array $entities, $entity_type) {
+  if ($entity_type !== 'rdf_entity') {
```
brummbar: You can use `hook_ENTITY_TYPE_storage_load()` :)

PR #1370, opened 2018-09-27
Isaicp 4777: Allow moderators to access contact_form files.

idimopoulos:


tests/features/contact_form/submit_contact_form.feature
```diff
@@ -43,6 +43,16 @@ Feature: Submit the contact form
     And I should see the following success messages:
       | success messages                                              |
       | Your message has been submitted. Thank you for your feedback. |
+    When I click the link matching the "#https?://[^/].*?/contact_form/\d{4}-\d{2}/logo(_\d+)?.png#" pattern from the email sent to "digit-joinup@ec.europa.eu"
```
pfrenssen: Let's try to rewrite this in a way that is understandable by the business stakeholders and the functional team. Regular expressions are a bit too technical :sweat_smile:

web/modules/custom/joinup_core/joinup_core.behat.inc
```diff
@@ -79,4 +81,57 @@ class JoinupCoreSubContext extends DrupalSubContextBase implements DrupalSubCont
...
+   * @param string $mail
+   *   The recipient's mail.
+   *
+   * @throws \Exception
+   *   If no mail was found linked to the given email or if the pattern did not
+   *   find any matches.
+   *
+   * @Given I click the link matching the :pattern pattern from the email sent to :mail
+   */
+  public function clickMailLinkMatchingPattern(string $pattern, string $mail): void {
```
pfrenssen: I would move this into `joinup_notification.behat.inc` because this depends on the `@email` tag that is declared in that file. Otherwise we introduce a dependency from `joinup_core.behat.inc` on `joinup_notification.behat.inc`.

tests/features/contact_form/submit_contact_form.feature
```diff
@@ -43,6 +43,16 @@ Feature: Submit the contact form
     And I should see the following success messages:
       | success messages                                              |
       | Your message has been submitted. Thank you for your feedback. |
+    When I click the link for the "logo.png" attachment in the contact form confirmation email sent to "digit-joinup@ec.europa.eu"
```
idimopoulos: This is not a confirmation email, simply an email.

web/modules/custom/joinup_notification/joinup_notification.behat.inc
```diff
@@ -346,60 +367,164 @@ class JoinupNotificationSubContext extends DrupalSubContextBase implements Drupa
...
+  }
 
-    $this->visitPath($match[0]);
+  /**
+   * Checks if the current scenario or feature has the @email tag.
+   *
+   * Call this in steps that use the test mail collector so that the developer
+   * is alerted if this tag is not present.
+   */
+  protected function assertEmailTagPresent(): void {
```
idimopoulos: I am fine with having something like this as the same idea is already used for the `@api` tag.
However, I have used the absence of the `@email` tag for debugging purposes, e.g. I knew that the assertion would fail, but I wanted to test the output of the mail in mailcatcher. 
Maybe use a warning instead of an assertion?

pfrenssen: If it is for debugging, you can just temporarily comment it out or put a `return;` statement at the top?

The idea behind this is to stop the test as early as possible if the tag is missing, so that the dev doesn't waste time.

tests/features/contact_form/submit_contact_form.feature
idimopoulos: I am creating a follow up on this.

PR #1372, opened 2018-09-28
Isaicp 4759: Provide a report page of solutions listed by licence.

idimopoulos:


web/profiles/joinup/src/Form/SolutionsByLicenceForm.php
```diff
@@ -0,0 +1,240 @@
...
+    }
+
+    $form['table'] = [
+      '#theme' => 'table',
+      '#header' => ['Licence', 'Solution'],
+      '#rows' => $rows,
+      '#empty' => $this->t('No solutions available.'),
+    ];
+    $form['pager'] = [
+      '#type' => 'pager',
```
pfrenssen: This pager is not being tested, but that's OK for me. This is a moderator-only functionality which is not usable by the public. I think it is not necessary to spend the testing time for creating a matrix of 50+ licenses and solutions for this feature which will probably be very little used.

tests/features/moderator/solutions_by_licence.feature
```diff
@@ -0,0 +1,68 @@
...
+  As a moderator of the site
+  In order to be able to manage licences of the website
+  I need to be able to list the licences and their related solutions.
+
+  Scenario: Only privileged users can access the report.
+    When I am logged in as a user with the "authenticated" role
+    And I go to "/admin/reporting/solutions-by-licences"
+    Then the response status code should be 403
+
+    When I am logged in as a user with the 'access joinup reports' permission
```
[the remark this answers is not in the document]
idimopoulos: I thought that it was a better way but yeah, you are right, we don't have the give/revoke permission procedure in Joinup so it is better to test as a moderator.

tests/features/moderator/solutions_by_licence.feature
```diff
@@ -0,0 +1,68 @@
...
+      | Solution_by_licence_1.exe | Free for all | Solution by licence 1 |
+    # Distribution that belongs to the release of a solution.
+    And the following distribution:
+      | title   | Solution_by_licence_2.tar |
+      | licence | Commercial                |
+    And the following release:
+      | title          | Solution by licence 2     |
+      | release number | 1                         |
+      | distribution   | Solution_by_licence_2.tar |
+      | is version of  | Solution by licence 2     |
```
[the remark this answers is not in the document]
idimopoulos: Usually that is what we have actually. The release is named the same as the solution, and when viewing the release, the version is concatenated with the title and shown on the header. But I will change it to make it clear for the tests.

web/profiles/joinup/src/Form/SolutionsByLicenceForm.php
idimopoulos: Actually that was the only reason I did not test it. I did test manually the extend of the pager and if all pages were ok, but to write a test with 50+ cases just for the pager seemed too much for me :(

web/profiles/joinup/src/Form/SolutionsByLicenceForm.php
```diff
@@ -159,28 +153,31 @@ protected function getLicenceOptions(): array {
    */
   protected function getCountQuery(string $licence_id = NULL): string {
     $query = <<<QUERY
-SELECT COUNT(*) as ?total
```
idimopoulos: Please note, that there was an update in the query to ensure that if a solution has more than one distribution referencing the same licence, there will be only one result shown in the table.

claudiu-cristea:
As the SPARQL triples are wrapped as Drupal entities, I don't like the approach that has been taken. And this is not a matter of taste but a We have to use the Drupal entity abstraction layer as much as we can. Ideally, we would fix this by using a more abstract layer: Drupal Views. But as it has been argued, that would have been technically too complex. Then, the next option should be the Drupal entity query API which we can use, as licences are Drupal `rdf_entity` entities. We are doing direct calls to triples only in some limited cases, where we have severe limitations on the Drupal API layer.

PR #1383, opened 2018-10-05
Unpin SAPI: 2nd try

claudiu-cristea:


tests/src/Traits/ContextualLinksTrait.php
```diff
@@ -83,7 +83,7 @@ protected function findContextualLinkButton(NodeElement $element): NodeElement {
     // Check if the wrapper for the contextual links is present on the page.
     // Since the button is appended by the contextual.js script, we might need
     // to wait a bit before the button itself is visible.
-    $button = $element->waitFor(5, function (NodeElement $element): ?NodeElement {
```
brummbar: How come we need such a big increase? It shouldn't take this much to render the links...

composer.json
```diff
@@ -238,7 +235,8 @@
                 "Download link returns 406 with Drupal ^8.5.0 @see https://www.drupal.org/project/views_data_export/issues/2951185": "https://www.drupal.org/files/issues/2951185-2.patch"
             },
             "drush/drush": {
-                "When an update fails, updatedb is still exiting with 0. @see https://github.com/drush-ops/drush/pull/3673": "https://patch-diff.githubusercontent.com/raw/drush-ops/drush/pull/3673.patch"
+                "When an update fails, updatedb is still exiting with 0. @see https://github.com/drush-ops/drush/pull/3673": "https://patch-diff.githubusercontent.com/raw/drush-ops/drush/pull/3673.patch",
+                "Call services that need destruction at the end of request": "https://github.com/drush-ops/drush/commit/346f165cb160c554293dc4839bcba28e4368051d.diff"
```
pfrenssen: This commit has in the meantime been merged in to Drush, we can remove it if we use version 9.5.0 as the minimum version.

tests/src/Traits/ContextualLinksTrait.php
pfrenssen: I think this was increased by @claudiu-cristea since in this PR a cache clear was added in the `featured_entities` test to temporarily work around an unknown issue. One of the steps following this cache clear is involving contextual links. I'm guessing it might take more than 5 seconds on test infrastructure to generate the contextual links on a cold cache. This is not surprising given the fact that they are generated in a separate request.

I'm cool with this change. It reduces the chance of random failures, and it doesn't impact the test speed at all, since this `ElementInterface::waitFor()` method is polling every 100ms and will exit as soon as the condition is met. Only when this exact step will fail, the test runner will wait 30 seconds instead of 5 before failing the test. This should only occur during development so it is acceptable for me.

I'm open to change this to another value though, but this is not blocking the PR for me.

PR #1386, opened 2018-10-09
ISAICP-4843: upgrade path to delete the sub pages menu block configuration.

brummbar:


pfrenssen:
Looking good, thanks! Since this is an update hook that removes a block and the block UI is not exposed to end users this can not be checked in user acceptance testing. Merging directly into the epic.

PR #1406, opened 2018-10-29
ISAICP-4870: Re-export index configuration to include the field_adtl_address_info.

idimopoulos:


idimopoulos:
The additional address field should not be included in the search index actually. Closing.

PR #1418, opened 2018-11-16
Isaicp 4486

sandervd:


web/modules/custom/rdf_serialization/src/Normalizer/RdfEntityNormalizer.php
```diff
@@ -0,0 +1,39 @@
...
+class RdfEntityNormalizer extends NormalizerBase {
+
+  use FieldableEntityNormalizerTrait;
+
+  /**
+   * The interface or class that this Normalizer supports.
+   *
+   * @var string
+   */
+  protected $supportedInterfaceOrClass = 'Drupal\rdf_entity\RdfInterface';
```
claudiu-cristea: Let's not hardcode the namespaced interface name. Use `RdfInterface::class` instead.

web/modules/custom/rdf_serialization/rdf_serialization.services.yml
```diff
@@ -0,0 +1,30 @@
...
+    class: Drupal\rdf_serialization\Encoder\RdfEncoder
+    tags:
+    - { name: encoder, format: n3 }
+  rdf_serialization.rdfsubscriber:
+    class: Drupal\rdf_serialization\EventSubscriber\RdfSubscriber
+    tags:
+      - { name: event_subscriber }
+  serializer.normalizer.entity.rdf:
+    class: Drupal\rdf_serialization\Normalizer\RdfEntityNormalizer
+    arguments: ['@rdf_export.serializer']
```
claudiu-cristea: For me this looks weird. Why depending on `rdf_export` module? I think this service should be moved in `rdf_serialization` and not the other way around. Then `rdf_export` would depend on `rdf_serialization`.

web/profiles/joinup/joinup.behat.inc
```diff
@@ -2562,4 +2562,28 @@ JS;
...
+   */
+  public function theOutputShouldMatchTheFile($file_path) {
+    if ($this->getMinkParameter('files_path')) {
+      $fullPath = rtrim(realpath($this->getMinkParameter('files_path')), DIRECTORY_SEPARATOR) . DIRECTORY_SEPARATOR . $file_path;
+      if (is_file($fullPath)) {
+        $file_path = $fullPath;
+      }
+    }
+    $expected_content = file_get_contents($file_path);
+    $browser_content = $this->getSession()->getPage()->getContent();
```
claudiu-cristea: Let's `trim()` both, `$expected_content` and `$browser_content`, to ensure that a trailing new line at the end of the file will not affect the assertion. For the sake of coding standards I would add the new line at the end of the fixture files.

PR #1424, opened 2018-11-22
ISAICP-4898: content type boosts are not working anymore

brummbar:


web/modules/custom/joinup_search/joinup_search.module
```diff
@@ -165,12 +165,15 @@ function joinup_search_search_api_solr_field_mapping_alter(IndexInterface $index
...
-  if (
-    !$solarium_query instanceof Query
-  ) {
+  if (!$solarium_query instanceof Query) {
     return;
   }
 
+  // Keep the edismax parser to allow document boosting to work properly.
+  // @todo remove when document boosting is reimplemented by search_api_solr
+  // @see https://webgate.ec.europa.eu/CITnet/jira/browse/ISAICP-4907
```
claudiu-cristea: Good to see that we have a follow-up.

PR #1429, opened 2018-11-27
Temporarily lock Facets on the previous release.

pfrenssen:
The recently released updates is causing test failures.


pfrenssen:
There is also a failure due to the update of Search API Solr :sweat:

Followup tickets to unlock both:
* https://webgate.ec.europa.eu/CITnet/jira/browse/ISAICP-4915
* https://webgate.ec.europa.eu/CITnet/jira/browse/ISAICP-4924

PR #1431, opened 2018-11-29
Isaicp 4372: Redirect to solution after release deletion.

idimopoulos:


web/modules/custom/asset_release/asset_release.module
```diff
@@ -314,3 +314,29 @@ function asset_release_form_rdf_entity_asset_release_edit_form_alter(&$form, For
...
+ * @param array $form
+ *   The form array.
+ * @param \Drupal\Core\Form\FormStateInterface $form_state
+ *   The form state.
+ */
+function asset_release_delete_redirect(array &$form, FormStateInterface $form_state) {
+  $release = $form_state->getFormObject()->getEntity();
+  /** @var \Drupal\asset_release\AssetReleaseRelations $relation_service */
+  $relation_service = \Drupal::service('asset_release.relations');
+  $solution = $relation_service->getReleaseSolution($release);
```
claudiu-cristea: Just a side-note. We should drop `::getReleaseSolution()` in favour of a `parent` computed field as we did with distributions in ADMS Importer epic.

claudiu-cristea: Created https://webgate.ec.europa.eu/CITnet/jira/browse/ISAICP-4948 as a follwup.

PR #1444, opened 2018-12-10
Isaicp-4471 - 2: Remove the username from the combined filter.

idimopoulos:


idimopoulos:
That PR is wrong. sorry for that. https://github.com/ec-europa/joinup-dev/pull/1427 is the correct one.

PR #1454, opened 2019-01-03
Isaicp 4955: Option to disable related solutions

idimopoulos:


tests/features/solution/related_solution.feature
```diff
@@ -46,3 +48,9 @@ Feature: Related solution
     And I press "Publish"
     And I visit the "Java" solution
     Then I should see the "Python" tile
+
+    # Solutions that have 'Solution related by type' off, should not show solutions related by type.
```
brummbar: Can we add a test that if we disable/enable it they are shown/hidden?

PR #1459, opened 2019-01-10
Isaicp 4925

pfrenssen:


web/modules/custom/joinup_event/config/install/field.field.node.event.field_location.yml
```diff
@@ -9,8 +9,8 @@ field_name: field_location
 entity_type: node
 bundle: event
 label: Location
-description: ''
-required: true
+description: 'If the event takes place in a physical location, provide the street address so a map can be shown.'
+required: false
```
claudiu-cristea: With this change the location is no more mandatory. That means an event could be created without a location. Probably we should have a follow-up to implement a custom validation that checks if at least on location is non-empty. EDIT: I've discussed with Katerina, we'll fill a follow-up to add a custom validation.

PR #1466, opened 2019-01-21
4953: Disable ToC from Custom pages without sub pages

claudiu-cristea:


web/modules/custom/custom_page/src/Plugin/Block/TableOfContentsOutline.php
```diff
@@ -124,9 +106,9 @@ public static function create(ContainerInterface $container, array $configuratio
       $plugin_id,
       $plugin_definition,
       $container->get('custom_page.og_menu_links_manager'),
-      $container->get('current_route_match'),
       $container->get('menu.link_tree'),
-      $container->get('menu.active_trail')
+      $container->get('menu.active_trail'),
+      $container->get('menu.tree_storage')
```
claudiu-cristea: This parameter is useless (it's left there from a previous iteration).

PR #1494, opened 2019-02-11
Isaicp 4906: Remove 'advanced search' link from collections and solutions

pfrenssen:
It will be replaced with the new advanced search functionality.


idimopoulos:
Nice cleanup of things. A follow up has been created for a couple of front-end entries.
Accepted in QA.

PR #1503, opened 2019-02-20
Issue #ISAICP-5108: Update drupal core to latest version.

sandervd:


composer.lock
```diff
@@ -13933,6 +13930,7 @@
                 "mock",
                 "xunit"
             ],
+            "abandoned": true,
```
pfrenssen: This is worrying but out of scope for this PR. I made a followup for this: https://webgate.ec.europa.eu/CITnet/jira/browse/ISAICP-5112

PR #1507, opened 2019-02-21
Isaicp 5084

junkuncz:


web/modules/custom/joinup_search/config/install/facets.facet.type.yml
```diff
@@ -11,7 +11,7 @@ name: Type
 url_alias: type
 weight: 0
 min_count: 1
-show_only_one_result: true
+show_only_one_result: false
```
claudiu-cristea: We need to test that multiple choices can be selected. Probably you can do that in `tests/features/joinup_search/search.feature`.

web/modules/custom/joinup_search/config/install/facets.facet.type.yml
```diff
@@ -22,6 +22,7 @@ widget:
     soft_limit_settings:
       show_less_label: 'Show less'
       show_more_label: 'Show more'
+    arbitrary_facet_plugin: _none
```
claudiu-cristea: What is the reason for this change. This made tests to fail.

junkuncz: Looks like config export put this line automatically during the export. Don't think that it relates to my current feature so will remove it.

web/modules/custom/joinup_search/config/install/facets.facet.type.yml
junkuncz: I'm on it!

junkuncz: Test added to multiple choices and broken tests also have been fixed.

PR #1520, opened 2019-02-27
4978: Content negotiation for taxonomy terms

claudiu-cristea:


composer.json
```diff
@@ -46,7 +46,7 @@
         "drupal/persistent_login": "~1.0",
         "drupal/phingdrushtask": "dev-7.x-2.x#f4b020f09da3c174fce40f2bd9a9617689f57124",
         "drupal/r4032login": "dev-1.x#391f10d69b9b8c466a2fddac2a6b99f1289e7781",
-        "drupal/rdf_entity": "^1.0.0-alpha12",
+        "drupal/rdf_entity": "dev-serializer",
```
idimopoulos: We should include this as a patch to avoid switching branches. Or at least create a new alpha version before proceeding.

claudiu-cristea: Please see my comment in the ticket.

idimopoulos:
While in `https://github.com/ec-europa/joinup-dev/pull/1520/files#diff-f6359d3c520c3fe95b6fa1b65e98b70fR14`(web/profiles/joinup/config/install/rest.resource.entity.taxonomy_term.yml) the "xml" format is not supported, when I insert `?_format=xml`, I do retrieve content in xml format with a text saying that "xml" is not supported as a format. When I use any other invalid keyword, the normal error page appears with the uuid attached.

composer.json
claudiu-cristea: I've been created the upsteream issue https://www.drupal.org/project/drupal/issues/3036459 to fix this for ever.

web/profiles/joinup/config/install/rest.resource.entity.taxonomy_term.yml
```diff
@@ -0,0 +1,22 @@
...
+configuration:
+  methods:
+    - GET
+  formats:
+    - jsonld
+    - rdfxml
+    - ntriples
+    - turtle
+    - n3
+    - json
```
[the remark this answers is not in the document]
claudiu-cristea: I removed this format as it overlaps with the core `json` format. There's a followup on how to proceed in the case of overlapping.

claudiu-cristea:
> While in `https://github.com/ec-europa/joinup-dev/pull/1520/files#diff-f6359d3c520c3fe95b6fa1b65e98b70fR14`(web/profiles/joinup/config/install/rest.resource.entity.taxonomy_term.yml) the "xml" format is not supported, when I insert `?_format=xml`, I do retrieve content in xml format with a text saying that "xml" is not supported as a format. When I use any other invalid keyword, the normal error page appears with the uuid attached.

When you use `_format=xml`, you require a response as XML. See https://joinup.ec.europa.eu/news/instructions-your-language?_format=json, which is a node. Imagine that there’s a consumer, such as a mobile phone. He doesn’t understand HTML, he’s requesting to talk the JSON or XML language with the server. Even errors should be returned accordingly. The meaning of the message is NOT _we not support XML_ but _this resource (entity) doesn’t have a XML representation_.

PR #1522, opened 2019-02-27
ISAICP-5030: Show a warning in deprecated terms from the EIRA taxonomy

idimopoulos:


web/modules/custom/joinup_core/src/Form/RdfEntityUriWebConverterForm.php
```diff
@@ -37,14 +38,19 @@ public function validateForm(array &$form, FormStateInterface $form_state) {
...
 
     // An absolute URL.
     if (UrlHelper::isExternal($id)) {
       // Potential RDF entity ID.
       if (!UrlHelper::externalIsLocal($id, $base_url)) {
         if (Rdf::load($id)) {
           $form_state->set('passed_as', 'id');
+          $form_state->set('entity_type', 'rdf_entity');
+        }
+        elseif (RdfTerm::load($id)) {
```
claudiu-cristea: This is now hardcoded. Normally, we would iterate over all content entities that have SPARQL storage and try to load with each, one by one. That would generalize to any new SPARQL entity type that might be added in the future. But for now I agree that this is a very "Joinup centric" functionality and we can go with the hardcoding.

web/modules/custom/joinup_core/src/Form/RdfEntityUriWebConverterForm.php
```diff
@@ -56,29 +62,35 @@ public function validateForm(array &$form, FormStateInterface $form_state) {
...
   }
 
   /**
    * {@inheritdoc}
    */
   public function submitForm(array &$form, FormStateInterface $form_state) {
     $passed_as = $form_state->get('passed_as');
-    $id = $form_state->getValue('rdf_entity_id');
+    $id = trim($form_state->getValue('entity_id'), " /");
+    $entity_type = $form_state->get('entity_type');
```
claudiu-cristea: It's a good practice to use core naming: `$entity_type_id`. `$entity_type` is the object.

PR #1530, opened 2019-03-04
5106: Cached computed field cron error

claudiu-cristea:


web/modules/custom/joinup_core/joinup_core.module
```diff
@@ -1381,47 +1380,12 @@ function joinup_core_module_implements_alter(array &$implementations, $hook) {
...
 }
-
-/**
- * Implements hook_ENTITY_TYPE_update().
- */
-function joinup_core_node_update(NodeInterface $node) {
-  /** @var \Drupal\node\NodeInterface $original */
-  $original = $node->original;
-
-  $log = function () use ($node, $original): bool {
```
pfrenssen: Is this debugging code that somehow ended up in the develop branch?

pfrenssen:
This is looking good. I think this bug fix is serious enough to warrant a new release for the Cached Computed Field module. I am now running a test against the latest 8.x-1.x branch of the module. If this goes green I will tag a new 8.x-1.0-beta4 release and revert back to the beta in `composer.json`.

web/modules/custom/joinup_core/joinup_core.module
claudiu-cristea: Yes. As we've been discussed Friday in the standup, we revert that in the scope of this issue. However, we've decided to keep the switch for cached computed field. So I needed first to bring the code in `develop`.

PR #1540, opened 2019-03-12
Isaicp 4974: Include oe_webtools_analytics in the project

idimopoulos:


web/modules/custom/isa2_analytics/src/EventSubscriber/WebtoolsAnalyticsSubscriber.php
```diff
@@ -48,6 +59,24 @@ public function setSiteSection(AnalyticsEventInterface $event) {
...
+
+        $collection = NULL;
+        // Normally, the solution has the `collections` computed field that
+        // holds the list of collections that are affiliates to the solution.
+        // However, this method is fired during the preparation of the page
+        // while the computation of the value takes place during the rendering
+        // of the page.
+        // @see \Drupal\Core\Render\MainContent\HtmlRenderer::renderResponse
+        if ($collections = solution_get_collection_ids($group)) {
+          $collection_id = reset($collections);
```
pfrenssen: Let's put an explanation here why we are only taking the first collection. It might not be obvious to future generations that this is because of a business decision.

PR #1550, opened 2019-03-18
Disable EC video tests

claudiu-cristea:


claudiu-cristea:
The main problem is fixed. Closing.

PR #1551, opened 2019-03-18
ISAICP-4983: Fixes in the newsletter tile

tom-waat:


pfrenssen:
I investigated the test failure and it appears to be caused by the AVPortal API endpoint change which is fixed in https://github.com/ec-europa/joinup-dev/pull/1552. I merged in latest develop so we can retest with the fixed endpoint.

PR #1559, opened 2019-03-20
ISAICP-5158: Only show download count when there is at least 1 download

junkuncz:


tests/features/solution/add_solution.feature
```diff
@@ -247,3 +247,35 @@ Feature: "Add solution" visibility options.
...
+    Then I visit "/solutions"
+    Then I should see the text "10"
+    When I go to the homepage of the "Climate change tracker" solution
+    # Check the same thing on the homepage of the solutions.
+    Then I should see the text "Downloads: 10"
+    When I delete the "Sample distribution" asset distribution
+    Given the cache has been cleared
+    And I reload the page
+    # Check it again after the first distribution is deleted and cache has been cleared.
+    Then I should not see the text "Downloads: 10"
```
claudiu-cristea: Same. Do a more exact check, see above.

web/themes/joinup/templates/rdf_entity/rdf-entity--solution--listing-item.html.twig
```diff
@@ -25,10 +25,12 @@
       </div>
     {% endif %}
     <div class="listing__field listing__stats mdl-card__actions">
-      <div class="listing__stat">
-        <span class="listing__text">{{ content.downloads.0 }}</span>
-        <span class="listing__icon icon icon--download"></span>
-      </div>
+      {% if content.downloads.0['#markup'] >= 1  %}
```
claudiu-cristea: This add business logic into the theme layer. I'm not very happy. I would suggest that we check here for `content.downloads`:

```
{% if content.downloads  %}
```
And we simply don't add the `#markup` on `solution_rdf_entity_view()` if `$total` equals 0.

web/themes/joinup/templates/form/container--stats-topic.html.twig
```diff
@@ -18,10 +18,12 @@
  */
 #}
 {% if children is not empty %}
-  <div class="stats__column">
-    <div class="stats__wrapper">
-      <span class="stats__text">{{ topic }}:</span>
-      <span class="stats__text bold">{{ children }}</span>
+  {% if children.__toString() >= 1 %}
```
claudiu-cristea: Can you explain this change?

tests/features/solution/add_solution.feature
```diff
@@ -247,3 +247,35 @@ Feature: "Add solution" visibility options.
...
+    # Go back to the solution tiles page and check downloads are disappeared.
+    Then I visit "/solutions"
+    Then I should not see a "icon--download" element
+    # Delete all created dummy data.
+    When I delete the "Sample distribution 2" asset distribution
+    And I delete the "Climate change tracker" solution
+    And I delete the "Ocean studies" collection
```
claudiu-cristea: You don't need the last 3 lines. Behat automatically cleans up the entities created via API. Only if you create via UI, you'll need to take care.

tests/features/solution/add_solution.feature
```diff
@@ -247,3 +247,35 @@ Feature: "Add solution" visibility options.
...
+    # Check the same thing on the homepage of the solutions.
+    Then I should see the text "Downloads: 10"
+    When I delete the "Sample distribution" asset distribution
+    Given the cache has been cleared
+    And I reload the page
+    # Check it again after the first distribution is deleted and cache has been cleared.
+    Then I should not see the text "Downloads: 10"
+    # Go back to the solution tiles page and check downloads are disappeared.
+    Then I visit "/solutions"
+    Then I should not see a "icon--download" element
```
claudiu-cristea: After this I would love to see adding a distribution with downloads > 0 and check again.

tests/features/solution/add_solution.feature
```diff
@@ -247,3 +247,35 @@ Feature: "Add solution" visibility options.
     Then I delete the "Climate change tracker" solution
     And I delete the "Climate change tracker" solution
     And I delete the "Angela Crespi" contact information
+
+  Scenario: Check solution downloads counter
+    # Create the dummy data to work with.
+    Given the following collection:
+      | title              | Ocean studies |
```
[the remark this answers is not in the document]
junkuncz: Okay!

tests/features/solution/add_solution.feature
```diff
@@ -247,3 +247,35 @@ Feature: "Add solution" visibility options.
...
+    Given the following solution:
+      | title       | Climate change tracker |
```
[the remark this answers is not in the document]
junkuncz: Okay!

tests/features/solution/add_solution.feature
```diff
@@ -247,3 +247,35 @@ Feature: "Add solution" visibility options.
...
+      | description | Atlantic salmon arrived after the Little Ice Age. |
+      | collection  | Ocean studies                                     |
+      | state       | validated                                         |
+    Given the following distributions:
+      | title        | description | parent | downloads |
```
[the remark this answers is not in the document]
junkuncz: Okay!

tests/features/solution/add_solution.feature
```diff
@@ -247,3 +247,35 @@ Feature: "Add solution" visibility options.
...
+      | Sample distribution | Sample description | Climate change tracker | 10 |
+      | Sample distribution 2 | Sample description 2 | Climate change tracker | 0 |
+    # Checks solutions tiles to we have the correct numbers.
+    Then I visit "/solutions"
+    Then I should see the text "10"
```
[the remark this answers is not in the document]
junkuncz: Okay, changed!

tests/features/solution/add_solution.feature
junkuncz: Removed!

tests/features/solution/add_solution.feature
junkuncz: What do you mean exactly? Add it like in the "Add a distribution to a solution as a facilitator." scenario or with "Given the following distribution..." and do a check then?

claudiu-cristea: > Given the following distribution ...

This should work, I guess.

web/themes/joinup/templates/rdf_entity/rdf-entity--solution--listing-item.html.twig
junkuncz: Fixed!

tests/features/solution/add_solution.feature
```diff
@@ -247,3 +247,35 @@ Feature: "Add solution" visibility options.
...
+    When I go to the homepage of the "Climate change tracker" solution
+    # Check the same thing on the homepage of the solutions.
+    Then I should see the text "Downloads: 10"
+    When I delete the "Sample distribution" asset distribution
+    Given the cache has been cleared
```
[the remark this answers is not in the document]
junkuncz: Okay I added the "short term solution". What do you mean by follow up? Create a Jira ticket or something?

web/themes/joinup/templates/form/container--stats-topic.html.twig
junkuncz: Well, since we changed the logic of the total downloads counter in solution_rdf_entity_view() this does not need anymore. Reverted!

tests/features/solution/add_solution.feature
junkuncz: I've commited some changes but it's still not perfect.
When I run this scenario getting an error: **Tile 'Test solution' not found.**
I assume that it will be a caching issue again, because if I clear the cache before I visit /solutions page then it's fine but not a valid use case as you described above somewhere.
Any ideas? Did I do something wrong?

web/modules/custom/solution/solution.module
```diff
@@ -260,8 +264,13 @@ function solution_rdf_entity_view(array &$build, EntityInterface $entity, Entity
     $build['downloads'] = [
       '#type' => 'container',
       '#stats_topic' => t('Downloads'),
-      0 => ['#markup' => (string) $total],
     ];
+    // Only add #markup if we want to show it.
+    if ($total >= 1) {
+      $build['downloads'][0]['#markup'] = (string) $total;
```
claudiu-cristea: We can omit 0 and do `$build['downloads'][] = ['#markup' => $total]`. Also I don't see why we need the string casting.

tests/features/solution/add_solution.feature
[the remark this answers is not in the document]
claudiu-cristea: Yes. Ticket in Jira and linkd from the current ticket as "follow-up"

web/modules/custom/solution/solution.module
junkuncz: True! Actually I also don't understand the string casting just it was there.. Will remove it!

tests/features/solution/add_solution.feature
```diff
@@ -247,3 +247,41 @@ Feature: "Add solution" visibility options.
...
+    # Check the same thing on the homepage of the solutions.
+    Then I should see the text "Downloads: 10"
+    When I delete the "Sample distribution" asset distribution
+    And I reload the page
+    # Check it again after the first distribution is deleted and cache has been cleared.
+    Then I should not see the text "Downloads: 10"
+    # Go back to the solution tiles page and check downloads are disappeared.
+    Then I visit "/solutions"
+    Then I should not see a "icon--download" element
+    Given the following solution:
```
claudiu-cristea: Also, we don't need a new solution. Let's add the distribution to the existing solution "Climate change tracker"

web/themes/joinup/templates/rdf_entity/rdf-entity--solution--listing-item.html.twig
```diff
@@ -25,10 +25,12 @@
       </div>
     {% endif %}
     <div class="listing__field listing__stats mdl-card__actions">
-      <div class="listing__stat">
-        <span class="listing__text">{{ content.downloads.0 }}</span>
-        <span class="listing__icon icon icon--download"></span>
-      </div>
+      {% if content.downloads  %}
+        <div class="listing__stat">
+          <span class="listing__text">{{ content.downloads.0 }}</span>
```
claudiu-cristea: We don't need the `.0` suffix here. Let's keep it clean:
```html
<span class="listing__text">{{ content.downloads }}</span>
```

tests/features/solution/add_solution.feature
junkuncz: Removed it!

web/modules/custom/solution/solution.module
junkuncz: Removed!

web/themes/joinup/templates/rdf_entity/rdf-entity--solution--listing-item.html.twig
junkuncz: Removed!

web/themes/joinup/templates/rdf_entity/rdf-entity--solution--listing-item.html.twig
junkuncz: 0 also removed!

tests/features/solution/add_solution.feature
```diff
@@ -247,3 +247,36 @@ Feature: "Add solution" visibility options.
...
+    Then I should not see the text "Downloads: 10"
+    # Go back to the solution tiles page and check downloads are disappeared.
+    Then I visit "/solutions"
+    Then I should not see a "icon--download" element
+    Given the following distribution:
+      | title       | Sample distribution 3  |
+      | description | Sample description 3   |
+      | parent      | Climate change tracker |
+      | downloads   | 20                     |
+    Then I should see the text "20" in the "Test solution" tile
```
claudiu-cristea: The first problem here is that, after adding the new distribution, you should reload the page. Otherwise  how can you see the changes?

The 2nd is that you are trying to see the downloads in a tile that doesn't exist: s/Test solution/Climate change tracker.

However, I confirm that the test fails is randomly on the last line. Seems that still a caching issue.

junkuncz: Well I think we should resolve the caching thing and then we can adopt the tests as well.

pfrenssen:
The download icon is still showing in the tile even though the download count is correctly hidden.

pfrenssen:
There was indeed a problem with the caching of this element. In fact there were two problems:

1. The cache tags for the download counter were attached to the download counter itself, but in the template there was some logic to exclude the download counter from being rendered if the count is zero. This means that if there are no downloads the cache tags were not included in the page, and the page would not be invalidated if a download is added later.
2. An empty cacheability metadata object was instantiated using `new CacheableMetadata()`, instead of starting from the existing metadata using `CacheableMetadata::createFromRenderArray()`. This was fine at the moment since the element did not have any other metadata yet, but we should always use the existing metadata anyway. Otherwise we will get problems later in case the element gets altered.

I also found that there was another caching issue causing a failure, but this was in the Behat scenario rather than in the code that takes care of the rendering of the download counter. When we are demonstrating caching of an element in a user scenario and are relying on the Drupal API for the actions that cause the invalidation of the cache, we need to add the `@clearStaticCache` tag to the scenario. This is because of a performance optimization that is present in the cache layers. This performance optimization is present in both of the caching layers we use for cache tags (`\Drupal\redis\Cache\RedisCacheTagsChecksum::invalidateTags()` and `\Drupal\Core\Cache\DatabaseCacheTagsChecksum::invalidateTags()`):

```
// Only invalidate tags once per request unless they are written again.
if (isset($this->invalidatedTags[$tag])) {
  continue;
}
```

This works fine in the normal use case of Drupal: a PHP process is started to handle a request, return a single web page, and then the process is terminated, after running typically less than a second. However when we are executing Behat scenarios the situation is very different: the "request" is now a long running process that may last for several minutes or even hours. If a tag gets invalidated once in a scenario, this code will prevent it from being invalidated again. The `@clearStaticCache` tag works around this by flushing Drupal's static caches after every step in the scenario.

I also optimized the code a bit. The loop that counts the total of downloads had the following unneeded line in it:

```
$cacheable_metadata->addCacheableDependency($distribution);
```

This was adding each of the solution's distributions to the cache tags, but this is not necessary. The distributions are group content, and they are already included in the `og-group-content` tag.

The caching works good now, but I am not sure if we should use it. Before this PR there was no invalidation happening when a download occurred, so the page stayed nicely cached until an actual content change occurs in one of the solutions being shown. Now the situation will be very different: every single time a download occurs, the element will be invalidated, and needs to be rendered again in the next request. Since the solution overview shows our most popular solutions this might mean that the page gets invalidated very regularly. I also noticed that the calculation of the download counter is a _very_ expensive operation; it loads the full entities of all distributions that belong to the solution. In my limited testing with a handful of distributions this was already causing a significant slowdown. I imagine with solutions that have a large number of distributions this will be very slow.

I'm afraid that this change might cause a huge performance hit on the solutions page. This needs some profiling and testing using the production database before it can be deployed.

pfrenssen:
This was discussed in a meeting and in scope of this ticket we want to just focus on hiding the download counter. We will move the fixing of the caching into a followup.

pfrenssen:
Created followup to handle the caching issue: https://github.com/ec-europa/joinup-dev/pull/1579

tests/features/solution/add_solution.feature
pfrenssen: This step definition is not understandable for our business stakeholders:

```
Then I should not see a "icon--download" element
```

Our Behat scenarios are intended so our business stakeholders and functional contractors can understand and validate the work that we have done. Not all these people are highly technically skilled and we cannot assume that they will understand what an `icon--download element` is and that they will known how to look for this in the DOM. We need to rewrite this in non-technical language so this will not be rejected in UAT. I would suggest to use an easy to understand step such as:

```
Then the download icon should not be shown in the "Climate change tracker" tile
```

Here is how a step like this could be implemented:

```
/**
 * @Then the download icon should not be shown in the :heading tile
 */
public function assertTileNotContainsDownloadIcon(string $heading): void {
  $tile = $this->getTileByHeading($heading);
  $elements = $tile->findAll('xpath', '//span[contains(@class, "icon--download")]');
  Assert::assertEmpty($elements);
}
```

tests/features/solution/add_solution.feature
```diff
@@ -247,3 +247,41 @@ Feature: "Add solution" visibility options.
     Then I delete the "Climate change tracker" solution
     And I delete the "Climate change tracker" solution
     And I delete the "Angela Crespi" contact information
+
+  @clearStaticCache
+  Scenario: Check solution downloads counter
```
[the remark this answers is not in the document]
junkuncz: Okay I've moved it to a different file inside the solution folder. I that okay or you want to place to a more general place like joinup_core?

tests/features/solution/add_solution.feature
```diff
@@ -247,3 +247,41 @@ Feature: "Add solution" visibility options.
...
+      | title       | Climate change tracker                            |
+      | description | Atlantic salmon arrived after the Little Ice Age. |
+      | collection  | Ocean studies                                     |
+      | state       | validated                                         |
+    Given the following distributions:
+      | title                 | description          | parent                 | downloads |
+      | Sample distribution   | Sample description   | Climate change tracker | 10        |
+      | Sample distribution 2 | Sample description 2 | Climate change tracker | 0         |
+    # Checks solutions tiles to we have the correct numbers.
+    Then I visit "/solutions"
```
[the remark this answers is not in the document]
junkuncz: Done!

tests/features/solution/add_solution.feature
```diff
@@ -247,3 +247,41 @@ Feature: "Add solution" visibility options.
...
+    Then I should see the text "10" in the "Climate change tracker" tile
```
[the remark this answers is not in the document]
junkuncz: Implemented in solution.behat.inc. Is that okay or this also needs to be placed to a general place?

tests/features/solution/add_solution.feature
junkuncz: Implemented!

tests/features/solution/add_solution.feature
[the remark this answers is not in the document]
pfrenssen: This has been fixed in scope of this issue.

web/themes/joinup/templates/rdf_entity/rdf-entity--solution--listing-item.html.twig
pfrenssen: I checked and this is fixed now, thanks!

web/themes/joinup/templates/rdf_entity/rdf-entity--solution--listing-item.html.twig
pfrenssen: Looking good now!

tests/features/solution/add_solution.feature
pfrenssen: This is working fine now and the cache clearing was fixed but now moved to a follow up ticket because it might invalidate the page too often.

tests/features/solution/add_solution.feature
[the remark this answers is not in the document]
pfrenssen: No this is great, it is checking the download counter on solutions after all.

tests/features/solution/add_solution.feature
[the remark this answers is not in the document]
pfrenssen: No this is fine since for the moment these downloads are only shown on solutions.

PR #1563, opened 2019-03-22
ISAICP-5132: Provide an SPDX Licence model.

idimopoulos:


claudiu-cristea:
Also the ticket description states "and make the license module depend on this.". But this was not achieved.

PR #1564, opened 2019-03-22
Isaicp 5161

junkuncz:


web/modules/custom/joinup_core/joinup_core.module
```diff
@@ -490,6 +490,27 @@ function joinup_core_entity_create(EntityInterface $entity) {
   }
 }
 
+/**
+ * Implements hook_entity_insert().
+ */
+function joinup_core_entity_insert(EntityInterface $entity) {
```
claudiu-cristea: We should not mix UI with API. The message should be triggered only in UI context, but `hook_entity_insert()` is a general API hook. This means the message will be triggered even a module is adding a collection/solution via code or when we import/federate solutions. We don't want this message to be displayed in such conditions. That's why we should hook in the form API flow rather than in the entity API. Probably, you'll need to add a new submit callback in `joinup_core_form_rdf_entity_form_alter()`, in certain conditions (i.e. check the bundle) and popup the message in that callback.

web/modules/custom/joinup_core/joinup_core.module
```diff
@@ -490,6 +490,27 @@ function joinup_core_entity_create(EntityInterface $entity) {
...
+  // List the related entity bundles.
+  $related_bundles = [
+    'collection',
+    'solution',
+  ];
+
+  // Check to the current bundle is in our list.
+  if (in_array($entity->bundle(), $related_bundles)) {
```
claudiu-cristea: Because you've implemented `hook_entity_insert()` this check is not consistent check. It should:

* Either, check also if `$entity->getEntityTypeId() === 'rdf_entity'`
* Or, use `hook_ENTITY_TYPE_insert()`, in this case `joinup_core_rdf_entity_insert()`,

Of course, we don't have a node-type named `collection` but we have be safe.

web/modules/custom/joinup_core/joinup_core.module
```diff
@@ -490,6 +490,27 @@ function joinup_core_entity_create(EntityInterface $entity) {
...
+    /** @var \Drupal\Core\Messenger\MessengerInterface $messenger */
+    $messenger = \Drupal::service('messenger');
+
+    // Add a status message when it was saved.
+    $messenger->addStatus(t('Thank you for proposing a @entity_type. 
+    Your request is currently pending approval by the site administrator.', ['@entity_type' => $entity->bundle()]));
```
claudiu-cristea: As our Drupal core benefits for this patch https://www.drupal.org/project/drupal/issues/2765065, we should use the lower-cased singular form of the bundle label. `$entity->bundle()` is more a "machine name", we have now a dedicated label on the bundle entity: In this case we should use `$entity->get('rid')->entity->getSingularLabel()`, which is the "human readable" singular label to be used on UI.

web/modules/custom/joinup_core/joinup_core.module
junkuncz: Alright! I've implemented it.

web/modules/custom/joinup_core/joinup_core.module
junkuncz: Since this logic is placed in joinup_core_form_rdf_entity_form_alter() I don't think we really need to check the EntityTypeID because this function is called only when it's an rdf entity itself.

web/modules/custom/joinup_core/joinup_core.module
junkuncz: Okay! I added these kind of things into the submit callback (_joinup_core_message_handler).
Please check!

web/modules/custom/joinup_core/joinup_core.module
```diff
@@ -152,6 +163,23 @@ function joinup_core_form_rdf_entity_form_alter(&$form, FormStateInterface $form
 
 }
 
+/**
+ * Displays status message after the user saved the related RDF content.
+ */
+function _joinup_core_message_handler(&$form, FormStateInterface $form_state) {
```
claudiu-cristea: It's not wrong but looks so vintage, with Drupal 6 or 7 flavor, to prefix functions with underscore character :)

Also I would suggest that we use a more specific function name. For me `_joinup_core_message_handler()` is to vague/generic. What about `joinup_core_propose_message_submit()`?

web/modules/custom/joinup_core/joinup_core.module
```diff
@@ -152,6 +163,23 @@ function joinup_core_form_rdf_entity_form_alter(&$form, FormStateInterface $form
...
+  /** @var \Drupal\rdf_entity\Form\RdfForm $form_object */
+  $form_object = $form_state->getFormObject();
+  /** @var \Drupal\rdf_entity\RdfInterface $rdf */
+  $rdf = $form_object->getEntity();
+
+  /** @var \Drupal\Core\Messenger\MessengerInterface $messenger */
+  $messenger = \Drupal::service('messenger');
```
claudiu-cristea: We have a dedicated method: `\Drupal::messenger()`.

web/modules/custom/joinup_core/joinup_core.module
```diff
@@ -152,6 +163,23 @@ function joinup_core_form_rdf_entity_form_alter(&$form, FormStateInterface $form
...
+
+  // Add a status message when it was saved.
+  $messenger->addStatus(t('Thank you for proposing a @entity_label.
```
claudiu-cristea: Let's not wrap the message in this way. Better on a single line.

web/modules/custom/joinup_core/joinup_core.module
junkuncz: Yea this was a bit ugly I changed it. :)

web/modules/custom/joinup_core/joinup_core.module
junkuncz: Changed!

web/modules/custom/joinup_core/joinup_core.module
junkuncz: Changed!

junkuncz:
> Problems:
> 
> * There are privileged users, such as moderators, that are allowed to directly publish new collections/solutions, without being necessary, an approval. If such a user creates a collection/solution by pressing the PUBLISH button he/she should not receive this message as the entity is already published. This should be proved in tests.
> * If I edit an existing collection/solution, it will show the every time I save the entity but the requirements are clear, to show only on new entities.

I've added an extra check in form_alter() to only add the related callback when it's on the "add content" form.

tests/features/collection/propose.feature
```diff
@@ -62,6 +62,7 @@ Feature: Proposing a collection
     And I should see the heading "Ancient and Classical Mythology"
     And I should see a logo on the header
     And I should see a banner on the header
+    And I should not see "Thank you for proposing a collection. Your request is currently pending approval by the site administrator."
```
idimopoulos: This is a nice spot to assert that the message is not shown as we save the entity as "Save as draft".
However, the file is called propose.feature and the scenario mentions that we are proposing a collection.

This is not a problem with your implementation, it seems however that the test is not working as expected.
Can you fix the scenario in total? SImply instead of saving as draft, try to propose the collection instead.

tests/features/solution/add_solution.feature
```diff
@@ -137,6 +137,7 @@ Feature: "Add solution" visibility options.
     And I press "Add owner"
     And I press "Propose"
     Then I should see the heading "V60 filter coffee solution"
+    And I should see "Thank you for proposing a solution. Your request is currently pending approval by the site administrator."
```
idimopoulos: This is fine but this is asserted on the second time the solution is proposed. After it was published.
Can we assert it also during initial proposal?

tests/features/collection/propose.feature
idimopoulos: Sorry, but now that you fixed the form submit, I noticed that this is actually not working. Tested it through the UI.
The message is not shown when proposing a collection but should according to the requirements. So this step needs to be changed to a "And I should see...." and be fixed.

junkuncz: Form is fixed and tests also fixed.

tests/features/solution/add_solution.feature
junkuncz: done!

PR #1567, opened 2019-03-25
Isaicp 5133: Deny access to the canonical route of the licences.

idimopoulos:


web/profiles/joinup/src/Access/CanonicalRouteRestrict.php
```diff
@@ -0,0 +1,45 @@
...
+   *   The route match object to be checked.
+   * @param \Drupal\Core\Session\AccountInterface $account
+   *   The account being checked.
+   *
+   * @return \Drupal\Core\Access\AccessResultInterface
+   *   The access result.
+   */
+  public function access(RouteMatchInterface $route_match, AccountInterface $account): AccessResultInterface {
+    $rdf_entity = $route_match->getParameter('rdf_entity');
+    if (!($rdf_entity instanceof RdfInterface) || $rdf_entity->bundle() !== 'spdx_licence') {
```
claudiu-cristea: I can see the benefit of having a generic entity canonical route blocker. Unfortunately, we break this promise in this line because we hardcode here the entity type and the bundle. Sad. With his hardcoding we should move the whole logic in the `joinup_licence` module as is no more a _generic solution_.

However, this has challenged me. What if...

* In `\Drupal\joinup\Routing\RouteSubscriber::alterRoutes()`, instead of hardcoding the requirement, we use this pattern: `'_canonical_route_restrict__' . $entity_type_id`. Where `$entity_type_id` is to be computed from the route name or other route data (didn't investigated).
* The service `joinup.canonical_route_restrict` is moved into `joinup_licence` module as:
  ```yaml
   joinup_licence.spdx_entity_route_restrict:
   class: Drupal\joinup\Access\CanonicalRouteRestrict
   arguments: ['@current_route_match', '@current_user', rdf_entity, spdx_licence]
   tags:
     - { name: access_check, applies_to: _canonical_route_restrict__rdf_entity }
   ``` 
* The access method would look like:
  ```php
  public function access(RouteMatchInterface $route_match, AccountInterface $account, string $entity_type_id, string $bundle === NULL) {
    $entity = $route_match->getParameter($entity_type_id);
    if ($entity instanceof ContentEntityInterface && ($bundle === NULL || $bundle === $entity->bundle()) {
       return ....restricted
    }
    ...
  }
  ```

Maybe is a crazy idea, I don't know if really works but it would keep the solution generic. The alternative is to move the whole logic in `joinup_licence`.

web/profiles/joinup/src/Access/CanonicalRouteRestrict.php
```diff
@@ -0,0 +1,45 @@
...
+      // In case the route the canonical route of the spdx_licence, return a
+      // positive access result as a neutral would deny the access to the route.
+      // @see \Drupal\Core\Access\AccessResult::andIf().
+      return AccessResult::allowed()->addCacheContexts(['user']);
+    }
+
+    if ($account->id() === 1) {
```
claudiu-cristea: I keep quoting myself (for how many times?):

> We should not perform access checks, neither of role name, nor on user ID. We should always check for permissions.

That said, we can add permission `bypass entity canonical route restriction` (this could be made more granular, on entity type, in the future) and do the check on `$account->hasPermission(...)`.

Note that the UID1 special behaviour will be removed in https://www.drupal.org/project/drupal/issues/540008

web/profiles/joinup/src/Access/CanonicalRouteRestrict.php
```diff
@@ -0,0 +1,45 @@
...
+      return AccessResult::allowed()->addCacheContexts(['user']);
```
claudiu-cristea: Using the `user` cache context here is wrong because this would vary the cache too much, basically will create a cache version for each user hitting this route. Instead, `user.is_super_user` should have been used. But if we switch to the ăroposed permission check, the it can be `user. permissions`. See https://www.drupal.org/docs/8/api/cache-api/cache-contexts.

There's also another issue with the cache contexts: shouldn't we vary also on route name (`route` context)? We should test if a user denied a SPDX licence route will not be denied also on a different bundle page.

web/profiles/joinup/src/Access/CanonicalRouteRestrict.php
idimopoulos: I copied this over from our UID1Only access handler. I am going to remove it entirely. No user will have access to the canonical route. Or do you think there is any value for that?

web/profiles/joinup/src/Access/CanonicalRouteRestrict.php
idimopoulos: Let's move this Access class to the `joinup_licence` module. It gets confusing to expand this like this because the main problem that this was built like this is that we don't get the parameters (like an actual rdf_entity entity object) in the route subscriber - since it works directly on routes and not paths.
Also, I will have to move the assignment of the route requirement to the licence so that it is independent as a module.

PR #1574, opened 2019-03-31
ISAICP-5147: Licence canonical page

pfrenssen:


tests/features/joinup_licence/add_licence.feature
```diff
@@ -1,43 +1,59 @@
 @api
-Feature: Add licence through UI
+Feature: Add a licence through the UI
   In order to manage licences
-  As a moderator or a license manager
+  As a moderator or a licence manager
```
idimopoulos: :+1:

web/modules/custom/joinup_licence/joinup_licence.module
```diff
@@ -32,3 +44,72 @@ function joinup_licence_rdf_entity_access(EntityInterface $entity, $operation, A
...
+
+/**
+ * Implements hook_entity_extra_field_info().
+ *
+ * - Append additional fields to the licence entity display containing data
+ *   from the SPDX licence.
+ *
+ * @see joinup_licence_rdf_entity_view()
+ */
+function joinup_licence_entity_extra_field_info(): array {
```
idimopoulos: I have a question on this extra fields here concerning best practices. This is to allow data from the spdx licence to be viewed in the licence page. However, only the canonical route of the spdx_licence is disallowed access which means that the spdx licence can be viewed anywhere else.
Why is it that we don't simply render the spdx_licence entity in a view mode that shows these fields and display this view mode as part of the reference field field_licence_spdx_licence?

web/modules/custom/joinup_licence/joinup_licence.module
```diff
@@ -32,3 +44,72 @@ function joinup_licence_rdf_entity_access(EntityInterface $entity, $operation, A
...
+  return $fields;
+}
+
+/**
+ * Implements hook_ENTITY_TYPE_view().
+ *
+ * - Exposes a number of fields from the SPDX Licence entity on the Licence
+ *   entity.
+ */
+function joinup_licence_rdf_entity_view(array &$build, RdfInterface $entity, EntityViewDisplayInterface $display, string $view_mode): void {
```
idimopoulos: This whole function can be removed if the above comment changes the approach.

web/modules/custom/joinup_licence/README.md
```diff
@@ -0,0 +1,13 @@
+Joinup Licence
```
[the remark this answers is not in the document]
pfrenssen: This shows which graph to use to import the licences in Joinup. If we don't have this here then we are forcing people to go dig in the config files for this information.

We cannot put this in the SPDX module since the graph `http://joinup.eu/spdx_licence/published` is specific to Joinup and doesn't make sense in any other project that uses the SPDX module.

web/modules/custom/joinup_licence/joinup_licence.module
pfrenssen: We need to show a number of fields from the Licence entity as well as a number of fields from the SPDX Licence entity. If I would just render the reference field to the SPDX entity using a view mode then it becomes more difficult to theme this:
* If a change to the theming is necessary then the themer needs to adapt 2 view modes instead of one.
* It would not be possible to reorder the fields freely in Field UI, for example it would not be possible to move a field from the SPDX Licence entity in between two fields of the Licence entity.
* There will be additional containers around the SPDX Licence fields which will make it necessary to write different CSS selectors for the Licence fields and the SPDX licence fields.

By embedding them like this as extra fields maximum flexibility is given to the themer to make any design possible by allowing them to position the fields freely.

2019-04-08-173347_3520x1080_scrot

idimopoulos: The thing with the theming is that when it comes to custom templates, it is already extremely flexible. The variables array already contains both the full view of the entity and also the content where each field is contained separately in a renderable array.

In case the themer wants to highly change the way the entity looks when viewing it, he can provide an html and then simply print the fields separately. These fields can either be retrieved as raw value or as a renderable array controlled by the view mode (e.g. the title display, prefixes, suffixes etc).

The same applies for sub-entities (referenced entities) and it makes it much cleaner to simply render a view mode of the entity, because it will be a template included in the mother template.
So in this specific case, Tom for example, will theme the sub entity in its specific view mode as a separate entity because it is cleaner and easier to maintain. Then, in the parent template (the licence parent), he will only need to print the field and everything else will be done automatically. 

And last, even if it was that for some reason extra fields are required, according to what I remember from Francesco, in case of theming, they should be generated and included in the preprocess hook so that they are cached along with the theme.

PR #1577, opened 2019-04-04
Isaicp 5095

junkuncz:


web/modules/custom/joinup_notification/src/JoinupMessageDelivery.php
```diff
@@ -117,7 +117,7 @@ public function sendMessageTemplateToUser(string $message_template, array $argum
     $recipients_metadata = [
       [
         'options' => $notifier_options,
-        'notifier' => $digest ? $this->getNotifierId($account) : 'email',
```
idimopoulos: We are removing the 'immediatelly' option and yet this somehow removes the digestion if I am reading it correctly? 
Shouldn't it be `'notifier' => $digest ? $this->getNotifierId($account) : 'email', ` instead? or am I missing something?

idimopoulos: Sorry, I meant "shouldn't it be `'notifier' => $this->getNotifierId($account),` instead?"

junkuncz: You are right! Fixed it.

PR #1578, opened 2019-04-08
Isaicp 5141:  Implement the actual JLA.

idimopoulos:


web/modules/custom/joinup_communities/eupl/eupl.behat.inc
```diff
@@ -0,0 +1,80 @@
...
+  /**
+   * Creates the standard 'EUPL' collection and several dependencies.
+   *
+   * @throws \Drupal\Core\Entity\EntityStorageException
+   *   Thrown when one of the entities could not be created, for example because
+   *   it already exists.
+   *
+   * @beforeScenario @eupl
+   */
+  public function setupTallinnData() {
```
sandervd: Copy-pasta?

scripts/update_uat.sh
```diff
@@ -12,10 +12,13 @@ cd ${PROJECT_ROOT}
 echo "Disabling config_readonly."
 touch disable-config-readonly
 
-./vendor/bin/drush cache:clear bin config --yes
-./vendor/bin/drush updatedb --yes
-./vendor/bin/drush cs-update --discard-overrides --yes
-./vendor/bin/drush cache-rebuild --yes
+./vendor/bin/drush cache:clear bin config --yes &&
```
sandervd: We don't need to concatenate this with the &&, as the -e parameter makes sure that it fails on the first error

web/profiles/joinup/src/Plugin/Block/LicenceFilterBlock.php
```diff
@@ -0,0 +1,99 @@
...
+      if (empty($parent)) {
+        $tree[$term->tid]['title'] = $term->name;
+        $tree[$term->tid]['class'] = 'licence-filter--' . strtolower($term->name);
+      }
+      else {
+        $child = $storage->load($term->tid);
+        $tree[$parent]['items'][] = [
+          'title' => $term->name,
+          'description' => $child->getDescription(),
+          'licence_category' => str_replace([' ', '/'], ['-', '-'], strtolower($term->name)),
```
sandervd: This might be a bit risky?

web/themes/joinup/joinup_theme.theme
```diff
@@ -1560,6 +1578,56 @@ function joinup_theme_preprocess_rdf_entity__collection__listing_item(&$variable
...
+ * Provides variables needed for the search_list_item view mode of licences.
+ */
+function joinup_theme_preprocess_rdf_entity__licence__search_list_item(&$variables) {
+  $entity = $variables['elements']['#rdf_entity'];
+  $variables['licence_title'] = $entity->label();
+  $variables['licence_description'] = $entity->field_licence_description->value;
+
+  // Calculate variables related to the spdx licence.
+  if (!empty($entity->field_licence_spdx_licence->getValue())) {
+    $spdx_licences = $entity->field_licence_spdx_licence->referencedEntities();
```
sandervd: Would it make sense to define this as extra fields, so it can be used across view modes?

web/profiles/joinup/src/Plugin/Block/LicenceFilterBlock.php
idimopoulos: Nahh, I think it is fine. The terms are controlled by our legal-type taxonomies. I am merely removing some blank spaces and slashes. However, just to be sure, I enclosed the values also with the `htmlentities` function. Just to be sure.

web/themes/joinup/joinup_theme.theme
idimopoulos: Again, I am not very fond of that idea.. This is how it looks to have extra fields right now in the entity display builder:
image
The code that creates these extra fields is the following (just a sample):
```
$display['spdx_licence_label'] = [
      'label' => t('SPDX licence label'),
      'description' => t('The label of the licence in the SPDX licence list.'),
      'visible' => FALSE,
    ];
    $display['spdx_licence_id'] = [
      'label' => t('SPDX licence ID'),
      'description' => t('The ID of the licence in the SPDX licence list.'),
      'visible' => FALSE,
    ];
```
I understand that creating these fields as extra fields in theory gives us the flexibility to set them anywhere in the view, but in practice, (unless we were to create a full blown solution that allows to position the fields of the referenced entity instead of the entity itself), is quite limiting..
These extra fields are defined and can be used across the view modes but there are implications.
There are no settings defined in the fields. Regardless of the type of the field, the settings are either hardcoded in the extra fields hook, or we anyway need to pre-process the fields to alter them.
The current state sets the fields to have the same view across all modes. Extra preprocessing is required to alter per view mode. That was solved up to now by allowing the view mode to be set on the referenced entity as a setting on the display mode.

So in general, that is an amazing idea as a contrib module. But normally it would require to allow the user to choose the view mode of the referenced entity and then replace it with the fields *and then* also allow the settings to be set on each of the fields. The idea that the user would either be able to select which view mode of the referenced entity would be used for the field displayed, or set the settings of the field as if it was a native field.

However, in our case, I still don't see the added value. We are creating extra fields that we can move around in the display mode. Then in the `joinup_licence_rdf_entity_view` we are loading the referenced entities with `$spdx_licence = $entity->get('field_licence_spdx_licence')->referencedEntities()[0] ?? NULL;` (exactly like we do in the preprocess function) and we are populating these fields by calling the `::view` method of their itemList. So basically, we are rendering the field according to the settings of the default view mode of the referenced entity and this is not configurable. 
Further more, in our specific case, we used these fields to show the values in the UI, and then Tom, created a template to group them together just like he would do with the actual entity referenced. So, the flexibility (which don't get me wrong, it is indeed quite handy to be able to move fields around in the display) was overridden so that the result is the default behaviour.
As far as the view mode that we are using in the JLA is concerned, we are overriding the default template anyway, so the only thing that these extra fields do is making them available for viewing. Which is again what we do in the preprocess function.
So, unless we are using something like display suite in order to build our views, or we are building a full blown solution where everything is customizable, I think it mostly restricts us to a specific view mode here.
Keep in mind, that on the theming level, we are normally using the `hook_preprocess_<template_suggestion>` function to override and create variables necessary for a specific view mode. Which makes the whole thing specific. The data are then stored in the template level and cached as such.

In any case, I created the field as you have requested but I merely extended the fields to also include the spdx label. I still generate the rest of the variables e.g. the spdx_licence_id as I need the raw value and not the rendered field. Also, the spdx licence URL is simply the ID of the entity along with an anchor in the end so there is no use in creating a field for that. I could have hardcoded it in the template as well.

pfrenssen: > Further more, in our specific case, we used these fields to show the values in the UI, and then Tom, created a template to group them together just like he would do with the actual entity referenced. So, the flexibility (which don't get me wrong, it is indeed quite handy to be able to move fields around in the display) was overridden so that the result is the default behaviour.

This is the key to the story. The theme should not override the positioning and visibility of elements. These are determined by the view modes. We should be able to switch to a different theme and keep the same basic structure.

idimopoulos: Yeah, still, I have so many thoughts on this that each of this is a discussion on its own.. From the fact that Drupal is trying to officially support headless but for some reason the view modes are supposed to control the positioning of the fields, to the fact that if we render the page in e.g. Angular, the Drupal html elements should make any sense or the weight or the position  whatsoever, and so many others.
Let's agree to disagree.. 
I would gladly accept some kind of a formatter for the referenced entity that would allow to replace the entity with the entity fields of the view mode selected with all settings available but it seems a bit weird to manually create extra fields each time we want to have a custom value to the frontend.
In total, I am fine with having an agreement against the tech leader's decision over it so I went on with this solution myself.

PR #1582, opened 2019-04-09
ISAICP-3599: Fix caching of collections

pfrenssen:


tests/src/Traits/PageCacheTrait.php
```diff
@@ -0,0 +1,82 @@
...
+
+  /**
+   * Returns whether the current page is cacheable.
+   *
+   * @return bool
+   *   TRUE if the current page is served from the cache, or eligible to be
+   *   served from the cache on subsequent requests.
+   */
+  protected function isPageCacheable(): bool {
+    return $this->isPageCached() || in_array('MISS', $this->getPageCacheStatus());
```
claudiu-cristea: Nice!

web/profiles/joinup/joinup.services.yml
```diff
@@ -1,4 +1,9 @@
 services:
+  cache_context.user.is_root:
```
claudiu-cristea: We don't need this context. Core always provide `cache_context.user.is_super_user`.

pfrenssen: Wow I didn't know about this one :sweat_smile: I see the implementations are virtually identical but mine has more strict typing :grin: 

I'll replace my implementation with the one provided by core.

PR #1584, opened 2019-04-10
ISAICP-5148: Basic styling for licence page

tom-waat:


idimopoulos:
Changed the base since the parent ticket was merged.

PR #1591, opened 2019-04-16
ISAICP-5125: Update dependencies and fix new coding style warnings

pfrenssen:


web/modules/custom/contact_form/src/EventSubscriber/SenderNotificationSubscriber.php
```diff
@@ -54,7 +54,7 @@ public static function getSubscribedEvents(): array {
   public function notifySender(NotificationEvent $event): void {
     if (!$event->isSuccessful()) {
       // If the submission didn't succeed, we are not sending any notification
-      // to the sender. Instead, he will get an error message on the screen.
+      // to the sender. Instead, they will get an error message on the screen.
```
brummbar:
```suggestion
      // to the sender. Instead, the user will get an error message on the screen.
```
Since we didn't say "user" in the previous part so the subject of this is not clear.

pfrenssen:
Thanks for the suggestions! I fixed all of them.

PR #1593, opened 2019-04-17
Isaicp 5175

junkuncz:


web/modules/custom/joinup_comment/joinup_comment.module
```diff
@@ -23,8 +24,17 @@ function joinup_comment_comment_view_alter(array &$build, EntityInterface $entit
...
  * Implements hook_form_FORM_ID_alter().
  *
- * Change the submit button label for the comment form.
+ * Disallow commenting for anonymous visitors and
+ * change the submit button label for the comment form.
  */
-function joinup_comment_form_comment_form_alter(&$form, FormStateInterface $form_state, $form_id) {
+function joinup_comment_form_comment_form_alter(array &$form, FormStateInterface $form_state, string $form_id) {
+  if (Drupal::currentUser()->isAnonymous()) {
+    $form['#access'] = FALSE;
```
idimopoulos: The access should not be determined on the form level but on the user permission level.
2 things need to be taken care of, the new installations and the update path.
For the new installations, the appropriate permissions should be revoked from the list of the permissions that the anonymous user is receiving.
For the updated path, a new post update is needed in the joinup_user.post_update.php file.

web/modules/custom/joinup_comment/joinup_comment.module
```diff
@@ -23,8 +24,17 @@ function joinup_comment_comment_view_alter(array &$build, EntityInterface $entit
...
+    $login_link = Link::createFromRoute('Login', 'user.login')->toString();
+    $register_link = Link::createFromRoute('create an account', 'user.register')->toString();
+    $text = sprintf('%s or %s to comment.', $login_link, $register_link);
```
idimopoulos: We can use the t() function both for the construction of the string and the variable replacement as it is also taking care of translations.

web/modules/custom/joinup_comment/joinup_comment.module
```diff
@@ -23,8 +24,17 @@ function joinup_comment_comment_view_alter(array &$build, EntityInterface $entit
...
+    $form['#markup'] = '<div class="center">' . $text . '</div>';
```
idimopoulos: The div wrappers should be part of the `#prefix` and `#suffix` elements of the form, not as inline markup.

web/modules/custom/joinup_comment/joinup_comment.module
idimopoulos: By the way, removing the permission from anonymous users will also probably remove this piece of code as the form will not even be shown.
A block is probably the best case scenario here.

tests/features/collection/collection.archived.feature
```diff
@@ -34,8 +34,8 @@ Feature: Creating content and commenting on archived collection
 
     When I am not logged in
     And I go to the "The Weeping's Stars" discussion
-    Then the following fields should be present "Your name, Email, Create comment"
-    And I should see the button "Post comment"
+    Then the following fields should not be present "Your name, Email, Create comment"
+    And I should not see the button "Post comment"
```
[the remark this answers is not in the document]
junkuncz: Added!

tests/features/comments/add_comment.feature
```diff
@@ -118,7 +72,7 @@ Feature: Add comments
       | discussion   | Is gossip bad?      | validated |
       | document     | Wikileaks           | validated |
 
-  Scenario Outline: Comments are disallowed for anonymous users in closed collections.
+  Scenario Outline: Comments are disallowed for anonymous users.
```
pfrenssen: It looks like it was the intention to add test coverage here to ensure that anonymous users can now not comment in both open and closed collections. The test was not changed though, so we now have no coverage any more for anonymous users commenting in open collections.

web/modules/custom/joinup_comment/joinup_comment.module
```diff
@@ -22,9 +22,7 @@ function joinup_comment_comment_view_alter(array &$build, EntityInterface $entit
 
 /**
  * Implements hook_form_FORM_ID_alter().
- *
- * Change the submit button label for the comment form.
```
pfrenssen: Why is this documentation removed? It is not great but still better than having to read the code.

web/themes/joinup/joinup_theme.theme
```diff
@@ -766,18 +766,10 @@ function joinup_theme_preprocess_field__comment(&$variables) {
...
   $user = \Drupal::currentUser();
-  if (
-    $user->isAnonymous() &&
-    $parent &&
-    $parent->getEntityTypeId() === 'rdf_entity' &&
-    $parent->bundle() === 'collection' &&
-    $parent->get('field_ar_closed')->value
-  ) {
+  if ($user->isAnonymous()) {
     $comments_status = 'closed_collection';
```
pfrenssen: This comment status is not relevant any more, let's change this to `authentication_required`.

PR #1604, opened 2019-04-19
Isaicp 5226: Add weight to the terms of the legal type vocabulary

idimopoulos:


web/modules/custom/joinup_licence/joinup_licence.post_update.php
```diff
@@ -42,3 +42,10 @@ function joinup_licence_post_update_import_legal_type_vocabulary() {
 function joinup_licence_post_update_eupl() {
   \Drupal::service('module_installer')->install(['eupl']);
 }
+
+/**
+ * Re import the legal type vocabulary so that the weight is imported.
+ */
+function joinup_licence_post_update_re_import_legal_type_vocabulary() {
+  joinup_licence_post_update_import_legal_type_vocabulary();
```
claudiu-cristea: If this will happen couple of times in the future, why not moving the logic into a service. Just calling a previous post-update looks odd.

idimopoulos: I have provided a service for this case. It seems a bit overkill for this case but this service can easily be extended for every fixture that we have (a simple array entry).

PR #1612, opened 2019-04-26
ISAICP-5230: Enssure the .html suffix in spdx licence urls.

idimopoulos:


web/themes/joinup/joinup_theme.theme
```diff
@@ -1548,7 +1548,8 @@ function joinup_theme_preprocess_rdf_entity__licence__search_list_item(&$variabl
 
     $variables['spdx_licence'] = $spdx_licence;
     $variables['spdx_title'] = $variables['content']['spdx_licence_label'];
-    $spdx_url = Url::fromUri($spdx_licence->id(), ['fragment' => 'licenceText']);
+    $url = preg_replace('/(.*)\.html$/', '\1', $spdx_licence->id()) . '.html';
```
pfrenssen: This is application logic which does not belong in the theme layer. We should be able to switch to a different theme and still have correct links rendered.

tests/features/communities/eupl/jla.feature
```diff
@@ -24,6 +24,9 @@ Feature:
     But I should not see the text "Random Licence"
     # Assert concatenated categories.
     And I should see the text "Governments/EU, Strong Community"
+    And I should see the following links:
+      | http://joinup.eu/licence/foo.html#licenseText |
+      | http://joinup.eu/licence/bar.html#licenseText |
```
pfrenssen: This test is failing.

web/themes/joinup/joinup_theme.theme
pfrenssen: We can do this by providing a template override in `joinup_licence.module`.

PR #1615, opened 2019-05-03
ISAICP-5214: Changed two pathauto patterns based on new requirements.

junkuncz:


web/modules/custom/joinup_core/joinup_core.post_update.php
```diff
@@ -525,6 +526,86 @@ function joinup_core_post_update_create_distribution_aliases(array &$sandbox) {
...
+    $sandbox['max'] = count($sandbox['entity_ids']);
+  }
+
+  $entity_storage = \Drupal::entityTypeManager()->getStorage('rdf_entity');
+  /** @var \Drupal\pathauto\PathautoGeneratorInterface $pathauto_generator */
+  $pathauto_generator = \Drupal::service('pathauto.generator');
+
+  $result = array_slice($sandbox['entity_ids'], $sandbox['current'], 50);
+  foreach ($entity_storage->loadMultiple($result) as $entity) {
+    $source_url = $entity->toUrl()->getInternalPath();
```
idimopoulos: As with the next function that also uses this line, this seems to be working in the tests but in reality, it causes an endless redirect. This is because `$entity->toUrl()->getInternalPath()` returns the core URL, `node/<node_id>`. That means that regardless of the target, the page will endlessly redirect to itself.
The tests pass because the path actually is the requested one but the page is blank due to the endless redirect.

You can use the `$entity->toUrl()->toString()` to get the "percieved" URL, which is the Alias, before and after the pathauto generator update. Please, look at `\spain_ctt_post_update_clean_ctt_duplicates` for an example.

web/modules/custom/joinup_core/joinup_core.post_update.php
```diff
@@ -525,6 +526,86 @@ function joinup_core_post_update_create_distribution_aliases(array &$sandbox) {
...
+    $sandbox['max'] = count($sandbox['entity_ids']);
+  }
+
+  $entity_storage = \Drupal::entityTypeManager()->getStorage('node');
+  /** @var \Drupal\pathauto\PathautoGeneratorInterface $pathauto_generator */
+  $pathauto_generator = \Drupal::service('pathauto.generator');
+
+  $result = array_slice($sandbox['entity_ids'], $sandbox['current'], 50);
+  foreach ($entity_storage->loadMultiple($result) as $entity) {
+    $source_url = $entity->toUrl()->getInternalPath();
```
idimopoulos: Same as the above comment.

web/modules/custom/joinup_core/joinup_core.post_update.php
junkuncz: Fixed!

web/modules/custom/joinup_core/joinup_core.post_update.php
junkuncz: Fixed!

PR #1616, opened 2019-05-07
Isaicp-5194: Support logos for news tiles

idimopoulos:


idimopoulos:
This ticket was accepted and approved yesterday during the sprint preparation and I was given the permission to merge it.

PR #1625, opened 2019-05-09
ISAICP-5191: Content items revision fix

idimopoulos:


web/modules/custom/joinup_core/joinup_core.install
```diff
@@ -62,6 +62,28 @@ function joinup_core_requirements($phase) {
         ];
       }
     }
+
+    // If there are forward published revisions in any entity, this will come up
+    // with this query. This is to ensure that the issue with the forward
+    // published revisions not being the latest published revision, is caught
+    // early if it happens again.
```
pfrenssen: This comment is assuming that everybody who will ever see this code will be familiar with "the issue with the forward published revisions not being the latest published revision".

It is allowed in core to publish older revisions, so can you explain in this comment exactly why Joinup cannot deal with this situation?

web/modules/custom/joinup_core/joinup_core.post_update.php
```diff
@@ -551,3 +551,19 @@ function joinup_core_post_update_enable_spdx() {
 function joinup_core_post_update_re_import_legal_type_vocabulary() {
   \Drupal::service('joinup_core.vocabulary_fixtures.helper')->importFixtures('licence-legal-type');
 }
+
+/**
+ * Corrects the versions of faulty news items.
+ */
+function joinup_core_post_update_set_news_default_version() {
+  $results = \Drupal::service('joinup_core.requirements_helper')->getNodesWithProblematicRevisions();
```
pfrenssen: Can you provide a comment here why this is needed? It is not clear why news items are considered to be "faulty". It is also not dealing with "versions" but with revisions.

This code looks quite dodgy without a proper explanation. This will just take a bunch of news items, update the published revision to the latest one. This is causing data loss and a security risk since if there are news items that have a more recent draft which is not ready for publication, then this code will blindly publish this. Apart from doing a database restore it will be impossible to recover from this action, the information about the currently published revision will be lost forever.

I am not sure what the problem is we are trying to fix but this doesn't seem like a good solution. It is perfectly allowed in core to publish older revisions and have a newer unpublished revision. Why is this needed?

web/modules/custom/joinup_core/src/RequirementsHelper.php
```diff
@@ -0,0 +1,67 @@
...
+   *
+   * @param \Drupal\Core\Database\Connection $connection
+   *   The connection class for the primary database storage.
+   */
+  public function __construct(Connection $connection) {
+    $this->connection = $connection;
+  }
+
+  /**
+   * Returns a query to find the forward revisions.
```
pfrenssen: This documentation is not correct. This does not return a query, it returns the result of a query. What is meant with "the forward revisions"?

This class is very strange and needs very clear documentation to explain why this code is required that can be understood by future generations which are not familiar with the problem.

pfrenssen:
Looking good now! The documentation now explains that this code is intended to perform a one time fix of a database discrepancy that occurred due to a cache synchronization issue. This makes it clear why we needed to do this.

PR #1631, opened 2019-05-16
Isaicp 5254: Allow the users to unsubscribe from all collections.

idimopoulos:


web/modules/custom/joinup_subscription/src/Form/SubscriptionDashboardForm.php
```diff
@@ -85,6 +86,15 @@ public function buildForm(array $form, FormStateInterface $form_state, AccountIn
     $memberships = $this->relationManager->getUserGroupMembershipsByBundle($user, 'rdf_entity', 'collection');
     $bundle_info = $this->entityTypeBundleInfo->getBundleInfo('node');
 
+    $form['usubscribe_all'] = [
+      '#type' => 'link',
```
claudiu-cristea: Here's a **UX point of view** about this approach: Allowing the user to _Unsubscribe from all_ when he has no subscription looks like a bad UX. I see a better way:

* Don't show this button (or show it as a disabled button - hm, it's a link!) if the use has no subscription. You can enclose this element into an `if (...)` or set the `#access` property.
* Add the same logic in `UnsubscribeFromAllCollectionsForm::access()`, so a user cannot hit the "unsubscribe all" confirmation form even by accident.
* You can extend the `joinup_subscription.subscription` service (and interface) with a new method `userHasSubscriptions()` and that can be used in both places. That would be the actual `UnsubscribeFromAllCollectionsForm::getUserMembershipIds()`

web/modules/custom/joinup_user/src/Form/UnsubscribeFromAllCollectionsForm.php
```diff
@@ -0,0 +1,281 @@
...
+        '#type' => 'item',
+        '#markup' => t('You are currently registered to be notified for the following collections:'),
+        '#tree' => TRUE,
+        'items' => [
+          '#theme' => 'item_list',
+          '#items' => $labels,
+        ],
+      ];
+    }
+    else {
```
claudiu-cristea: If we go with the UX change, this should throw an exception because such a case should not exist and reveals a development problem.

web/modules/custom/joinup_user/src/Form/UnsubscribeFromAllCollectionsForm.php
```diff
@@ -0,0 +1,281 @@
...
+  public function getCancelUrl() {
+    return Url::fromRoute('joinup_subscription.subscriptions_page');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildForm(array $form, FormStateInterface $form_state) {
+    if ($memberships = $this->getUserMembershipIds()) {
+      $memberships = $this->entityTypeManager->getStorage('og_membership')->loadMultiple($memberships);
```
claudiu-cristea: Let's not use the same name `$memberships` for both, list of IDs and list of entities. The first could be `$membership_ids`?

web/modules/custom/joinup_user/src/Form/UnsubscribeFromAllCollectionsForm.php
```diff
@@ -0,0 +1,281 @@
+<?php
```
claudiu-cristea: Lacks `declare(strict_types = 1)`.

web/modules/custom/joinup_user/src/Form/UnsubscribeFromAllCollectionsForm.php
```diff
@@ -0,0 +1,281 @@
...
+    return Url::fromRoute('joinup_subscription.subscriptions_page');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildForm(array $form, FormStateInterface $form_state) {
+    if ($memberships = $this->getUserMembershipIds()) {
+      $memberships = $this->entityTypeManager->getStorage('og_membership')->loadMultiple($memberships);
+      $labels = array_map(function (OgMembershipInterface $membership) {
```
claudiu-cristea: The closure return should be strict typed.

web/modules/custom/joinup_user/src/Form/UnsubscribeFromAllCollectionsForm.php
```diff
@@ -0,0 +1,281 @@
...
+        return $membership->getGroup()->label();
+      }, $memberships);
+      asort($labels);
+      $form = parent::buildForm($form, $form_state);
+      $form['information'] = [
+        '#type' => 'item',
+        '#markup' => t('You are currently registered to be notified for the following collections:'),
```
claudiu-cristea: s/t()/$this->()

web/modules/custom/joinup_user/src/Form/UnsubscribeFromAllCollectionsForm.php
```diff
@@ -0,0 +1,281 @@
...
+        '#tree' => TRUE,
```
claudiu-cristea: Why?

web/modules/custom/joinup_user/src/Form/UnsubscribeFromAllCollectionsForm.php
```diff
@@ -0,0 +1,281 @@
...
+    $operations = [];
+    foreach ($membership_ids as $membership_id) {
+      $operations[] = [
+        '\Drupal\joinup_user\Form\UnsubscribeFromAllCollectionsForm::membershipUnsubscribe',
+        [$membership_id],
+      ];
+    }
+
+    $batch = [
+      'title' => t('Unsubscribe from collections'),
```
claudiu-cristea: s/t()/$this->()

web/modules/custom/joinup_user/src/Form/UnsubscribeFromAllCollectionsForm.php
```diff
@@ -0,0 +1,281 @@
...
+      'operations' => $operations,
+      'finished' => [$this, 'membershipUnsubscribeFinish'],
+      'init_message' => t('Initiating...'),
```
claudiu-cristea: s/t()/$this->()

web/modules/custom/joinup_user/src/Form/UnsubscribeFromAllCollectionsForm.php
```diff
@@ -0,0 +1,281 @@
...
+      'progress_message' => t('Processed @current out of @total. Estimated time: @estimate.'),
```
claudiu-cristea: s/t()/$this->()

web/modules/custom/joinup_user/src/Form/UnsubscribeFromAllCollectionsForm.php
```diff
@@ -0,0 +1,281 @@
...
+   */
+  public function submitForm(array &$form, FormStateInterface $form_state) {
+    $membership_ids = $this->getUserMembershipIds();
+    $redirect_url = $this->getCancelUrl();
+    $form_state->setRedirect($redirect_url->getRouteName(), $redirect_url->getRouteParameters());
+
+    $operations = [];
+    foreach ($membership_ids as $membership_id) {
+      $operations[] = [
+        '\Drupal\joinup_user\Form\UnsubscribeFromAllCollectionsForm::membershipUnsubscribe',
```
claudiu-cristea: As of PHP 5.5.0 we have a constant that allows us to get the fully qualified name of the class w/o needing to hardocode it: `UnsubscribeFromAllCollectionsForm::class . '::membershipUnsubscribe'`

web/modules/custom/joinup_user/src/Form/UnsubscribeFromAllCollectionsForm.php
```diff
@@ -0,0 +1,281 @@
...
+   * @param array $results
+   *   The value set in $context['results'] by callback_batch_operation().
+   * @param array $operations
+   *   If $success is FALSE, contains the operations that remained unprocessed.
+   *
+   * @return \Symfony\Component\HttpFoundation\RedirectResponse
+   *   The redirect response.
+   */
+  public function membershipUnsubscribeFinish(bool $success, array $results, array $operations): RedirectResponse {
+    // @see \callback_batch_finished.
```
claudiu-cristea: The `@see` should not end with a dot. Also the function name should end with `()`.

web/modules/custom/joinup_user/src/Form/UnsubscribeFromAllCollectionsForm.php
```diff
@@ -0,0 +1,281 @@
...
+  /**
+   * Access check for the UnsubscribeFromAllCollectionsForm.
+   *
+   * @param \Drupal\Core\Session\AccountInterface $account_proxy
+   *   The user from the route.
+   *
+   * @return \Drupal\Core\Access\AccessResult
+   *   The access result object.
+   */
+  public function access(AccountInterface $account_proxy) {
```
claudiu-cristea: Missed method return strict typing.

web/modules/custom/joinup_user/src/Form/UnsubscribeFromAllCollectionsForm.php
```diff
@@ -0,0 +1,281 @@
...
+      return AccessResult::forbidden();
+    }
+
+    return AccessResult::allowedIf($this->user->id() === $account_proxy->id());
+  }
+
+  /**
+   * Returns an array of membership ids that the user has active subscriptions.
+   *
+   * @return \Drupal\og\OgMembershipInterface[]
```
claudiu-cristea: This is wrong. As I see, it should be `int[]`.

tests/features/bootstrap/FeatureContext.php
```diff
@@ -405,22 +406,67 @@ public function assertOptionNotSelected($text) {
...
    *
    * @Then the option with text :option from select :select is selected
    */
   public function assertFieldOptionSelected($option, $select) {
-    $element = $this->findSelect($select);
+    $this->assertFieldOptionSelectedInRegion($option, $select);
+  }
+
+  /**
+   * Find the selected option of the select in a card and check the text.
```
claudiu-cristea: s/Find/Finds

tests/features/bootstrap/FeatureContext.php
```diff
@@ -405,22 +406,67 @@ public function assertOptionNotSelected($text) {
...
+   * @param string $heading
+   *   The card heading.
+   *
+   * @throws \Exception
+   *   Thrown when the select is not found in the page or the selected option is
+   *   not the expected one.
+   *
+   * @Then the option with text :option from select :select is selected in the :heading card
+   */
+  public function assertFieldOptionSelectedInCard($option, $select, $heading) {
```
claudiu-cristea: Missed method return strict typing.

tests/features/bootstrap/FeatureContext.php
```diff
@@ -405,22 +406,67 @@ public function assertOptionNotSelected($text) {
...
+   * @param string $select
+   *   CSS selector of the select field.
+   * @param \Behat\Mink\Element\TraversableElement $region
+   *   (optional) The region to search in. Defaults to the whole page.
+   *
+   * @throws \Exception
+   *   Thrown when the select is not found in the page or the selected option is
+   *   not the expected one.
+   */
+  protected function assertFieldOptionSelectedInRegion($option, $select, TraversableElement $region = NULL) {
```
claudiu-cristea: Missed method return strict typing and strict typing for `$option` and `$select` arguments.

Starting with the word "assert" would confuse some because we usually use such namings for step definitions, but this is more a helper (code reusing). Probably `isFieldOptionSelectedInRegion`? However, it's still correct... I have no strong feeling.

tests/src/Traits/TraversingTrait.php
```diff
@@ -15,16 +16,22 @@ trait TraversingTrait {
...
    *
-   * @return \Behat\Mink\Element\NodeElement
+   * @return \Behat\Mink\Element\TraversableElement
    *   The select element.
    *
    * @throws \Exception
    *   Thrown when no select field is found.
    */
-  protected function findSelect($select) {
+  protected function findSelect($select, TraversableElement $region = NULL) {
```
claudiu-cristea: Missing strict typing for `$select` and return.

tests/src/Traits/TraversingTrait.php
```diff
@@ -157,28 +164,47 @@ protected function getTiles($region = NULL) {
    *
    * @param string $heading
    *   The heading of the tile to find.
+   */
+  protected function getTileByHeading($heading) {
```
claudiu-cristea: Missing strict typing for argument and return.

tests/src/Traits/TraversingTrait.php
```diff
@@ -157,28 +164,47 @@ protected function getTiles($region = NULL) {
...
+
+  /**
+   * Finds a card element by its heading.
    *
-   * @return \Behat\Mink\Element\NodeElement|null
-   *   The tile element, or null if not found.
+   * @param string $heading
+   *   The heading of the tile to find.
+   */
+  protected function getCollectionSubscriptionCardByHeading($heading) {
```
claudiu-cristea: Missing strict typing for argument and return.

tests/src/Traits/TraversingTrait.php
```diff
@@ -157,28 +164,47 @@ protected function getTiles($region = NULL) {
    *
    * @param string $heading
    *   The heading of the tile to find.
+   */
```
claudiu-cristea: Missing the `@return` tag.

tests/src/Traits/TraversingTrait.php
```diff
@@ -157,28 +164,47 @@ protected function getTiles($region = NULL) {
...
+  }
+
+  /**
+   * Finds a card element by its heading.
    *
-   * @return \Behat\Mink\Element\NodeElement|null
-   *   The tile element, or null if not found.
+   * @param string $heading
+   *   The heading of the tile to find.
+   */
```
claudiu-cristea: Missing the `@return` tag.

tests/src/Traits/TraversingTrait.php
```diff
@@ -157,28 +164,47 @@ protected function getTiles($region = NULL) {
...
+  }
+
+  /**
+   * Finds a list item element by its heading.
+   *
+   * @param string $type
+   *   The class of the element that is searched for.
+   * @param string $heading
+   *   The heading on the item.
    *
```
claudiu-cristea: Missing the `@return` tag.

tests/src/Traits/TraversingTrait.php
```diff
@@ -157,28 +164,47 @@ protected function getTiles($region = NULL) {
...
    * @throws \Behat\Mink\Exception\ElementNotFoundException
    *   Thrown when the tile is not found.
    */
-  protected function getTileByHeading($heading) {
-    // Locate all the tiles.
-    $xpath = '//*[@class and contains(concat(" ", normalize-space(@class), " "), " listing__item--tile ")]';
+  protected function getListingByHeading($type, $heading) {
```
claudiu-cristea: Missing strict typing for arguments & return.

web/modules/custom/joinup_user/src/Form/UnsubscribeFromAllCollectionsForm.php
idimopoulos: From your suggestions above, I will add the access check in the form itself as well. That means that even by entering the url, there will be no access there. Which means that this is not needed anymore.

web/modules/custom/joinup_user/src/Form/UnsubscribeFromAllCollectionsForm.php
idimopoulos: Yeah that was probably a leftover from a previous iteration.

web/modules/custom/joinup_user/src/Form/UnsubscribeFromAllCollectionsForm.php
idimopoulos: Removed actually. Wrong assumption.

web/modules/custom/joinup_user/src/Form/UnsubscribeFromAllCollectionsForm.php
idimopoulos: I think we need something like that as an automatic check for me :D

pfrenssen: There is in fact an automatic check for this included in the `DrupalPractice` standard for PHP_CodeSniffer: https://git.drupalcode.org/project/coder/blob/8.x-3.x/coder_sniffer/DrupalPractice/Sniffs/Objects/GlobalFunctionSniff.php#L41

I will make a ticket to run this.

pfrenssen: @idimopoulos see #1633

tests/features/bootstrap/FeatureContext.php
pfrenssen: I think in this case `assertFieldOptionSelectedInRegion()` is still the better choice. It is true that we usually use this for step definitions, but it is in fact consistent with other methods that throw an exception if the expected condition is not met (for example the asserts in PHPUnit or Drupal core test traits).

The `is...` naming usually indicates that a boolean is being returned which will indicate if the condition is met. This is not the case here.

Anyway this is a small nitpick :)

web/modules/custom/joinup_subscription/src/Form/UnsubscribeFromAllCollectionsForm.php
```diff
@@ -0,0 +1,271 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function create(ContainerInterface $container) {
+    return new static(
+      $container->get('entity_type.manager'),
+      $container->get('og.membership_manager'),
+      $container->get('renderer'),
+      $container->get('current_route_match')
```
pfrenssen: We actually don't need to retrieve the route matcher from the container, we can call `::getRouteMatch()` which is supplied by the base class.

web/modules/custom/joinup_subscription/src/Form/UnsubscribeFromAllCollectionsForm.php
```diff
@@ -0,0 +1,271 @@
...
+          '#theme' => 'item_list',
+          '#items' => $labels,
+        ],
+      ];
+    }
+    return $form;
+  }
+
+  /**
+   * Submit callback for the return button.
```
pfrenssen: Return button? I suppose you mean submit button.

web/modules/custom/joinup_subscription/src/Form/UnsubscribeFromAllCollectionsForm.php
```diff
@@ -0,0 +1,271 @@
...
+   *
+   * Returns the user to the same destination as the cancel URL.
+   *
+   * @param array $form
+   *   The form array.
+   * @param \Drupal\Core\Form\FormStateInterface $form_state
+   *   The form state object.
+   */
+  public function returnSubmit(array &$form, FormStateInterface $form_state): void {
```
pfrenssen: Is this method even used? I cannot find any code that calls it.

web/modules/custom/joinup_subscription/src/Form/UnsubscribeFromAllCollectionsForm.php
```diff
@@ -0,0 +1,271 @@
...
+   * @return int[]
+   *   An array of memberships.
+   */
+  protected function getUserMembershipIds(): array {
+    $query = $this->entityTypeManager
+      ->getStorage('og_membership')
+      ->getQuery()
+      ->condition('uid', $this->user->id())
+      ->condition('entity_bundle', 'collection')
+      ->exists('subscription_bundles');
```
pfrenssen: Ooooh a rare `->exists()` appears!

PR #1633, opened 2019-05-22
Replace calls to t() with $this->t()

pfrenssen:


pfrenssen:
How come these were not detected by the sniffer?

pfrenssen:
It seems that all the cases which were not detected have an argument list split over multiple lines.

pfrenssen:
It was not because of the split lines, but because these are in classes that are not objects, but are implementing `ContainerInjectionInterface`.

Ref https://github.com/pfrenssen/coder/pull/46

PR #1635, opened 2019-05-23
ISAICP-5238: Update dependencies

pfrenssen:


composer.json
```diff
@@ -47,15 +47,16 @@
         "drupal/persistent_login": "~1.0",
         "drupal/phingdrushtask": "dev-7.x-2.x#f4b020f09da3c174fce40f2bd9a9617689f57124",
         "drupal/r4032login": "dev-1.x#391f10d69b9b8c466a2fddac2a6b99f1289e7781",
-        "drupal/rdf_entity": "^1.0",
+        "drupal/rdf_entity": "^1.0.0-alpha17",
```
idimopoulos: Actually, this, along with the spdx, was appointed by me guided by Claudiu. Why is it bad to have the `^1.0` annotation? And why would it install an older version? We are controlling both of these libraries so every time we make a new release, we should get it by default.

composer.json
```diff
@@ -47,15 +47,16 @@
...
         "drupal/redirect": "^1.0",
         "drupal/redis": "^1.1",
         "drupal/search_api": "^1.11",
         "drupal/search_api_solr": "2.2",
         "drupal/simplenews": "^1.0-beta1",
         "drupal/smart_trim": "^1.0",
         "drupal/social_media_links": "^2.5",
-        "drupal/spdx": "^1.0",
+        "drupal/sparql_entity_storage": "dev-8.x-1.x",
```
idimopoulos: This is a dependency of the `drupal/rdf_entity`. We do not need this line.

composer.json
pfrenssen: The `composer.json` lists the dependency requirements of our project. If we put `^1.0` then we are declaring that we are compatible with all possible 1.x versions of RDF Entity. Since our `minimum-stability` is set to `dev` Composer is allowed to use any version of RDF Entity starting with the very first commit of `8.x-1.x` and all tags including alpha1 and above.

But we are not in fact compatible with anything below `^1.0.0-alpha17` so we should not allow Composer to use anything below this.

composer.json
pfrenssen: I added this because Composer was not resolving this correctly before the RDF Split ticket was merged. You're right, it can be deleted now.

PR #1639, opened 2019-05-24
ISAICP-5174: Added user's origin to its profile.

junkuncz:


web/themes/joinup/templates/user/user--group-header.html.twig
```diff
@@ -50,6 +50,12 @@
             <span class="stats__text">{{ 'Member for'|t }}:</span>
             <span class="stats__text bold">{{ member_for }}</span>
           </div>
+          {% if content.field_user_nationality|render %}
```
idimopoulos: Do we need the render function here? I tried it and it works without the render function. If you remove it from the view mode, it does not show, it does not log any errors. But here we are rendering it once more.
Let's remove it (I know it is like this in the rest) if there is no reason for it.

tests/features/user/profile.feature
```diff
@@ -58,8 +58,12 @@ Feature: User profile
     # A user should not be able to edit the profile page of another user.
     When I go to the public profile of "Domenico Ghirlandaio"
     Then I should not see the link "Edit"
+    # Verify that the user's "Country of origin" field is visible on its profile.
+    When I go to the public profile of "Leonardo Da Vinci"
+    Then I should see the text "Country of origin:" in the "Header" region
+    And I should see the link "Italy" in the "Header" region
```
idimopoulos: In the screenshot attached in the roiginal ticket, this is not a link. Plus, we do not have anything styled for the taxonomy pages. Please, change this into the "featured" view mode of the user to not link to the original entity.

web/themes/joinup/templates/user/user--group-header.html.twig
junkuncz: Really!
Changed!

tests/features/user/profile.feature
junkuncz: Alright!
I also adjusted the tests.

PR #1640, opened 2019-05-24
ISAICP-5281: Fixed maps on events.

junkuncz:


pfrenssen:
It seems like the use case that was originally reported is not covered by a test.

Let's add a test that adds a physical location, checks that the map is displayed, then edits the event, removes the phyical location, and then checks that the map is no longer displayed.

In order to do this we will need to propose a PR for the https://github.com/openeuropa/oe_webtools_location project to add the step definition for `Then I should not see a map on the page` in `OpenEuropaWebtoolsMapsSubContext`.

tests/features/joinup_event/event.edit.feature
```diff
@@ -4,6 +4,24 @@ Feature: "Event page" editing.
   As an owner of the event
   I need to be able to edit it.
 
+  Scenario: Add and remove map
+    Given collections:
+      | title            | logo     | banner     | state     |
+      | Stream of Dreams | logo.png | banner.jpg | validated |
```
pfrenssen: This is using the same name of a collection that is already used in another test. Please use a unique name so we can easily recognize failing tests.

tests/features/joinup_event/event.edit.feature
```diff
@@ -4,6 +4,24 @@ Feature: "Event page" editing.
...
+    And I am logged in as a facilitator of the "Stream of Dreams" collection
+    When I go to the homepage of the "Stream of Dreams" collection
+    And I click "Add event" in the plus button menu
+    When I fill in the following:
+      | Title             | An amazing event                      |
```
pfrenssen: There is already an event used with this name, description and location in the `add_event.collection` test. Please choose a unique name, description and location.

tests/features/joinup_event/event.edit.feature
junkuncz: Changes added.

tests/features/joinup_event/event.edit.feature
junkuncz: Changes added.

PR #1650, opened 2019-05-29
ISAICP-5278: Fix demo installation on conflicts

idimopoulos:


scripts/rpmbuild/make_joinup_build_uat.sh
```diff
@@ -32,7 +32,7 @@ JOINUP_DIR=${SOURCES_DIR}/Joinup-${BUILD_VERSION}
 mkdir -p ${JOINUP_DIR} || exit 1
 
 # Build the site.
-${COMPOSER_PATH} install || exit 1
+COMPOSER_DISCARD_CHANGES=true ${COMPOSER_PATH} install --no-interaction --prefer-source --prefer-dist || exit 1
```
pfrenssen: It doesn't make sense to specify both `--prefer-source` and `--prefer-dist` since they are the opposites of eachother.

See https://github.com/composer/composer/blob/master/doc/03-cli.md#options-1

> **--prefer-source**: There are two ways of downloading a package: source and dist. For stable versions Composer will use the dist by default. [...]
> **--prefer-dist**: Reverse of --prefer-source, Composer will install from dist if possible. [...]

idimopoulos: Thanks, indeed. It is now fixed. `--prefer-dist` is now the only one.

PR #1665, opened 2019-06-05
5304: EU Login: Moderators should be able to cancel users with deletion

claudiu-cristea:


idimopoulos:
Merging as the failure was a random one.

PR #1681, opened 2019-06-24
Isaicp-5338: Allow users with edit access to unpublish custom pages

idimopoulos:


web/modules/custom/custom_page/custom_page.module
```diff
@@ -187,3 +188,21 @@ function custom_page_node_access(NodeInterface $node, $op, AccountInterface $acc
   }
   return AccessResult::neutral();
 }
+
+/**
+ * Implements hook_form_alter().
+ */
+function custom_page_form_alter(&$form, FormStateInterface $form_state, $form_id) {
```
claudiu-cristea: Instead of hacking into the form, I would prefer to use the proper entity field API: `hook_entity_field_access()`. Also we don't need to check if the user is able to edit because if he/she cannot edit is also not allowed to edit a field. I've been discussed this aspect with @idimopoulos and we agreed to use `hook_entity_field_access()`.

PR #1683, opened 2019-06-25
5310: Entity type provided by a module installed via update hook not available 

claudiu-cristea:


composer.json
```diff
@@ -66,7 +66,7 @@
         "drupal/view_unpublished": "~1.0@alpha",
         "drupal/views_bulk_operations": "^2.0",
         "drupal/views_data_export": "^1.0-beta1",
-        "drush/drush": "~9.7",
+        "drush/drush": "dev-master",
```
pfrenssen: The patch applies cleanly on 9.7.0, let's not switch to master.

claudiu-cristea: It doesn't, because of https://github.com/drush-ops/drush/issues/4104

pfrenssen: Oh I see! Let's keep it on master for now.

PR #1685, opened 2019-06-28
Isaicp 5295

pfrenssen:


pfrenssen:
In addition to the integration with the CAS mock server I had to provide a Drush configuration file that configures the base URL of the website using the global `--uri` parameter. This is required since the CAS mock server module needs to override the URL of the CAS server with the base URL of the Drupal site. This information is normally taken from the webserver through the PHP `$_SERVER` magic variable, but this is not populated when bootstrapping Drupal though the CLI (i.e. when using Drush).

Support for the Drush configuration files was already added upstream in the `drupal-composer/drupal-project` project so instead of reinventing the wheel I updated Joinup to the latest version of drupal-project. We now also benefit from the latest improvements made by them.

claudiu-cristea:
In the meantime I've opened https://github.com/pfrenssen/cas_mock_server/pull/4 but I need Git push permissions also on https://www.drupal.org/project/cas_mock_server because Drupal CI knows only to handle Drupal repo and `cas_mock_server` is not yet added to Packagist.

PR #1695, opened 2019-07-05
ISAICP-5370: Coding standards adherence for info files

pfrenssen:


pfrenssen:
I used this script to generate the module dependency namespaces: https://gist.github.com/pfrenssen/f2a0c94c5a30da89f95880c22244edc3

phpcs-ruleset.xml.dist
```diff
@@ -26,7 +26,6 @@
         <exclude name="DrupalPractice.Constants" />
         <exclude name="DrupalPractice.FunctionDefinitions" />
         <exclude name="DrupalPractice.General" />
-        <exclude name="DrupalPractice.InfoFiles" />
```
claudiu-cristea: Why are we removing this rule?

pfrenssen: It is being removed from the _exclude list_, meaning that it is in fact being added!

PR #1697, opened 2019-07-07
ISAICP-5347: Implement the Cookie Concent Kit

idimopoulos:


tests/features/joinup_cck/joinup_cck.feature
```diff
@@ -0,0 +1,33 @@
...
+    And I click "Sign in"
+    And I fill in "Email or username" with "test_cck"
+    And I fill in "Password" with "test_cck"
+    And I press "Sign in"
+    Then I should not see the "Cookie consent banner" region
+
+    Examples:
+      | link   |
+      | accept |
+      | refuse |
```
claudiu-cristea: Missing empty new-line at EOF.

web/modules/custom/joinup_cck/joinup_cck.info.yml
```diff
@@ -0,0 +1,5 @@
+name: 'Joinup CCK'
```
claudiu-cristea: Having a module only to hang a script on each page looks like an arrogance to me :) What if we rename to `joinup_gdpr` so that we can drop here also other related functionality in the future (but I'm not so sure that CCK is part of GDPR, is it?).

But, as I remember, we've been discussed the alternative to provide a PR against this module https://github.com/openeuropa/oe_webtools. Wouldn't that be better?

web/modules/custom/joinup_user/joinup_user.install
```diff
@@ -51,6 +51,7 @@ function joinup_user_install() {
     'access content',
     'access tour',
     'access user profiles',
+    'never autoplay videos',
```
claudiu-cristea: Still don't understand why we're doing this crap. Why the permissions are not set via config management. That would make obsolete also the post-update.

web/modules/custom/joinup_video/joinup_video.module
```diff
@@ -25,3 +29,35 @@ function joinup_video_form_video_embed_dialog_alter(&$form, FormStateInterface $
...
+    // The query parameters are being built and constructed using the
+    // http_build_query rather than allowing URL to parse them. This is because
+    // the \Drupal\Component\Utility\UrlHelper::buildQuery constructs the query
+    // parameters differently than how PHP does and the result is that boolean
+    // values are not parsed properly.
+    // @todo: Remove the workaround once this is fixed in core.
+    // @see \Drupal\Component\Utility\UrlHelper::buildQuery
+    // @see https://www.drupal.org/project/drupal/issues/2248257
+    $query_parameters = http_build_query($variables['query']);
+    $url = Url::fromUri(rtrim($variables['url']) . '?' . $query_parameters, [
```
claudiu-cristea: Probably a better idea is to prepare first `$variables['query']` and replace booleans with strings. I.e. `TRUE` -> `'1'`, `FALSE` -> `'0'`. Then we can use `Url::fromUri()` in a standard way.

web/modules/custom/joinup_user/joinup_user.install
idimopoulos: Hmm, I tested this again now. When we decided this approach it was that no matter how we tried it, we would get the "the config object already exists in active configuration" error. No matter if we were to place the files in the `joinup_user` module, or the `joinup` profile.
I see that now, we still cannot place the files in the `joinup_user` module but we can place them in the `joinup` profile. I will change this.

web/modules/custom/joinup_cck/joinup_cck.info.yml
idimopoulos: I renamed it to joinup_privacy as gdpr can also be quite specific. Now we can have all privacy settings in this module. 
About the second part, I do not recall something about that. Is the webtools planning to have this CCK as a default functionality? We could go with creating the PR towards them but we would need collaboration.

web/modules/custom/joinup_user/joinup_user.post_update.php
```diff
@@ -129,3 +129,10 @@ function joinup_user_post_update_reset_default_icons() {
   include_once __DIR__ . '/joinup_user.install';
   joinup_user_setup_default_avatar();
 }
+
+/**
+ * Grant the anonymous users the 'never autoplay videos' permission.
+ */
+function joinup_user_post_update_add_never_autoplay_anonymous_permission(): void {
```
claudiu-cristea: Why? Isn't this handled by config management now?

web/modules/custom/joinup_user/joinup_user.install
```diff
@@ -6,66 +6,12 @@
...
-  ];
-  user_role_grant_permissions(AccountInterface::ANONYMOUS_ROLE, $anonymous_permissions);
-
-  joinup_user_setup_default_avatar();
-}
-
-/**
- * Set a default image for the users avatar.
- */
-function joinup_user_setup_default_avatar() {
```
claudiu-cristea: Can we keep this function as is used in `joinup_user_post_update_reset_default_icons()`? I know that the post update already ran on PROD but is not necessary true for a 3rd party using our platform.

PR #1702, opened 2019-07-09
5348: Legal notice check on registration/change

idimopoulos:


web/modules/custom/legal/joinup_legal.behat.inc
```diff
@@ -0,0 +1,171 @@
...
+   *
+   * @param \Behat\Gherkin\Node\TableNode $collection_table
+   *   The collection data.
+   *
+   * @throws \Exception
+   *   Thrown when a column name is incorrect.
+   *
+   * @Given (the following )legal document version(s):
+   */
+  public function givenLegalDocumentVersions(TableNode $collection_table) {
```
idimopoulos: `: void` return is missing.

PR #1711, opened 2019-07-11
ISAICP-5377: Provide an update path to delete orphaned files.

idimopoulos:


web/modules/custom/joinup_core/joinup_core.post_update.php
```diff
@@ -674,3 +674,20 @@ function joinup_core_post_update_enable_nio() {
...
+ * Deletes unused files.
+ */
+function joinup_core_post_update_delete_orphaned_files() {
+  $query = \Drupal::database()->select('file_managed', 'fm')
+    ->fields('fm', ['fid']);
+  $query->leftJoin('file_usage', 'fu', 'fm.fid = fu.fid');
+  $query->leftJoin('node_revision', 'nr', 'fu.id = nr.nid');
+  $query->condition('fu.type', 'node');
+  $query->isNull('nr.nid');
+  $results = $query->execute()->fetchCol();
```
pfrenssen: This query should have proper documentation explaining why this is needed, why there are revisions in the database that have `NULL` for the node ID, and why this is not considering the fact that there might be other existing references to the same files that do not have `NULL` as the node ID.

idimopoulos: I actually think you are right but on the second point. There is nothing wrong with the query, the decision is the problem. I think it was a bad approach that based on two files that was explicitly asked to be deleted, we went with the approach to delete everything unused. 
Very helpful catch. Updated the code to simply remove the two files requested explicitly.

PR #1731, opened 2019-08-01
ISAICP-5451: Split the organisation and the eulogin organisation fields.

idimopoulos:


pfrenssen:
I'm sorry but none of the changes proposed here are making any improvements for us. We don't need these fields split, we need to keep the original approach unchanged.

If we would merge this we are not getting any closer to releasing the epic, on the contrary this introduces more things we will need to remove.

pfrenssen:
New PR that reverts to the original approach is pushed in #1735

PR #1734, opened 2019-08-02
ISAICP-5455: Temp fix for the news publication issue.

idimopoulos:


web/modules/custom/joinup_core/joinup_core.post_update.php
```diff
@@ -681,3 +681,104 @@ function joinup_core_post_update_install_publication_date() {
...
+    $sandbox['nids'] = $database->query('SELECT nid FROM {node} n')->fetchCol();
+    $sandbox['progress'] = 0;
+    $sandbox['total'] = count($sandbox['nids']);
+  }
+
+  $nids = array_splice($sandbox['nids'], 0, 300);
+  $query = $database->select('node', 'n');
+  $query->leftJoin('node_field_revision', 'nr', 'n.nid = nr.nid');
+  $query->fields('nr', ['nid', 'vid', 'changed']);
+  $query->condition('status', 1);
```
pfrenssen: This field is on the `node_field_revision` table, let's add the prefix `nr.` to make this explicit.

web/modules/custom/joinup_core/joinup_core.post_update.php
```diff
@@ -681,3 +681,104 @@ function joinup_core_post_update_install_publication_date() {
...
+        ->update($node_storage->getDataTable())
+        ->expression('published_at', PUBLICATION_DATE_DEFAULT)
+        ->condition('nid', $nid)
+        ->execute();
+      $database
+        ->update($node_storage->getRevisionDataTable())
+        ->expression('published_at', PUBLICATION_DATE_DEFAULT)
+        ->condition('nid', $nid)
+        ->execute();
+    }
```
pfrenssen: We can improve performance by keeping track of the node IDs of the nodes that were never published, and then at the end of the batch perform these update queries to all of them at once. This is more efficient than doing two separate queries for each single node.

web/modules/custom/joinup_core/joinup_core.post_update.php
```diff
@@ -681,3 +681,104 @@ function joinup_core_post_update_install_publication_date() {
...
+        ->execute();
+
+      // The revisions that are left have a null publication date and need to
+      // receive the default value.
+      $database
+        ->update($node_storage->getRevisionDataTable())
+        ->expression('published_at', PUBLICATION_DATE_DEFAULT)
+        ->condition('nid', $nid)
+        ->isNull('published_at')
+        ->execute();
```
pfrenssen: This last step can also be done in a single query, at the end of the entire batch operation. If we have 20000 published nodes, this now does 20000 queries. But if we wait until all the published revisions are processed, and then run this query at the end (without the `->condition('nid', $nid)` call) then we only need to do 1 query.

idimopoulos:
I cannot approve this PR because I opened it and it does not allow me to. But I approve these changes and this PR :D

PR #1738, opened 2019-08-06
5459: Run CPHP PHPUnit tests with MySQL rather than SQLite

claudiu-cristea:


pfrenssen:
These changes are intended to be used in combination with the new `refs/heads/ISAICP-5459` pipeline which can be inspected on `https://continuousphp.com/git-hub/ec-europa/joinup-dev/settings`.

I see that in that pipeline there is a phing target being executed called "import-legacy-test-db". I am assuming this was intended for the tests in `joinup_migrate` which have been disabled. Can we delete this Phing target in scope of this ticket?

pfrenssen:
Discussed with @claudiu-cristea in private chat and we decided to keep the removal of the legacy DB in a separate ticket, so that we can remove all migration related code at the same time. This means this is good to go, great results!

PR #1753, opened 2019-08-13
ISAICP-5443: Order of actions.

idimopoulos:


composer.lock
```diff
@@ -3044,7 +3044,8 @@
                     "Node revision routes don't convert {node} param @see https://www.drupal.org/project/drupal/issues/2730631": "https://www.drupal.org/files/issues/2019-04-04/2730631_85_0.patch",
                     "Disable tableselect element options @see https://www.drupal.org/project/drupal/issues/2895352": "https://www.drupal.org/files/issues/2019-01-18/2895352-8.patch",
                     "Make it easier to trigger a tour @see https://www.drupal.org/project/drupal/issues/3012027": "https://www.drupal.org/files/issues/2018-11-07/3012027-2.patch",
-                    "Cache tags are not invalidated on revision delete @see https://www.drupal.org/project/drupal/issues/2945928": "https://www.drupal.org/files/issues/entity-cache_revision_deletion-2945928-2.patch"
+                    "Cache tags are not invalidated on revision delete @see https://www.drupal.org/project/drupal/issues/2945928": "https://www.drupal.org/files/issues/entity-cache_revision_deletion-2945928-2.patch",
+                    "Allow sorting of actions @see https://www.drupal.org/project/drupal/issues/2381293": "resources/patch/2381293-8.7.x.patch"
```
pfrenssen: By hacking the `composer.lock` file it is now marked as being invalid. It needs to be regenerated by running `composer update`, never be edited by hand because the checksums will no longer match.

web/profiles/joinup/joinup.post_update.php
```diff
@@ -178,3 +178,16 @@ function joinup_post_update_legal() {
...
+ * Sets the default weight for specific og actions with configuration.
+ */
+function joinup_post_update_set_default_og_action_weights() {
+  $config_factory = \Drupal::configFactory();
+  $action = $config_factory->getEditable('system.action.og_membership_remove_single_role_action.facilitator');
+  $data = $action->getRawData();
+  $data['weight'] = 1;
+
+  $action->setData($data);
+  $action->save();
```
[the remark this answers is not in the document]
idimopoulos: This is because of the `config_sync` module. The configuration files I provided in the patch were not exported up to now an most probably `config_sync` considers them as part of the configuration already since they already are. I could have reset the snapshot but this is very dangerous as it would override all other config updates from the patches included in the release.
I thought I should not waste much time here since we are planning to ditch the thing.

composer.lock
idimopoulos: I will rename the file to match also the comment number so that it gets updated properly.

PR #1756, opened 2019-08-14
ISAICP-5476: Install the 'published_at' field storage definition

claudiu-cristea:


pfrenssen:
This looks good, but isn't this affecting production right now? If this is the case we should probably cherry-pick it on the master as a hotfix.

PR #1757, opened 2019-08-14
5485: EU Login: Fix failures when creating new account

claudiu-cristea:
**WARNING** This is branched off from ISAICP-5419 and should be merged after the other one is merged in the epic.


pfrenssen:
This looks good but it is not clear to me how this should be merged in combination with the work done in #1741. My guess is that this should be merged into that PR, and then that PR into the epic.

I'm assigning this to @claudiu-cristea to handle the merge according to his insights.

PR #1760, opened 2019-08-16
ISAICP-5491: Refix the publication dates.

pfrenssen:


pfrenssen:
This seems to properly fix the published_at dates in the database. See the upstream patch for details.

However it seems that the Keep Up To Date page still doesn't show the most recent community content at the top. I am getting some news from 2018 after reindexing. I stepped through it for some time but the right published_at dates are being passed to Solr for indexing.

PR #1763, opened 2019-08-20
5453: Display a confirmation message before a membership is deleted

claudiu-cristea:


claudiu-cristea:
The remark is correct. Provided a post-update to remove the config.

PR #1767, opened 2019-08-26
ISAICP-5462: Update dependencies.

pfrenssen:


pfrenssen:
I approve the latest commit by @idimopoulos. We don't need the block_content_permissions module any more. It was removed in https://webgate.ec.europa.eu/CITnet/jira/browse/ISAICP-5348

PR #1771, opened 2019-08-29
Isaicp 5509

pfrenssen:
This is an work in progress commit from the code I wrote while investigating this issue. Obviously the test needs to be restored into its full glory. The cache context is missing, I forgot to commit this and do not have access to this machine right now.


pfrenssen:
It turns out that I actually do have access to this machine remotely :grin: I committed the missing cache context.

PR #1786, opened 2019-09-09
5350: Accept a new version of Legal Notice on EU Login

claudiu-cristea:


tests/features/eulogin/eulogin.feature
```diff
@@ -37,6 +42,13 @@ Feature: Log in through EU Login
 
     Given I select the radio button "I am a new user (create a new account)"
     When I press "Next"
+
+    Then I should see the warning message "You must accept this agreement before continuing."
+    But I should not see the success message "Fill in the fields below to let the Joinup community learn more about you!"
+
+    Given I check "I have read and accept the Legal notice"
```
pfrenssen: This is an action taken by the user that is part of the test steps, so this should have `When` instead of `Given`. 

`Given` is only used for steps that are setting up the condition of the test environment, similar to the `setUp()` method in a unit test.

tests/features/eulogin/eulogin.feature
```diff
@@ -86,6 +98,13 @@ Feature: Log in through EU Login
     Given I fill in "Email or username" with "chuck_the_local_hero"
     And I fill in "Password" with "12345"
     When I press "Sign in"
+
+    Then I should see the warning message "You must accept this agreement before continuing."
+    But I should not see the success message "Your EU Login account chucknorris has been successfully linked to your local account Chuck Norris."
+
+    Given I check "I have read and accept the Legal notice"
```
pfrenssen: Here too, it is better to use `When` instead of `Given`.

tests/features/eulogin/eulogin.feature
```diff
@@ -102,6 +121,8 @@ Feature: Log in through EU Login
     And users:
       | Username             | Password | E-mail                           | First name | Family name | Organisation |
       | chuck_the_local_hero | 12345    | chuck_the_local_hero@example.com | LocalChick | LocalNorris | ACME         |
+    # Test that the user is not redirected anymore to accept the 'Legal notice'.
+    And User "chuck_the_local_hero" accepts the "Legal notice" agreement
```
pfrenssen: This is a step used to set up the test environment so it is better to not word it like an action taken by the user, but to refer to it in a passive form. The idea of "@Given" is to describe the condition of the test environment.

Suggestion:

`And the "Legal notice" agreement has been accepted by the "chuck_the_local_hero" user`

web/modules/custom/legal/joinup_legal.behat.inc
```diff
@@ -161,17 +161,29 @@ class JoinupLegalSubContext extends DrupalSubContextBase {
...
    *   The legal document label.
+   * @param string|null $username
+   *   The name of the user that accepts the agreement.
    *
    * @throws \Exception
    *   When the current user is not logged in.
    * @throws \InvalidArgumentException
    *   When the given legal document has no published version.
    *
    * @Given I accept the :document_label agreement
```
pfrenssen: This step refers to the current user performing the test taking an action, so it is a `@When` step.

PR #1787, opened 2019-09-09
5262: Disallow local registration

claudiu-cristea:


web/modules/custom/eulogin/src/Controller/UserRegisterRedirectController.php
```diff
@@ -0,0 +1,25 @@
...
+class UserRegisterRedirectController extends ControllerBase {
+
+  /**
+   * Provides an alternative controller for 'user.register' route.
+   *
+   * @return \Drupal\Core\Routing\TrustedRedirectResponse
+   *   The redirect response.
+   */
+  public function redirectUserRegister(): TrustedRedirectResponse {
+    return new TrustedRedirectResponse($this->config('joinup_eulogin.settings')->get('register_url'));
```
pfrenssen: Since this URL is stored in config we cannot 100% rely on this value being present. Let's add a check if it is not empty and show a 404 otherwise.

PR #1802, opened 2019-09-17
5354: Convert SPDX to Joinup licences

claudiu-cristea:


claudiu-cristea:

PR #1804, opened 2019-09-18
Prequel to ISAICP-4910

pfrenssen:
The current test coverage for workflow state changes are not taking into account some things like button ordering. This PR updates test coverage to be ready for more strict changes that are going to be added in ISAICP-4910. It will allow us to verify that the changes are not affecting existing test coverage.


web/modules/custom/contact_information/tests/src/ExistingSite/ContactInformationWorkflowTest.php
```diff
@@ -85,15 +85,12 @@ public function testWorkflow(): void {
...
         // Override the user to be checked for the allowed transitions.
-        $actual_transitions = $this->workflowHelper->getAvailableTransitions($content, $this->$user_var);
-        $actual_transitions = array_map(function ($transition) {
-          return $transition->getId();
-        }, $actual_transitions);
+        $actual_states = $this->workflowHelper->getAvailableStates($content, $this->$user_var);
 
-        sort($actual_transitions);
-        sort($transitions);
+        sort($actual_states);
```
idimopoulos: In general I highly personally dislike the idea to call these `actual/expected_states` and not `actual/expected_target_states`. I know it is only one current state but given that the workflows are working with multiple `from` states to a single `target` state, in the future, for the dev that will have to get down with it, it might be quite confusing.. I would suggest to change the suffix to all these cases to either `_target_states` or `_to_states` (the `to` keyword is the one the module uses) to both the variables and the comments..

web/modules/custom/joinup_core/src/WorkflowHelper.php
```diff
@@ -79,6 +79,24 @@ public function getAvailableStatesLabels(FieldableEntityInterface $entity, Accou
     return $allowed_states;
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function getAvailableStates(FieldableEntityInterface $entity, AccountInterface $account = NULL) {
```
idimopoulos: Same as above. For someone that got down with the state_machine before jumping into this piece of code, they would probably confuse this with the retrieval of getting multiple states from the transitions. In the transitions, the multiple states list is the `from` states, so the source states. The target states in the transition is the `to` state and is only one. If you have worked a lot with state_machine, that would get confusing in your mind as, reading this method name standalone, you know that since it is in the form it probably refers to target states but your mind thinks of source states for some reason.

idimopoulos:
Thanks @pfrenssen for the changes :)

PR #1816, opened 2019-09-26
Isaicp 5558

pfrenssen:


claudiu-cristea:
This commit https://github.com/ec-europa/joinup-dev/commit/3d31369090a9bfbcb89a89eb1bfbddd77c8b070f is missed from merged 5502 branch. It seems that you forgot to pull the latest changes of 5502 or to fetch the origin.

web/profiles/joinup/config/install/geocoder.geocoder_provider.webtools_geocoding.yml
```diff
@@ -6,4 +6,5 @@ dependencies:
 id: webtools_geocoding
```
claudiu-cristea: This config is now a "default configuration" (even, is the same as the sync version). We should move it in the module where it belongs as functionality. Probably, `joinup_event`. Let's start to cleanup of the profile :)

PR #1818, opened 2019-09-27
Isaicp 5553

pfrenssen:


config/sync/block.block.group_facet.yml
```diff
@@ -9,20 +9,17 @@ dependencies:
     - joinup_core
   theme:
     - joinup_theme
-_core:
-  default_config_hash: Q-yF1_SXvQ4zJe7jAbEOjlIknalAqWN6dK4hw2rrFWw
 id: group_facet
 theme: joinup_theme
-region: filters
```
idimopoulos: What is this change of region to the configs? Why are they moved to first sidebar? Can we have a visual verification that they are not really moved to the sidebar? Or is it meant to be like this in this epic?

pfrenssen: Yes in the new advanced search the filters are in the side bar.

PR #1845, opened 2019-10-15
ISAICP-5385: Notifications of new membership with subscriptions.

idimopoulos:


config/sync/message.template.og_membership_subscribed_approve.yml
```diff
@@ -0,0 +1,19 @@
...
+dependencies: {  }
+template: og_membership_subscribed_approve
+label: 'Group membership approval with subscription'
+description: 'Group privileged user accepts a membership request that is also subscribed.'
+text:
+  -
+    value: '<p>Joinup: Your request to join the @entity:bundle @entity:title was approved</p>'
+    format: full_html
+  -
+    value: '<p>@actor:full_name has approved your request to join the "@entity:title" @entity:bundle.</p><p>You will receive weekly notifications for newly created content on this collection.<br /> To manage your notifications go to "<a href="@user:my_subscriptions">My subscriptions</a>" in the user menu.</p><p>If you think this action is not clear or not due, please contact Joinup Support at <a href="@site:contact_url">@site:contact_url</a>.</p>'
```
pfrenssen: According to the Jira ticket this should be "... has approved your request to join **and subscribe to** the...".

web/modules/custom/joinup_notification/config/install/message.template.og_membership_approve_with_subscription.yml
```diff
@@ -0,0 +1,19 @@
...
+dependencies: {  }
+template: og_membership_decision
+label: 'Group membership approval/rejection'
+description: 'Group privileged user decides on a membership request.'
+text:
+  -
+    value: "<p>Joinup: Your request to join the @entity:bundle @entity:title was approved</p>"
+    format: full_html
+  -
+    value: "<p>@actor:full_name has approved your request to join the \"@entity:title\" @entity:bundle.</p><p>You will receive weekly notifications for newly created content on this collection.<br /> To manage your notifications go to \"<a href=\"@user:my_subscriptions\">My subscriptions</a>\" in the user menu.</p><p>If you think this action is not clear or not due, please contact Joinup Support at <a href=\"@site:contact_url\">@site:contact_url</a>.</p>"
```
pfrenssen: Here also "...to join **and subscribe to** the..."

idimopoulos: Sorry, it seems I removed almost all instances of this old renaming but forgot this. The `og_membership_approve_with_subscription` is more than 32 characters and cannot be used as a message type. I had to shorten it to the `og_membership_subscribed_approve` above (which is exactly 32 characters). Removed this one.

tests/src/Traits/UtilityTrait.php
```diff
@@ -49,6 +49,36 @@ protected function formatMailBodyText(string $mail_body): string {
...
+    if (empty($user)) {
+      throw new \Exception("User {$user->getAccountName()} was not found.");
+    }
+
+    $user_email = $user->getEmail();
+    \Drupal::state()->resetCache();
+    $mails = \Drupal::state()->get('system.test_mail_collector');
+    if (empty($mails)) {
+      throw new \Exception('No mail was sent.');
+    }
```
pfrenssen: This section is duplicated in a few locations. Let's factor it out in a separate method so it can be reused.

Maybe we can make an `EmailTrait` to collect all reusable code related to inspecting emails?

web/modules/custom/joinup_user/joinup_user.behat.inc
```diff
@@ -58,6 +58,41 @@ class JoinupUserSubContext extends DrupalSubContextBase implements DrupalSubCont
...
 
+  /**
+   * Clicks an html link in the last email matching its title.
+   *
+   * @param string $title
+   *   The link title.
+   * @param string $user
+   *   The user that the email was sent to.
+   *
+   * @Given /^I click the "(?P<title>[^"]+)" link from the email sent to "(?P<user>[^"]+)"$/
```
pfrenssen: We can make this easier to read by replacing the regular expressions with the string substitutions that were introduced in Behat 3.0:

```
@Given I click the :title link from the email sent to :user
```

idimopoulos: I know... But there seems to be a compatibility problem with my PHP Storm and when I leave it only to `@Given I click the :title link from the email sent to :user` and I start typing, I only get the `@Given I click the <param>" and stops there. And I do not see the full step definition..

pfrenssen: I researched this and it turns out that the PHPStorm plugin (ref. https://plugins.jetbrains.com/plugin/9164-gherkin) is following the reference implementation of Gherkin that is supplied by the Cucumber project.

The reference implementation supports "Cucumber expressions" which are in this format:

```
@Given I click the {string} link in the email sent to {string}
```

The `:title` syntax for variable substitution that the "Behat Gherkin" implementation supports is not included in the "Cucumber Gherkin" specification (ref. https://cucumber.io/docs/cucumber/step-definitions/ and https://cucumber.io/docs/cucumber/cucumber-expressions/).

The regex variant is supported by both. Unfortunately it seems this is not working as expected in PHPStorm either: https://youtrack.jetbrains.com/issue/WI-25135

Also the regex variant is a nightmare to review. The fact that this is copied from some other part of the code does not make me trust it :)

I am quite fluent with regular expressions but I had to google the `?P` part (it turns out to be a _symbolic group match_ ref. https://docs.python.org/3/library/re.html#index-17). The `:title` syntax on the other hand is very clear and easy to review.

Now I think that we should not make decisions based on whether or not something works in PHPStorm. We do not require people to use this proprietary software for working on Joinup. There are 2 possibilities: either Jetbrains can include support for the "Behat Gherkin" syntax in their proprietary Behat plugin, or the "Behat Gherkin" implementation can add support for the "Cucumber Expressions". Both of these solutions would make it work out of the box for everyone. I will make a feature request to support this in Behat.

Now, on the other hand now that I have fully reviewed the regular expression with the symbolic group match I know it is correct, so I have no reason to reject it any more. It can stay :)

pfrenssen: I did some more research and the "Cucumber expressions" syntax results in the same problems as the "Behat syntax". Only the "regex syntax" works in PHPStorm. It has been reported in their bug tracker: https://youtrack.jetbrains.com/issue/WI-49121

idimopoulos: According to the above, I will leave it with the `:parameter` syntax and hope it is fixed upstream soon.

tests/src/Traits/UtilityTrait.php
```diff
@@ -3,13 +3,82 @@
...
+   *   The user name.
+   *
+   * @return array
+   *   An array of mails as stored in the email collector.
+   *
+   * @throws \Exception
+   *   Thrown if the user is not found or no emails have been sent.
+   */
+  protected function getUserMails(string $user): array {
+    $user = $this->getUserByName($user);
```
pfrenssen: Let's try not to make one trait depend on another trait unless it is unavoidable. In this case an easy solution would be to let this method accept a `UserInterface $user` object, so the responsibility of including the `UserTrait` is held by the calling code.

tests/src/Traits/UtilityTrait.php
```diff
@@ -3,13 +3,82 @@
...
+    if (empty($user)) {
+      throw new \Exception("User {$user->getAccountName()} was not found.");
+    }
```
pfrenssen: The `$user->getAccountName()` call will throw a fatal error in this case, since `$user` is empty.

PR #1856, opened 2019-10-22
ISAICP-5623: Provide a command to flush redis.

idimopoulos:


scripts/update_production.sh
```diff
@@ -38,6 +38,7 @@ fi
 echo "Disabling config_readonly."
 touch disable-config-readonly
 
+./vendor/bin/phing redis-flush
```
claudiu-cristea: We'll have to chain with AND operator the commands so that the failure of one makes the whole process fail.

build.redis.xml
```diff
@@ -30,4 +33,10 @@
             </else>
         </if>
     </target>
+
+    <target name="redis-flush">
+        <echo msg="Cleaning redis flush." />
```
claudiu-cristea: Not so sure about the phrasing. Aren't "clean" and "flush", as verbs, the same in this context? Also "Redis" is a name and should be "word uppercased". Maybe "Flushing Redis backend"?

src/Phing/RedisFlush.php
```diff
@@ -0,0 +1,28 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public function main(): void {
+    try {
+      ClientFactory::getClient()->flushAll();
+      $this->log('Redis cache flushed.');
+    }
+    catch (\Exception $e) {
+      $this->log('Error flushing redis cache: ' . $e->getMessage(),\Project::MSG_WARN);
```
claudiu-cristea: Need a space after comma. Also, s/redis/Redis

src/Phing/RedisFlush.php
```diff
@@ -0,0 +1,28 @@
...
+    }
+
```
claudiu-cristea: Extra new line (I wonder why PHPCS didn't catch the coding standards violations!!!)

PR #1857, opened 2019-10-22
ISAICP-5582: Implement compare checkboxes and button

tom-waat:


claudiu-cristea:
For some reasons the test are showed for commit 82e7277. Here they are app.continuousphp.com/git-hub/ec-europa/joinup-dev/build/5052f61d-cd2c-4f1f-b86e-b99ebe11b935

PR #1859, opened 2019-10-22
Isaicp 4980

pfrenssen:
This is intended for ISAICP-2242 but it blocked by ISAICP-4273. The base branch is set to ISAICP-4273 to make it easy to review. This needs to be switched back to ISAICP-2242 when accepted.


tests/src/Traits/EntityTrait.php
```diff
@@ -55,6 +55,9 @@ protected static function getEntityByLabel(string $entity_type_id, string $label
 
     if ($result) {
       $result = reset($result);
+      // Make sure we get a fresh entity from the database, to avoid testing
+      // with stale data.
+      $storage->resetCache([$result]);
```
idimopoulos: No sorry, I disagree with this point right here. This is to avoid working with stale data when we shouldn't have stale data. What about tests with proper cache clear? We have so many test that we have created to ensure that cache is being cleared. If we always reset the cache right before we load an entity, we cannot be sure that these tests are going to work fine.

pfrenssen: I don't get your reservation?

This is in Behat, this is for functional tests. If we have a step that checks that a certain entity is present for the user scenario, we want to make sure we get the current one from the database. Not some entity that was cached in the Behat process in a prior step.

PR #1860, opened 2019-10-23
ISAICP-4778: Share solutions in collections

idimopoulos:


web/modules/custom/joinup_core/src/EventSubscriber/JoinupCoreOgSubscriber.php
```diff
@@ -0,0 +1,42 @@
...
+   */
+  public function provideOgGroupPermissions(OgPermissionEventInterface $event) {
+    $event->setPermission(
+      new GroupPermission([
+        'name' => 'administer shared entities',
+        'title' => $this->t('Administer shared entities'),
+        'restrict access' => TRUE,
+      ])
+    );
+  }
```
pfrenssen: This entire file seems to be unneeded and duplicating the code in `\Drupal\joinup_community_content/EventSubscriber/EventSubscriber`. I tried to comment out this code and the test still completes without problems. This seems to make sense since this subscriber is not defined in `joinup_core.services.yml` so it is never called.

web/modules/custom/solution/src/EventSubscriber/SolutionOgSubscriber.php
```diff
@@ -0,0 +1,51 @@
...
+      ]),
+      new GroupContentOperationPermission([
+        'name' => "share solution rdf_entity",
+        'title' => $this->t('solution: Unshare from a group'),
+        'operation' => 'unshare',
+        'entity type' => 'rdf_entity',
+        'bundle' => 'solution',
+      ]),
+    ]);
+  }
```
pfrenssen: This file also seems unused. The test passes without it.

web/modules/custom/joinup_core/joinup_core.module
```diff
@@ -1352,3 +1354,37 @@ function joinup_core_module_implements_alter(array &$implementations, $hook) {
...
+ * @param string $route
+ *   The route to pass the share link in.
+ *
+ * @return array
+ *   A link icon theme array.
+ */
+function joinup_core_generate_share_link(EntityInterface $entity, string $route): array {
+  $url = Url::fromRoute('joinup_community_content.share_content_form', [
+    'node' => $entity->id(),
+  ]);
```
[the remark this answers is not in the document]
idimopoulos: This was a bug actually that was fixed in one of the followups and is now fixed. Please, re-check. That was indeed a bad miss on my side.

web/modules/custom/joinup_core/src/Form/ShareForm.php
```diff
@@ -171,55 +180,74 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
...
-   *   A list of collections where the current node can be shared.
+   *   A list of collections where the current entity can be shared in.
    */
   protected function getShareableCollections() {
-    // If the node has no field, it's not shareable anywhere.
-    if (!$this->node->hasField('field_shared_in')) {
+    // Being part also for the access check, do not allow the user to access
+    // this page for entities without a field to store collections it is shared
+    // in.
+    if (!$this->entity->hasField($this->getSharedInFieldName())) {
```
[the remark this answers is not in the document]
idimopoulos: I will set this to `FieldableEntityInterface` as the entity might be an Rdf entity to so `ContentEntityInterface` would not be valid.

web/modules/custom/joinup_core/src/EventSubscriber/JoinupCoreOgSubscriber.php
idimopoulos: The `administer shared entities` was removed from the `joinup_community_content`'s event subscriber as now, solutions can also be shared, thus the permission belongs to a higher level module like `joinup_core`. Leaving the necessity to our private chat and will report back here.

PR #1866, opened 2019-10-29
ISAICP-5647: Update the values in the contact form

idimopoulos:


idimopoulos:

web/modules/custom/joinup_core/joinup_core.install
```diff
@@ -460,3 +460,29 @@ function joinup_core_update_8113(array &$sandbox): ?string {
 
   return "Finished collecting {$sandbox['progress']} node visit counts.";
 }
+
+/**
+ * Update contact form categories.
+ */
+function joinup_core_update_8114(array &$sandbox) {
```
claudiu-cristea: Is there a reason why this update function belongs to `joinup_core` rather than `contact_form`?

PR #1882, opened 2019-11-11
ISAICP-5607: Send a direct email for comment creation.

idimopoulos:


web/modules/custom/joinup_discussion/src/EventSubscriber/SubscribedDiscussionCommentSubscriber.php
```diff
@@ -179,6 +178,9 @@ protected function getRecipients(): array {
       if (!$this->comment->getOwner()->isAnonymous()) {
         unset($this->recipients[$this->comment->getOwnerId()]);
       }
+      $this->recipients = array_filter($this->recipients, function (UserInterface $recipient) {
```
idimopoulos: Disclaimer: This is mainly fixing the tests as there are cases where we create a test without an owner.
In the previous version of the implementation, the same check was taking place for all recipients regardless, and so the tests were passing.
Now, the recipients are all together and I put it here in the end of the calculation.
The other way would be to always create users and assign them as authors in the tests.

pfrenssen: I looked into this and the anonymous user is not in fact subscribed. It is added a few lines earlier in this method. We can rework this to simply not add the user in the first place, this is better than adding it and then removing it again.

I will also add an `assert()` to `\Drupal\joinup_subscription\JoinupDiscussionSubscriptionInterface::getSubscribers()` to ensure we will not subscribe anonymous users in tests.

PR #1888, opened 2019-11-18
ISAICP-4254

pfrenssen:


pfrenssen:
The `discussion.subscribe.feature` test is freezing. We'll need to investigate what is wrong with it.

pfrenssen:
There is a limitation in the Message Digest module that prevents us from applying specific theming to the collection content subscription digest and not to the discussion comment subscription digest. We need to be able to set a different "key" for each digest. The key is currently hardcoded to `digest` in `DigestManager::deliverDigest()`.

web/modules/custom/joinup_subscription/joinup_subscription.behat.inc
```diff
@@ -44,6 +44,15 @@ class JoinupSubscriptionSubContext extends DrupalSubContextBase implements Drupa
   use TagTrait;
   use TraversingTrait;
 
+  /**
+   * A list of the available digest message intervals.
+   *
+   * This is hardcoded instead of being retrieved dynamically from the config so
+   * that our tests will detect it if one of these intervals would unexpectedly
+   * be removed from the code base.
```
claudiu-cristea: 💖

web/modules/custom/joinup_subscription/joinup_subscription.module
```diff
@@ -86,9 +86,17 @@ function joinup_subscription_message_digest_aggregate_alter(array &$context, Use
...
  */
 function joinup_subscription_mail_alter(array &$message): void {
   if ($message['key'] === 'digest') {
     $message['headers']['Content-Type'] = SWIFTMAILER_FORMAT_HTML;
+    $notifier = $message['params']['notifier'] ?? NULL;
+    if ($notifier instanceof MessageNotifierInterface) {
+      $interval = $notifier->getPluginDefinition()['id'];
+      $message['subject'] = t('Joinup: @interval digest message', [
+        '@interval' => $interval,
+      ]);
```
claudiu-cristea: This wrong. The plugin ID is not translatable and it's an ID, it may have a value that has no meaning for a human. We should use the plugin's title as that is extracted from the `message_digest_interval` config entity and being a label is translatable. See `DigestDeriver::getDerivativeDefinitions()`.

PR #1892, opened 2019-11-20
ISAICP-4620: Restructure code to more closely represent the data model

pfrenssen:


pfrenssen:
These are just some small improvements I did when I was working on the spike. Let's get them in anyway so this work is not wasted :)

PR #1897, opened 2019-11-24
ISAICP-5483: Allow facilitators to delete contact information.

idimopoulos:


web/modules/custom/contact_information/contact_information.module
```diff
@@ -112,29 +112,50 @@ function contact_information_entity_auto_label(EntityInterface $entity) {
...
-    return AccessResult::neutral();
-  }
-  // Only revoke access for anonymous users.
-  if ($account->isAuthenticated()) {
-    return AccessResult::neutral();
+
+  // Inline entity form module is using the 'update' as a name for the edit
+  // operation as it is with all entities apart from nodes that retained the
+  // 'edit' permission for legacy reasons. However, rdf entities also use this
+  // naming convention. Rename 'update' to 'edit' when inline_entity_form checks
```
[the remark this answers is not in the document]
pfrenssen: Created issue in RDF Entity to deprecate passing in `edit`: https://github.com/ec-europa/rdf_entity/pull/107

I would suggest to update our code to pass in `update`.

web/modules/custom/contact_information/contact_information.module
```diff
@@ -112,29 +112,50 @@ function contact_information_entity_auto_label(EntityInterface $entity) {
...
+  elseif ($operation == 'edit') {
+    $allowed_transitions = \Drupal::service('joinup_core.workflow.helper')->getAvailableTargetStates($entity, $account);
+    return empty($allowed_transitions) ? AccessResult::forbidden() : AccessResult::allowed();
   }
-  if ($entity->id() == $route_entity->id()) {
-    return AccessResult::forbidden('Anonymous users should not see contact information on the canonical route.');
+  // Use an elseif to avoid checking memberships for anonymous users.
+  elseif ($operation === 'delete') {
+    // The contact information is not a group content rather than it is simply
+    // referenced by a collection or a solution.
```
pfrenssen: IIRC contact information can be referenced by multiple collections and solutions. This documentation seems to hint that it can only be referenced by a single group.

web/modules/custom/contact_information/contact_information.module
```diff
@@ -112,29 +112,50 @@ function contact_information_entity_auto_label(EntityInterface $entity) {
...
+    if ($groups = $relation_manager->getContactInformationRelatedGroups($entity)) {
+      /** @var \Drupal\og\MembershipManagerInterface $membership_manager */
+      $membership_manager = \Drupal::service('og.membership_manager');
+      foreach ($groups as $group) {
+        /** @var \Drupal\og\Entity\OgMembership $membership */
+        $membership = $membership_manager->getMembership($group, $account->id());
+        if ($membership && $membership->hasPermission("delete contact_information rdf entity")) {
+          return AccessResult::allowed();
+        }
+      }
```
pfrenssen: But, this means that if a user has access to delete a contact information from a single collection or solution, then it can delete the contact information for all the groups that reference it?

What happens in this case? Is the contact information removed for all groups?

web/modules/custom/contact_information/src/Plugin/WorkflowStatePermission/ContactInformationWorkflowStatePermission.php
```diff
@@ -75,10 +112,72 @@ public function isStateUpdatePermitted(AccountInterface $account, EntityInterfac
     }
 
     $allowed_conditions = $this->configFactory->get('contact_information.settings')->get('transitions');
+    $matrix = $allowed_conditions[$to_state][$from_state];
+    $access = FALSE;
+
+    if (!empty($matrix) && $this->userHasOwnAnyRoles($entity, $account, $matrix)) {
+      $access = TRUE;
+    }
```
pfrenssen: This can be replaced with:

```
$access = !empty($matrix) && $this->userHasOwnAnyRoles($entity, $account, $matrix);
```

web/modules/custom/joinup_core/src/JoinupRelationManager.php
```diff
@@ -219,25 +227,60 @@ public function getSolutionIds(): array {
...
+      }
+      catch (InvalidPluginDefinitionException $e) {
+        throw new \RuntimeException('The RDF entity storage is not valid.');
+      }
+      catch (PluginNotFoundException $e) {
+        throw new \RuntimeException('The RDF entity storage is not defined.');
+      }
+    }
+    return $this->sparqlStorage;
+  }
```
pfrenssen: Don't provide a static cache for the storage, it is already statically cached inside the `EntityTypeManager`. Having this here is a potential source of very difficult to track caching issues.

This not only increases our memory footprint without being necessary, but this not be updated when `EntityTypeManager::clearCachedDefinitions()` is called and will return stale results. Also this does not respect `EntityTypeManager::useCaches(FALSE)` which is called during dangerous operations like update hooks and entity storage schema changes.

web/modules/custom/joinup_core/src/JoinupRelationManager.php
```diff
@@ -35,6 +36,13 @@ class JoinupRelationManager implements JoinupRelationManagerInterface, Container
    */
   protected $entityTypeManager;
 
+  /**
+   * The sparql entity storage.
+   *
+   * @var \Drupal\sparql_entity_storage\SparqlEntityStorageInterface
+   */
+  protected $sparqlStorage;
```
pfrenssen: This static cache should be removed, see the comment below at `::getRdfStorage()` for a motivation why this is not a good idea.

tests/features/contact_information/contact_information.moderation.feature
```diff
@@ -94,17 +104,50 @@ Feature: Contact Information moderation
...
-    And I go to the "Æthelred, Lord of the Mercians" contact information page
-    When I click "Edit" in the "Entity actions" region
-    Then I should see the link "Delete"
+    And I should see the link "Delete"
     When I click "Delete"
     # Confirm the deletion.
     And I press "Delete"
     Then I should not see the link "EU healthy group"
+
+  Scenario: Owners can request deletion when they are not facilitators and facilitators can delete.
```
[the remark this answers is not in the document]
idimopoulos: There is this scenario actually. I though of this afterwards. User A creates a solution within a collection. He is now the owner (also the facilitator) of the solution.
After a while, he transfers the ownership of the solution to someone else and his membership is revoked. 
The contact information entity is still his.
I know this is a very very rare case but this is a case that can be reproduced in the UI.
However, the steps on the UI don't need to be manually set up and have the whole process gradually taking place (create solution, transfer ownership, login as the new owner, delete  membership, try to edit as owner etc). Instead, I simply assign the correct roles.

web/modules/custom/contact_information/contact_information.module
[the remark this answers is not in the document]
idimopoulos: We need a solid plan to update to `update` operation instead of `edit`.
The rdf route is still requesting an edit permission. In https://github.com/ec-europa/rdf_entity/pull/107/files we deprecated the 'edit' permission but the route definition is still
```
entity.rdf_entity.edit_form:
  path: '/rdf_entity/{rdf_entity}/edit'
  defaults:
    _entity_form: rdf_entity.edit
    _title: 'Edit Rdf'
  requirements:
    _entity_access: 'rdf_entity.edit'
```
That last part is requesting the `edit` operation.

web/modules/custom/contact_information/contact_information.module
idimopoulos: I have raised this question to the Functional Contractor. They will decide what happens in this case. Probably they should not be able to delete these entities. This should be available only to moderators. However, this refers only to very old contact information entities. It seems that we have a small number (36) that belong to multiple groups. It will indeed be weird to have them deleted even by accident, let alone have this break the website.

web/modules/custom/contact_information/contact_information.module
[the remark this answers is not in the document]
idimopoulos: I updated the comment actually. I was wrong to say that this is the case for nodes but the code is still needed until the operations are updated in rdf entity.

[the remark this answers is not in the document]
pfrenssen: OK this will start throwing deprecation warnings as soon as the next version of RDF Entity is released, we can then update it.

web/modules/custom/contact_information/contact_information.module
pfrenssen: OK so the new expected behavior as defined by the functional team is that we will only consider the first collection or solution that was associated with a contact information, and ignore any subsequent ones. There are a small number of old collections and solutions that will be affected if their contact information is deleted, but it is decided that the risk is very small and can be rectified manually by moderators by creating new contact information, if this issue is raised through support channels.

I personally would have preferred a solution that safeguards the data integrity without requiring manual intervention but I respect the decision of the functional team, and can confirm this is now implemented according to the requested specification.

PR #1898, opened 2019-11-24
ISAICP-5405: Provide a kickoff rdf mappign for the news entity.

idimopoulos:


idimopoulos:
This solution has been abandoned!

PR #1902, opened 2019-11-28
ISAICP-5590: Update dependencies

claudiu-cristea:


composer.json
```diff
@@ -5,74 +5,69 @@
...
-        "composer/composer": "^1.7.1",
-        "composer/installers": "^1.2",
-        "continuousphp/aws-sdk-phing": "~0.1",
-        "cweagans/composer-patches": "^1.6.4",
-        "drupal-composer/drupal-scaffold": "^2.2",
-        "drupal/allowed_formats": "^1.1",
+        "composer/composer": "^1.9.1",
+        "composer/installers": "^1.7",
+        "continuousphp/aws-sdk-phing": "~0.5",
+        "cweagans/composer-patches": "^1.6",
```
pfrenssen: `cweagans/composer-patches` has regressed from 1.6.4 to 1.6.0, I'm assuming this is not intended. There must have been a particular reason for us to have picked 1.6.4 previously, it probably contains a bug fix that affected us.

claudiu-cristea: Fixed.

PR #1905, opened 2019-11-29
ISAICP-5676: Update the account descriptions for eu login account details.

idimopoulos:


web/modules/custom/joinup_user/joinup_user.module
```diff
@@ -69,7 +69,7 @@ function joinup_user_user_update(UserInterface $account) {
     $account->getPreferredLangcode(),
     $params
   );
-  drupal_set_message(t('An e-mail has been send to the user to notify him on the change to his account.'));
+  drupal_set_message(t('An e-mail has been sent to the user to notify him on the change to his account.'));
```
claudiu-cristea: If we already touched this line why not dropping the deprecated `drupal_set_message()`?

web/modules/custom/eulogin/joinup_eulogin.module
```diff
@@ -108,7 +108,22 @@ function joinup_eulogin_form_user_form_alter(array &$form, FormStateInterface $f
...
+    '#type' => 'html_tag',
+    '#tag' => 'h3',
+    '#value' => t('EU Login Account information'),
+  ];
+  $form['header_description'] = [
+    '#weight' => -99,
+    '#type' => 'html_tag',
+    '#tag' => 'span',
+    '#value' => t('Your name and E-mail data is inherited from EU Login. To update this information, you can visit your EU Login account page <a href="https://webgate.ec.europa.eu/cas/eim/external/restricted/edit.cgi" rel="nofollow" target="_blank">here</a>. Synchronisation will take a few minutes and it will be visible the next time you login on Joinup.'),
+  ];
```
claudiu-cristea: This is not resolving the separation of "EU Login Account information" block from the rest of the page as it is shown in the ticket's description mockup. The decision was to implement field groups. Also, we have config for service URLs (see the `register_url` key from `joinup_eulogin.settings.yml`. Let's create a similar `edit_url` instead of hardcoding it here.

claudiu-cristea: We keep the actual solution as it has been discussed. Probably we need a frontend followup.

web/modules/custom/eulogin/joinup_eulogin.module
```diff
@@ -119,6 +132,8 @@ function joinup_eulogin_form_user_form_alter(array &$form, FormStateInterface $f
   if (isset($form['account']['mail']) && \Drupal::config('cas.settings')->get('user_accounts.restrict_email_management')) {
     $form['account']['mail']['#disabled'] = TRUE;
   }
+  $form['field_user_first_name']['widget'][0]['value']['#description'] = t('Your first name is publicly visible.');
+  $form['field_user_family_name']['widget'][0]['value']['#description'] = t('Your last name is publicly visible.');
```
claudiu-cristea: I fear that we'll forget to cleanup the code when EU Login will the only login method. Normally, such a hook alter is used to override a default behaviour, which is now in config. But this description is the one that will survive after February 2020. So, instead of overriding here, I'm suggesting that we move this text in config now and override the text only for non-CAS users here. Then we only need to remove these lines in Feb 2020.

PR #1913, opened 2019-12-03
ISAICP-5496: Don't allow deletion of collections with children solutions

idimopoulos:


web/modules/custom/joinup_group/joinup_group.module
```diff
@@ -52,6 +53,11 @@ function joinup_group_entity_type_alter(array &$entity_types) {
   // Swap the default user cancel form implementation with a custom one that
   // prevents deleting users when they are the sole owner of a group.
   $entity_types['user']->setFormClass('cancel', 'Drupal\joinup_group\Form\UserCancelForm');
+
+  // Override the delete form to disallow deletion of groups with children.
+  /** @var \Drupal\Core\Entity\ContentEntityTypeInterface $rdf_entity */
+  $rdf_entity = $entity_types['rdf_entity'];
```
claudiu-cristea: Should be `$rdf_entity_type`?

web/modules/custom/joinup_group/src/Form/RdfDeleteForm.php
```diff
@@ -0,0 +1,77 @@
...
+ */
+class RdfDeleteForm extends OriginalForm {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getQuestion(): TranslatableMarkup {
+    $entity = $this->getEntity();
+    if ($entity->bundle() !== 'collection' || $entity->get('field_ar_affiliates')->isEmpty()) {
+      return $this->t('Are you sure you want to delete entity %name?', [
```
claudiu-cristea: We can do more, we can be more specific as now we have the undefined singular label. The word "entity" is too vague for an end-user: `Are you sure you want to delete @type %name?`. Then `@type` is set to `$entity->get('rid')->entity->getSingularLabel()`.

web/modules/custom/joinup_group/src/Form/RdfDeleteForm.php
```diff
@@ -0,0 +1,77 @@
...
+  public function buildForm(array $form, FormStateInterface $form_state): array {
+    $form = parent::buildForm($form, $form_state);
+
+    /** @var \Drupal\rdf_entity\RdfInterface $entity */
+    $entity = $this->getEntity();
+    if (empty($entity) || $entity->isNew() || $entity->bundle() !== 'collection') {
+      return $form;
+    }
+
+    $affiliates = $entity->get('field_ar_affiliates')->referencedEntities();
```
claudiu-cristea: If we exit the method one line below, it doesn't make sense to compute the referenced entities as that is expensive because in needlessly loading entities. Instead we can use the `->isEmpty()` method and get the referenced entities, later, after `if (..)`.

web/modules/custom/joinup_group/src/Form/RdfDeleteForm.php
```diff
@@ -0,0 +1,77 @@
...
+   */
+  public function getQuestion(): TranslatableMarkup {
+    $entity = $this->getEntity();
+    if ($entity->bundle() !== 'collection' || $entity->get('field_ar_affiliates')->isEmpty()) {
+      return $this->t('Are you sure you want to delete entity %name?', [
+        '%name' => $entity->label(),
+      ]);
+    }
+
+    return $this->t('The collection @collection cannot be deleted because it contains the following solutions:', [
```
claudiu-cristea: Let's `s/@collection/%collection` to be more readable.

tests/features/joinup_group/collection.delete_with_children.feature
```diff
@@ -0,0 +1,29 @@
+@api
+Feature:
+  In order to not leave orphaned entities
+  As a collection owner
+  I need to be unable to delete collections that have children entities.
+
+  @terms
```
claudiu-cristea: Why do we need terms to run this test?

PR #1917, opened 2019-12-04
ISAICP-5716: Assign affiliations to the appropriate graph.

idimopoulos:


web/modules/custom/solution/src/SolutionAffiliationFieldItemList.php
```diff
@@ -169,4 +206,61 @@ protected function solutionInOfficialGraph(): bool {
...
 
+  /**
+   * Returns the graph ids that the collection exists in.
+   *
+   * @return array
+   *   An array of graph URIs that the collection exists in indexed by graph id.
+   */
+  protected function getAvailableGraphs(): array {
+    $graph_handler = $this->getGraphHandler();
+    $graph_ids = $graph_handler->getEntityTypeGraphIds('rdf_entity');
```
claudiu-cristea: Shouldn't we restrict to the official graphs? This implementation grabs also the `staging` graph but there's already a policy in place that for non-official graphs the code that use them is responsible with ensuring the solution <-> collection relation.

web/modules/custom/solution/src/SolutionAffiliationFieldItemList.php
```diff
@@ -102,55 +127,67 @@ protected function getAffiliation(): array {
...
+      }
 
-    $new_ids = array_map(function (EntityReferenceItem $field_item): string {
-      return $field_item->target_id;
-    }, $this->list);
-    sort($new_ids);
+      $new_ids = array_map(function (EntityReferenceItem $field_item): string {
+        return $field_item->target_id;
+      }, $this->list);
+      sort($new_ids);
```
claudiu-cristea: It seems that this code block is computing the same `$new_ids` each iteration. Could we move it outside `foreach () {...}`?

web/modules/custom/solution/src/SolutionAffiliationFieldItemList.php
```diff
@@ -102,55 +127,67 @@ protected function getAffiliation(): array {
...
-    $insert[] = "WITH <{$graph_uri}>";
-    $insert[] = 'INSERT {';
-    foreach ($new_ids as $id) {
-      $insert[] = "  <{$id}> <{$field_uri}> <{$solution->id()}> .";
+      // Delete existing affiliation.
+      $delete = [];
+      $delete[] = "WITH <{$graph_uri}>";
+      $delete[] = "DELETE { ?id <{$field_uri}> <{$solution->id()}> . }";
+      $delete[] = "WHERE { ?id <{$field_uri}> <{$solution->id()}> . }";
+      $connection->query(implode("\n", $delete));
```
claudiu-cristea: As I performance improvement, I would run a single query after `foreach (...) {...}` across all graphs.

web/modules/custom/solution/src/SolutionAffiliationFieldItemList.php
```diff
@@ -102,55 +127,67 @@ protected function getAffiliation(): array {
...
+
+      // Insert new affiliation.
+      $insert = [];
+      $insert[] = "WITH <{$graph_uri}>";
+      $insert[] = 'INSERT {';
+      foreach ($new_ids as $id) {
+        $insert[] = "  <{$id}> <{$field_uri}> <{$solution->id()}> .";
+      }
+      $insert[] = '}';
+      $connection->query(implode("\n", $insert));
```
claudiu-cristea: Same here. I think there's a SPARQL `INSERT` syntax that allows inserting multiple values in multiple graphs. Should be something like:


```
INSERT DATA {
  GRAPH <http://graph1> {
    <a1> <b1> <c1> .
    <a2> <b2> <c2> .
    ...
  }
  GRAPH <http://graph2> {
    <a1> <b1> <c1> .
    <a2> <b2> <c2> .
    ...
  }
  ...
} 
```

Then we use the `foreach (...) {...}` only to construct the query blob and we run a single query after `foreach (...) {...}`.

claudiu-cristea:
Also, we need a post-update script that scans all the databases and makes sure that, if there are solutions and collections with both graphs they are în sync and are also fixing the orphan triples.

web/modules/custom/solution/src/SolutionAffiliationFieldItemList.php
```diff
@@ -169,4 +206,61 @@ protected function solutionInOfficialGraph(): bool {
...
+    return $this->connection;
+  }
+
+  /**
+   * Loads and returns the graph handler service.
+   *
+   * @return \Drupal\sparql_entity_storage\SparqlEntityStorageGraphHandlerInterface
+   *   The graph handler service.
+   */
+  protected function getGraphHandler(): SparqlEntityStorageGraphHandlerInterface {
```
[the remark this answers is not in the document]
idimopoulos: In an earlier comment from @pfrenssen, he mentioned that the services are already cached within the container so further caching the services in the class's properties is kinda wrong. I am going to remove this and also remove the rest of the instances.

web/modules/custom/solution/src/SolutionAffiliationFieldItemList.php
```diff
@@ -102,55 +127,67 @@ protected function getAffiliation(): array {
...
+      }, $this->list);
+      sort($new_ids);
 
-    // The affiliation has been preserved. Exit here.
-    if ($new_ids === $existing_ids) {
-      return;
-    }
+      // The affiliation has been preserved. Exit here.
+      if ($new_ids === $existing_ids) {
+        return;
```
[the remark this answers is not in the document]
idimopoulos: You are completely right. I think this was there from before when we were checking the specific graph and I missed this. Thanks.

web/modules/custom/solution/src/SolutionAffiliationFieldItemList.php
idimopoulos: I had troubles in the past with multiple `WITH` statements. It seemed that it was accepting the query as valid but was running the query across ALL graphs instead of the ones in the `WITH` restriction. If possible, I would keep it isolated to a single graph.

web/modules/custom/solution/src/SolutionAffiliationFieldItemList.php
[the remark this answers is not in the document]
idimopoulos: This was reverted to remain a return since the foreach loop is now removed. We do not need to iterate through all graphs and run a separate query. We can simply go with a single query for this.

web/modules/custom/solution/src/SolutionAffiliationFieldItemList.php
idimopoulos: I tried to merge in the two queries but was a failure because the delete query works with a variable for an entity id but in the above way the insert works with a direct uri. So the delete and the insert are separated..

web/modules/custom/solution/src/SolutionAffiliationFieldItemList.php
```diff
@@ -169,4 +173,111 @@ protected function solutionInOfficialGraph(): bool {
...
+   * Returns the graph ids that the collection exists in.
+   *
+   * @return array
+   *   An array of graph URIs that the collection exists in indexed by graph id.
+   */
+  protected function getAvailableGraphs(): array {
+    $graph_handler = \Drupal::service('sparql.graph_handler');
+    $return = [];
+
+    foreach (['default', 'draft'] as $graph_id) {
```
claudiu-cristea: We should not hardcode the "official" graph list. Graph handler has dedicated method.

PR #1918, opened 2019-12-05
ISAICP-5563: Max filesize of 5MB for logos.

idimopoulos:


pfrenssen:
Looking good but I was curious why there isn't a test. I see in the commits that there originally was a check in the edit form for the text "5 MB limit". Was it not feasible to check if this help text belongs to the correct field?

idimopoulos:
> Looking good but I was curious why there isn't a test. I see in the commits that there originally was a check in the edit form for the text "5 MB limit". Was it not feasible to check if this help text belongs to the correct field?

We discussed this yesterday in the standup. Originally there were tests that were passing locally. On my local machine, the max upload file PHP limit is >5MB so after setting the limit to the field, the limit was indeed 5MB. The tests, however, were failing upstream. Apparently, on the bot, the limit is less than 5MB and thus, the text "5MB" could not be found. I offered to create a test that checks that the limit is less or equal to 5MB but claudiu thought that we don't need to go that far as this is merely Drupal Core basic field configuration.

pfrenssen:
Ok I am satisfied with this, this change is too minor to warrant a big effort on test coverage.

PR #1922, opened 2019-12-10
ISAICP-5655: Do not allow deletion of account before transfering solution.

idimopoulos:


web/modules/custom/joinup_core/src/JoinupRelationManagerInterface.php
```diff
@@ -128,7 +128,7 @@ public function getUserMembershipsByRole(AccountInterface $user, string $role, a
    * @return \Drupal\rdf_entity\Entity\Rdf[]
    *   An array of collections.
    */
-  public function getCollectionsWhereSoleOwner(AccountInterface $user): array;
+  public function getGroupsWhereSoleOwner(AccountInterface $user): array;
```
claudiu-cristea: Now that we have the `joinup_group` module, I think is worth to move this method there, in a new service, something like `JoinupGroupManager`, or any other better naming.

web/profiles/joinup/src/Form/UserCancelForm.php
```diff
@@ -61,45 +60,54 @@ public static function create(ContainerInterface $container) {
    * {@inheritdoc}
    */
   public function buildForm(array $form, FormStateInterface $form_state) {
```
claudiu-cristea: As we're changing a lot here, let's move the whole class in `joinup_group` where this really belongs. Also the corresponding part from `joinup_entity_type_alter()` should follow there.

web/profiles/joinup/src/Form/UserMultipleCancelConfirm.php
```diff
@@ -64,30 +63,40 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
-      $collections = $this->relationManager->getCollectionsWhereSoleOwner($account);
+      $groups = $this->relationManager->getGroupsWhereSoleOwner($account);
 
-      if ($collections) {
+      if ($groups) {
         $build[$account->id()] = [
           'warning' => [
-            '#markup' => $this->t('User @name cannot be deleted as it is currently the sole owner of these collections:', [
+            '#markup' => $this->t('User @name cannot be deleted as they are currently the sole owner of these groups:', [
               '@name' => $account->getAccountName(),
```
claudiu-cristea: This wrongly uses `->getAccountName()` which shows the username. But we want to show the real name, as we do everywhere on the site: `->getDisplayName()`. Should we change?

web/profiles/joinup/src/Form/UserCancelForm.php
```diff
@@ -61,45 +60,54 @@ public static function create(ContainerInterface $container) {
...
+        $bundle_type = $rdf_storage->load($bundle);
+        if (!empty($group_data[$bundle])) {
+          $form[$bundle] = [
+            '#theme' => 'item_list',
+            '#items' => $group_data[$bundle],
+            '#title' => $bundle_type->getCountLabel(count($group_data[$bundle])),
+            '#weight' => 0,
+          ];
+        }
+      }
```
claudiu-cristea: Let's move this part in a trait, as is duplicated code, and reuse it in both: `UserMultipleCancelConfirm` and `UserCancelForm`.

web/profiles/joinup/src/Form/UserMultipleCancelConfirm.php
```diff
@@ -64,30 +63,40 @@ public function buildForm(array $form, FormStateInterface $form_state) {
     $form = parent::buildForm($form, $form_state);
```
claudiu-cristea: Same, we should move the class in the right place. Also the corresponding part from `\Drupal\joinup\Routing\RouteSubscriber::alterRoutes()`.

web/profiles/joinup/src/Form/UserMultipleCancelConfirm.php
```diff
@@ -64,30 +63,40 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
+          if (!empty($group_data[$bundle])) {
+            $build[$account->id()][$bundle] = [
+              '#theme' => 'item_list',
+              '#title' => $bundle_type->getCountLabel(count($group_data[$bundle])),
+              '#items' => $group_data[$bundle],
+            ];
+          }
+        }
       }
     }
```
claudiu-cristea: To be reused from the trait (see above).

web/profiles/joinup/src/Form/UserCancelForm.php
```diff
@@ -61,45 +60,54 @@ public static function create(ContainerInterface $container) {
...
-          '#markup' => $this->t('Before removing this account, please verify and take action to modify the owner of the collections mentioned above.'),
+          '#markup' => $this->t('Before removing this account, please verify and take action to modify the owner of the groups mentioned above.'),
           '#weight' => 5,
         ];
       }
       else {
         $form['warning'] = [
-          '#markup' => $this->t('User @name cannot be deleted as it is currently the sole owner of these collections:', [
+          '#markup' => $this->t('User @name cannot be deleted as they are currently the sole owner of these groups:', [
             '@name' => $this->entity->getAccountName(),
```
claudiu-cristea: Same, I should be `->DisplayName()`.

web/modules/custom/joinup_core/src/JoinupRelationManagerInterface.php
idimopoulos: I would really prefer not to do that now. I attempted to do it but I got into a series of dependencies that I don't grasp right now (I haven't checked the other ticket in total). For example, the method is using the `::getUserMembershipsByRole` which is another method in the relation manager and then this method is simply fetching memberships by roles which brings the questions at which point should this be moved as well or the `JoinupRelationManager` should be a dependency of the `joinup_group` module and then what happens with the rest of the methods etc etc.... So, can we skip that and leave it to be dealt with entirely in the other ticket?

web/profiles/joinup/src/Form/UserCancelForm.php
idimopoulos: The code is slightly different in terms that it has different depth within the form array (the first is `$form[$bundle']` and the second one is `$form[$account->id()]['bundle']`.
Also, it would be weird to have a method trait of 15 lines have entries added to the final build array.
Can we skip this one?

web/modules/custom/joinup_core/src/JoinupRelationManagerInterface.php
idimopoulos: After looking at the rest of the comment, I have moved also the relevant method to the joinup_group service. I will move as much code as possible because it will help relax the work on the follow up ticket.

web/profiles/joinup/src/Form/UserCancelForm.php
claudiu-cristea: Yesh, I agee.

PR #1924, opened 2019-12-11
ISAICP-5729: Address newly discovered coding standards violations

pfrenssen:


idimopoulos:
I have checked the develop and indeed there are some minor remarks that are solved here.

PR #1936, opened 2019-12-19
ISAICP-5566: Sortable menu front page

idimopoulos:


web/modules/custom/joinup_search/joinup_search.module
```diff
@@ -132,13 +132,17 @@ function _joinup_search_search_api_has_condition($conditions, $field, $value, $o
...
   foreach ($documents as $document) {
     if (empty($document->getFields()['bs_sticky'])) {
       $document->setField('bs_sticky', FALSE);
     }
-    if (!isset($document->getFields()['bs_site_pinned'])) {
-      $document->setField('bs_site_pinned', FALSE);
+    if (!empty($document->getFields()['ss_search_api_id'])) {
+      /** @var \Drupal\search_api\Item\Item[] $items */
+      $id = $document->getFields()['ss_search_api_id'];
+      $document->setField('bs_site_pinned', (bool) $front_page_helper->getFrontPageMenuItem($items[$id]->getOriginalObject()->getValue()));
```
claudiu-cristea: I'm sad that we are still patching and altering rather than providing robust APIs. Providing a computed field would had been a far better solution as it would provide a reusable functionality, i.e. getting the pin weight by simply invoking `$entity->pinned` which would return the menu item weight or `NULL`, if not pinned. At least this was my understanding, after our agreement.

idimopoulos: The PR is not ready yet. I found this line which makes it too easy to patch it quickly so I made one change to quickly have a green test before moving to the computed field as we agreed.

web/modules/custom/joinup_front_page/joinup_front_page.behat.inc
```diff
@@ -0,0 +1,49 @@
...
+   *
+   * @param string $type
+   *   The entity type id.
+   * @param \Behat\Gherkin\Node\TableNode $table
+   *   A list of titles.
+   *
+   * @throws \InvalidArgumentException
+   *   Thrown when the user has entered an invalid type.
+   *
+   * @Given site pinned :type entities:
```
pfrenssen: I would not use "site pinned" as wording because this is not how it is understood by the end user. How about: 

```
@Given the following :type entities are pinned to the front page:
```

web/modules/custom/joinup_front_page/joinup_front_page.module
```diff
@@ -0,0 +1,58 @@
...
+ */
+function joinup_front_page_form_menu_edit_form_alter(&$form, FormStateInterface $form_state, $form_id): void {
+  /** @var \Drupal\system\Entity\Menu $menu */
+  $menu = $form_state->getFormObject()->getEntity();
+  if ($menu->id() !== 'front-page') {
+    return;
+  }
+
+  $form['#title'] = t('Front page pinned entities');
+  $form['links']['links']['#empty'] = t('There are no pinned entities. Start by pinning an entity site wide.');
```
pfrenssen: There are now two conflicting ways to describe the same behavior, "pin to front page" and "site wide". This is confusing. Let's completely drop all mentions of "pin site wide" and only use "pin to front page" from now on. Also don't put the word "entities" in user facing strings. This is a technical term which is meaningless to the end user. How about:

```
No content is pinned to the front page yet.
```

web/modules/custom/joinup_front_page/joinup_front_page.permissions.yml
```diff
@@ -0,0 +1,13 @@
+'feature entity':
+  title: 'Feature entity'
+  description: 'Promote an entity as featured in the website.'
+'unfeature entity':
+  title: 'Unfeature entity'
+  description: 'Removes an entity node from the featured ones in the website.'
+
+'pin entity':
+  title: 'Pin entity'
+  description: 'Promote an entity as pinned in the website.'
```
pfrenssen:
```
Pins an entity to the front page.
```

web/modules/custom/joinup_front_page/joinup_front_page.permissions.yml
```diff
@@ -0,0 +1,13 @@
...
+'unpin entity':
+  title: 'Unpin entity'
+  description: 'Removes an entity from the pinned ones in the website.'
```
pfrenssen:
```
Removes a pinned entity from the front page.
```

web/modules/custom/joinup_front_page/joinup_front_page.module
```diff
@@ -0,0 +1,58 @@
...
+ * Implements hook_form_FORM_ID_alter().
+ */
+function joinup_front_page_form_menu_edit_form_alter(&$form, FormStateInterface $form_state, $form_id): void {
+  /** @var \Drupal\system\Entity\Menu $menu */
+  $menu = $form_state->getFormObject()->getEntity();
+  if ($menu->id() !== 'front-page') {
+    return;
+  }
+
+  $form['#title'] = t('Front page pinned entities');
```
pfrenssen: Let's not mention "entities" in user facing text. How about:

```
Pinned content
```

idimopoulos: The entities I used was because we might have rdf_entities as well and `content` in the Drupal world refers to nodes :/

idimopoulos: `Pinned items`????

web/modules/custom/joinup_front_page/joinup_front_page.permissions.yml
idimopoulos: You know, even though this is a purely technical description with no impact on the user, I like it really too much for its clear message :) I am reading the current one and can only think about the frustration it might cause!

web/modules/custom/joinup_search/joinup_search.module
idimopoulos: Keep in mind that though a computed field has been created, this cannot be automated as either the computed field has to be applied to all indexable entities, or this solution has to be applied (because all other entities have to retrieve the value 0).

idimopoulos:
Further changes are approved. They do not really change the functionality. They only move code around and update strings. Tests are still green.

PR #1944, opened 2020-01-10
ISAICP-5761: Log in through EU Login in tests

pfrenssen:


tests/features/eulogin/eulogin.feature
```diff
@@ -134,13 +134,14 @@ Feature: Log in through EU Login
...
-    And I fill in "Email or username" with "jb007_local"
-    And I fill in "Password" with "123"
-    And I wait for the honeypot time limit to pass
-    When I press "Sign in"
-    Then I should see the error message "Please sign in with your EU Login account."
+    Then I should see the heading "Sign in to continue"
+    # Also the "user" page which in a normal Drupal installation redirects to
+    # the Drupal login form should now redirect to EU Login.
+    When I visit "user"
+    Then I should see the heading "Sign in to continue"
```
claudiu-cristea: Is this assertion correct? We've been decided that there's no redirect but a page w/o the form and still wit the warning banner.

tests/features/eulogin/eulogin.feature
```diff
@@ -327,11 +328,6 @@ Feature: Log in through EU Login
     Then I should see the error message "First name field is required."
     And I should see the error message "Family name field is required."
 
-  Scenario: The Drupal login form shows a warning message.
-    When I visit "/user/login"
-    Then I should see the warning message "As of 01/02/2020, EU Login will be the only authentication method available on Joinup. So, we strongly recommend you to choose EU Login as your preferred sign-in method!"
-    And I should see the link "EU Login"
-
```
claudiu-cristea: This should not be deleted but replaced with the new warnung text, provided by Filip.

web/modules/custom/eulogin/src/Routing/JoinupEuLoginRouteSubscriber.php
```diff
@@ -29,6 +29,12 @@ protected function alterRoutes(RouteCollection $collection): void {
         ->setDefaults(['_controller' => UserRegisterRedirectController::class . '::redirectUserRegister'])
         ->setRequirements(['_user_is_logged_in' => 'FALSE']);
     }
+
+    // Always deny access to '/user/login'. Users are expected to log in through
+    // EU Login.
+    if ($route = $collection->get('user.login')) {
+      $route->setRequirement('_access', 'FALSE');
+    }
```
claudiu-cristea: This is not correct. It was decided in the standup that we don't deny the access to the route but we disable (or hide) the form.

tests/features/eulogin/eulogin.feature
```diff
@@ -327,11 +328,6 @@ Feature: Log in through EU Login
     Then I should see the error message "First name field is required."
     And I should see the error message "Family name field is required."
 
-  Scenario: The Drupal login form shows a warning message.
-    When I visit "/user/login"
-    Then I should see the warning message "As of 01/02/2020, EU Login will be the only authentication method available on Joinup. So, we strongly recommend you to choose EU Login as your preferred sign-in method!"
-    And I should see the link "EU Login"
```
claudiu-cristea: This shouldn't be remover. We keep this route/form. In ISAICP-5760, we only hide/disable the form and replace the text with one provided by Filip

tests/features/legal/legal_notice.feature
```diff
@@ -21,19 +21,29 @@ Feature:
...
+    And I press "Log in"
 
-    When I press "Sign in"
-    Then I should see the warning message "You must accept this agreement before continuing."
+    And I select the radio button "I am a new user (create a new account)"
+    Then I should see "I have read and accept the Legal notice"
+
+    # Submit without accepting the 'Legal notice'.
+    When I press "Next"
+    Then I should see the error message "You must accept the Legal notice in order to use our platform."
```
claudiu-cristea: Out-of-scope. In this stage the normal login is still available. It will be remove later, in ISAICP-5760.

tests/features/legal/legal_notice.feature
```diff
@@ -157,6 +167,9 @@ Feature:
     When I am on "/contact"
     Then I should not see "I have read and accept the Legal notice"
 
+  # Todo: in ISAICP-5760 this scenario can be removed, since this is already
+  # described above in "User login when a new 'Legal notice' version is
+  # released".
```
claudiu-cristea: See previous comment.

tests/features/user/login.feature
```diff
@@ -4,6 +4,8 @@ Feature: User login
   As a user of the website
   I need to be able to login with my credentials.
 
+  # Todo: needs to be adapted to use EU Login in ISAICP-5760.
+  @wip
```
claudiu-cristea: Why adding these TODOs? We know that, in ISAICP-5760, we'll have to adapt them. Otherwise, in ISAICP-5760, the tests will fail.

tests/features/user/password.feature
```diff
@@ -1,4 +1,6 @@
-@api
+@api @wip
+# Todo: this should be adapted when moving to EU Login in ISAICP-5760. Possibly
+# this needs to be removed entirely.
```
claudiu-cristea: Same here.

tests/features/user/register.feature
```diff
@@ -3,12 +3,6 @@ Feature: User registration
   As a user I must be able to register to the site and complete my user profile
   and receive appropriate notifications.
 
-  Scenario: User can find the register page
-    Given I am an anonymous user
-    When I am on the homepage
-    And I click "Sign in (legacy)"
-    Then I should see the heading "Sign in"
-
```
claudiu-cristea: Not true. This will still work as, we've decided in standup.

tests/features/user/register.feature
```diff
@@ -35,7 +29,9 @@ Feature: User registration
       | body      | The Joinup Support Team created your account on Joinup. Please log in through the following link in order to set your password. |
     And the account for SuperUser should be active
 
-  @email
+  # This test should be converted for EU Login but seems incomplete and is
+  # probably no longer applicable and can be removed in ISAICP-5760.
+  @email @wip
```
claudiu-cristea: Why this TODO as we have ISAICP-5760?

tests/features/eulogin/eulogin.feature
pfrenssen: This is temporary and will change in ISAICP-5760 which is going to enable the user login again and show the warning as written in the ticket description. The user login has been completely disabled in this PR to be 100% sure that no tests can sneakily log in a user through the old form.

I will add a `Todo` here so that it will not be forgotten.

web/modules/custom/eulogin/src/Routing/JoinupEuLoginRouteSubscriber.php
pfrenssen: This is temporary and is going to be removed in scope of ISAICP-5760. I will add a `@todo` here.

tests/features/eulogin/eulogin.feature
pfrenssen: This comment is duplicated from above. Possibly a bug in Github?

tests/features/legal/legal_notice.feature
pfrenssen: Yeah I already started this conversion when I found out that this is in fact in scope of ISAICP-5760. I left it here to reduce the work in ISAICP-5760 which is marked as a followup of this ticket.

tests/features/user/login.feature
pfrenssen: I added them because it took a lot of time to go over all these scenarios and identify which ones needed to be updated, after multiple full test runs. It will make it much easier to do the first pass of ISAICP-5760, since most things will already be fixed before running the first test. Also it will help to not miss any of these tests which are marked with `@wip`, since they won't run and won't fail.

tests/features/user/register.feature
pfrenssen: Will add it back and add a `@wip` tag and a `Todo` with a mention of ISAICP-5761. I don't even remember this being discussed in the standup, but it makes a lot of sense. It would be better to write down these decisions in the ticket instead of relying verbal agreements and people's bad memory ;)

tests/features/user/register.feature
pfrenssen: Because of the `@wip`, when the test is disabled it will not fail.

PR #1956, opened 2020-01-21
Isaicp 5759

pfrenssen:


tests/features/eulogin/eulogin.feature
```diff
@@ -217,6 +206,50 @@ Feature: Log in through EU Login
...
+    Given CAS users:
+      | Username   | E-mail              | Password | First name  | Last name |
+      | jeanclaude | muscles@brussels.be | dragon12 | Jean-Claude | Van Damme |
+    And users:
+      | Username | Password | E-mail         | First name | Family name |
+      | jclocal  | dragonne | jcvd@gmail.com | JC         | VD          |
+
+    Given I am on the homepage
+    And I click "Sign in"
+    And I click "EU Login"
```
idimopoulos: This selects the link with a partial check, meaning that either "Sign in with EU Login" or "Sign in with EU Login (recommended)" is accepted.
I do accept that since we are dropping the parenthesis according to the ticket, there is no point to check for text _not_ being shown. However, the `(recommended) ` partial text is still appearing on the link.

idimopoulos: As discussed with @pfrenssen, this is left for a different implementation since it was based in the idea that the legacy login would be removed but is still there with a warning.

PR #1958, opened 2020-01-22
ISAICP-5740: Debug functionality for Joinup.

idimopoulos:


web/modules/custom/joinup_debug/joinup_debug.module
```diff
@@ -0,0 +1,71 @@
...
+}
+
+/**
+ * Logs the wrong revision error to the logger as a critical issue.
+ *
+ * @param \Drupal\node\NodeInterface $node
+ *   The related node.
+ */
+function _joinup_debug_log_wrong_revision_error(NodeInterface $node): void {
+  $backtrace = debug_backtrace();
```
pfrenssen: This has the potential to exceed the memory limits on production depending on where in the stack this code is called. Since we will deploy this on production, let's stay on the safe side and ignore the arguments, call object, and limit the length of the backtrace:

```
$backtrace = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 50);
```

web/modules/custom/joinup_debug/tests/src/Functional/DbLogTest.php
```diff
@@ -0,0 +1,94 @@
...
+      'name' => 'Basic page',
+      'display_submitted' => FALSE,
+    ]);
+  }
+
+  /**
+   * Tests the wrong revision update.
+   */
+  public function testDbLogRevisionUpdate() {
+    /** @var \Drupal\node\NodeStorageInterface $node_storage */
```
pfrenssen: This is not testing the exact format of the watchdog messages, only the fact that they are being generated. I agree with this approach. I have validated manually that the message format is correct. We don't need in depth coverage since this is a temporary measure with no end user facing consequences if this regresses.

web/modules/custom/joinup_debug/joinup_debug.info.yml
```diff
@@ -0,0 +1,5 @@
+name: 'Joinup debug'
+type: module
+description: 'Debug module for the Joinup platform.'
```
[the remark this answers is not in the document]
idimopoulos: I was thinking that this `joinup_debug` module could be used for possible future issues as well, but since this is a very rare case, I will change it accordingly.

PR #1962, opened 2020-01-23
5506: Replace passwords for EU Login users

claudiu-cristea:


web/modules/custom/eulogin/src/Event/Subscriber/JoinupEuLoginCasAccountLinkEventsSubscriber.php
```diff
@@ -53,11 +56,31 @@ public function setEmailCollisionMessage(CasAccountLinkEmailCollisionEvent $even
...
+  /**
+   * Sets a random password on the account..
+   *
+   * @param \Drupal\cas_account_link\Event\Events\CasAccountLinkPostLinkEvent $event
+   *   The CAS Account Link post-linking event object.
+   *
+   * @throws \Drupal\Core\Entity\EntityStorageException
+   *   Thrown if the bundle does not exist or was needed but not specified.
+   */
+  public function setFakePassword(CasAccountLinkPostLinkEvent $event): void {
```
pfrenssen: Just a small suggestion, but wouldn't it be better to call it `setRandomPassword()`? That would better explain what this method does. Also the password is not really fake, it is still a real password, and even a very good one with 30 characters ;)

pfrenssen:
Looking good, tested manually against the production database with sample sets of 5 linked user accounts, 5 unlinked user accounts, and 5 removed user accounts. I had 1 small remark but I fixed it myself.

PR #1964, opened 2020-01-27
5519: Masquerade functionality

claudiu-cristea:


tests/features/user/masquerade.feature
```diff
@@ -0,0 +1,102 @@
...
+      | target_account_2     |            |             |
+    And I am logged in as a moderator
+
+    When I visit the profile of i_want_to_masquerade
+    And I click Edit
+    Then the following fields should be present "Masquerade as"
+
+    When I fill in "masquerade_as[0][target_id]" with "target_account_1"
+    And I press "Save"
+    And I click Edit
```
pfrenssen: I found a vulnerability here, a moderator can edit their own account and enter `joinup-admin` as a masquerade target and elevate their permission level this way. Normally the admin user is blocked on production but there are times when they are temporarily enabled. Also a user could possibly masquerade as a moderator, which would also exceed their permission level.

I think we should implement `hook_masquerade_access()` and limit the way people can exceed their own permission levels. The simplest way would be to only masquerade as users that are authenticated but have no other roles.

edit: This was discussed during standup, and it should be possible to masquerade as moderators, but not as administrator or super user.

claudiu-cristea: Great catch! As we've been discussed in the today's standup, this will be a follow-up in the _Masquerade_ epic. I've been opened ISAICP-5812.

PR #2000, opened 2020-02-17
Isaicp-5787: Enable a paragraphs demo.

idimopoulos:


pfrenssen:
On clean install I get:

> Unable to install the <em class="placeholder">Layout Builder</em> module since it requires the <em class="placeholder">Field UI</em> module.

PR #2005, opened 2020-02-20
Update Gulp packages.

papegaill:


web/themes/joinup/prototype/gulpfile.js
```diff
@@ -4,7 +4,7 @@ var gulp = require('gulp'),
     concat = require('gulp-concat'),
     watch = require('gulp-watch'),
     sass = require('gulp-sass'),
-    styleguide = require('sc5-styleguide'),
+    // styleguide = require('sc5-styleguide'),
```
claudiu-cristea: Do we still need this commented out?

papegaill: no this will be updated, we had confirmation that we could remove styleguide task.

PR #2014, opened 2020-02-25
ISAICP-5829: Introduce select2 for the solution type field.

idimopoulos:


idimopoulos:
@claudiu-cristea please give a bit of attention to the commits https://github.com/ec-europa/joinup-dev/pull/2014/commits/f366660ecafdcbef1b851694eb387df05c150c0a and https://github.com/ec-europa/joinup-dev/pull/2014/commits/44bec62b4f203a6ac55b2ec890ed05589aaaee9e as I have a small dilemma. I left the second one to be the one I find more appropriate. There is a small issue related to the select methods you overrode. 
Normally, for the default option select field, there are two methods, the `::selectOption` and the `::additionallySelectOption`. The first one selects anew an option in the select field which means that even if the field is multiselect, all other fields are deselected first.
The second respects the multi select and adds an additional option. Under the same principles, when "selecting" and not "additionally selecting" an option, the other options should be removed first. However, one could argue that the "additional selection kinda mimics the CTRL holding in the element in order to additionally add the extra option.
Personally, I prefer to have the select method deselect all current options first.

idimopoulos:
Also, the same is being done in the normal select where it ends up in the `\Behat\Mink\Driver\Selenium2Driver::selectOptionOnElement` (if use the selenium driver for example) where the current options are deselected if the multiple flag is false.

tests/src/Context/MinkContext.php
```diff
@@ -86,4 +89,108 @@ public function iWaitForAjaxToFinish($event = NULL) {
...
+   *
+   * @param string $option
+   *   The option to be deselected.
+   * @param string $select
+   *   The select field name.
+   *
+   * @throws \Exception
+   *   Thrown if the field is not a select2 widget.
+   *
+   * @Given I deselect the option :option from the :select select2 widget
```
claudiu-cristea: I cannot find any usage of this step definition but let's keep it, it might be useful in the future.

Again s/select2/Select2 :)

tests/src/Context/MinkContext.php
```diff
@@ -86,4 +89,108 @@ public function iWaitForAjaxToFinish($event = NULL) {
...
+  /**
+   * Deselects a Select2 option by clicking the x button in the chip.
+   *
+   * @param string $option
+   *   The option to be deselected.
+   * @param string $select
+   *   The select field name.
+   *
+   * @throws \Exception
+   *   Thrown if the field is not a select2 widget.
```
claudiu-cristea: s/select2/Select2

tests/src/Context/MinkContext.php
```diff
@@ -86,4 +89,108 @@ public function iWaitForAjaxToFinish($event = NULL) {
...
+   *
+   * @Given I deselect the option :option from the :select select2 widget
+   */
+  public function deselectSelect2Option(string $option, string $select): void {
+    if (!$field = $this->select2IsUsed($select, $option)) {
+      throw new \Exception('This method can only be used for a select2 widget.');
```
claudiu-cristea: s/select2/Select2

tests/src/Context/MinkContext.php
```diff
@@ -86,4 +89,108 @@ public function iWaitForAjaxToFinish($event = NULL) {
...
+  /**
+   * Checks if a given select field is using Select2.
+   *
+   * @param string $select
+   *   The select.
+   * @param string $option
+   *   The option to be selected.
+   *
+   * @return \Behat\Mink\Element\NodeElement|false
+   *   It returns the field as node element, if Select2 is used it or FALSE
```
claudiu-cristea: Typo. My mistake: "...if Select2 is used it or FALSE" should be "...if Select2 is use or FALSE"

tests/src/Context/MinkContext.php
```diff
@@ -86,4 +89,108 @@ public function iWaitForAjaxToFinish($event = NULL) {
...
+  public function selectOption($select, $option): void {
+    if ($field = $this->select2IsUsed($select, $option)) {
+      $xpath = '//li[contains(@class, "select2-selection__choice")]/span[contains(@class, "select2-selection__choice__remove")]';
+      if ($selected_remove_buttons = $field->getParent()->findAll('xpath', $xpath)) {
+        // This method selects anew an option in a normal select list. To have
+        // the same results in a select2 widget, all selected options have to
+        // be removed manually.
+        // To select an additional item, use the ::additionallySelectOption
+        // method.
+        foreach ($selected_remove_buttons as $selected_remove_button) {
```
claudiu-cristea: Great catch! I fully agree with your point, the 1st method, `::selectOption()`, is meant to do a single selection, like doing a simple click on a multi-value select. That meaning all options previously selected are deselected. While `::additionallySelectOption()` would append new selections to the already selected options, simulating holding the CTRL key.

pfrenssen:
This has been replaced by a custom approach after functional review, see new proposal in ISAICP-5757. This approach has been abandoned.

PR #2017, opened 2020-02-27
ISAICP-5650: Solution RSS feed.

idimopoulos:


tests/src/Context/RssContext.php
```diff
@@ -121,7 +121,8 @@ public function assertRssItems(TableNode $table): void {
         $nodes = $xpath->query("//channel/item[$xpath_node_index]/$key");
         // Support is limited to nodes that appear once only.
         Assert::assertCount(1, $nodes, "Invalid count for $key element.");
-        Assert::assertEquals($value, trim($nodes->item(0)->nodeValue));
+        $actual_value = trim($nodes->item(0)->nodeValue);
+        Assert::assertEquals($value, $actual_value, "'{$value}' is equal to '{$actual_value}'.");
```
claudiu-cristea:
```suggestion
        Assert::assertSame($value, $actual_value);
```

When asserting that 2 strings are equal is recommended to use `::assertSame()`. The assertion message is useless and also wrong. In case of failure there's also a similar standard message displayed. In PHPUnit the assertion message should describe the failure not the success.

idimopoulos: Ok, we can revert!

PR #2033, opened 2020-03-12
ISAICP-5882: Fatal error when saving a document with an orphaned file reference

pfrenssen:


web/modules/custom/joinup_licence/joinup_licence.module
```diff
@@ -184,6 +184,9 @@ function joinup_licence_theme($existing, $type, $theme, $path) {
  *
  * Adds extension and file size to the available variables when rendering a
  * document node.
+ *
+ * @todo Why is this in joinup_licence? What does this have to do with licences?
+ * @todo Update documentation, this is not retrieving extensions nor file sizes.
```
claudiu-cristea: _A `@todo` without a link to a follow-up it's like a scream in space_, somebody from the community once said :) Could you please open a follow-up and link it here?

pfrenssen: Good point, I was in a rush to get this out to unblock Manos. Now that things are less hectic I can even just fix the problem instead of delegating it to a followup.

PR #2038, opened 2020-03-14
ISAICP-5844: Prohibit linking UID1 with a CAS account. Don't limit its access

claudiu-cristea:
**Note:** This is blocked & branched on top of #1989


web/modules/custom/eulogin/tests/src/ExistingSite/JoinupEuLoginTest.php
```diff
@@ -128,16 +130,31 @@ public function testLimitedAccess(): void {
...
+    $this->account->save();
+
+    $this->drupalLogin($this->account);
+
+    // Try to navigate to pages that are not accessible to regular users.
+    $this->assertAccess('<front>');
+    $this->assertAccess('/collections');
+    $this->assertAccess('/solutions');
+    $this->assertAccess('/keep-up-to-date');
+    $this->assertAccess('/search');
```
pfrenssen: It is not clear why this is being tested, the documentation says that these pages are not accessible to regular users, but they are, and this is already tested above. Was it the idea to have pages here such as the moderator only pages? But the permission doesn't grant access to anything.

web/modules/custom/eulogin/joinup_eulogin.permissions.yml
```diff
@@ -0,0 +1,4 @@
+unlimited access:
+  title: 'Unlimited access'
+  description: 'Allows unlimited access for user not linked to an EU Login account.'
+  restrict access: true
```
pfrenssen: This permission does not have a good name. It implies that the users with this permission are given unlimited access to ... something? But in fact the only thing happening is that their access to the EU account linking is being taken away from them. This is quite strange, that granting this permission is actually reducing the functionality for the user.

Wouldn't it be better to invert this, and call it something like `link account with eulogin`?

web/modules/custom/eulogin/src/Event/Subscriber/JoinupEuLoginKernelSubscriber.php
```diff
@@ -206,7 +209,9 @@ protected function userHasUnlimitedAccess(): bool {
     // - is anonymous.
     return $this->currentUser->isAnonymous()
     // - or is an EU Login linked user.
-    || $this->authmap->get($this->currentUser->id(), 'cas');
+    || $this->authmap->get($this->currentUser->id(), 'cas')
+    // - or is granted with 'unlimited access' permission.
+    || $this->currentUser->hasPermission('unlimited access');
```
pfrenssen: We are checking a role, a state and a permission. We can already avoid one by giving the permission to the anonymous user now.

Also this could use some more explanation, for me as a reviewer not familiar with this code base I am unsure what this means that the user has "unlimited access". It sounds very drastic, as if they get access rights similar to UID1, but it is probably just a naming issue.

web/modules/custom/eulogin/tests/src/ExistingSite/JoinupEuLoginTest.php
pfrenssen: This was clarified in a call with @claudiu-cristea, we will update the documentation with additional information.

web/modules/custom/eulogin/src/Event/Subscriber/JoinupEuLoginKernelSubscriber.php
pfrenssen: I reconsider the suggestion to assign this permission to anonymous users. I think it is better to check it explicitly like is done now.

PR #2040, opened 2020-03-16
ISAICP-5885: Bring phpunit.xml.dist up to date with the latest change…

pfrenssen:
…s in core.

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -11,15 +11,8 @@
          beStrictAboutOutputDuringTests="true"
          beStrictAboutChangesToGlobalState="true"
          checkForUnintentionallyCoveredCode="false"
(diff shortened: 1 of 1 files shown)


phpunit.xml.dist
```diff
@@ -11,15 +11,8 @@
          beStrictAboutOutputDuringTests="true"
          beStrictAboutChangesToGlobalState="true"
          checkForUnintentionallyCoveredCode="false"
+         printerClass="\Drupal\Tests\Listeners\HtmlOutputPrinter"
```
claudiu-cristea: I love this! Everytime I'm adding this manually :)

pfrenssen: Yeah it turns out this bug in PHPStorm that was originally causing it to be commented out was already fixed years ago, but the issue was never set to "fixed" :grin:

PR #2042, opened 2020-03-16
The entity manager has been deprecated and will be removed in Drupal 9.

pfrenssen:


pfrenssen:
This is failing because of a bug in drupalextension which is fixed in https://github.com/jhedstrom/drupalextension/pull/561

PR #2048, opened 2020-03-19
ISAICP-5692: Move JoinupRelationManager out of joinup_core

pfrenssen:


web/modules/custom/joinup_group/src/JoinupGroupRelationInfoInterface.php
```diff
@@ -2,17 +2,17 @@
...
 
 /**
- * An interface for Joinup relation manager services.
+ * An interface for services that provide information about group relations.
  */
-interface JoinupRelationManagerInterface {
+interface JoinupGroupRelationInfoInterface {
 
   /**
    * Retrieves the parent of the entity.
```
claudiu-cristea: This method is useless and should be removed. Instead of:

```php
/** @var \Drupal\joinup_group\JoinupGroupRelationInfoInterface $relation_info */
$relation_info = \Drupal::service('joinup_group.relation_info');
$parent = $parent = $relation_info->getParent($node);
```

we should use:

```php
$parent = $node->get('og_audience')->entity;
```

or:

```php
$parent = $node->get(OgGroupAudienceHelperInterface::DEFAULT_FIELD)->entity;
```

web/modules/custom/joinup_group/src/JoinupGroupRelationInfoInterface.php
```diff
@@ -34,8 +34,8 @@ public function getParent(EntityInterface $entity): ?RdfInterface;
    *
    * @return int
    *   The moderation status. Can be one of the following values:
-   *   - NodeWorkflowAccessControlHandler::PRE_MODERATION
-   *   - NodeWorkflowAccessControlHandler::POST_MODERATION
+   *   - CommunityContentWorkflowAccessControlHandler::PRE_MODERATION
+   *   - CommunityContentWorkflowAccessControlHandler::POST_MODERATION
    */
   public function getParentModeration(EntityInterface $entity): ?int;
```
claudiu-cristea: This method is used only once, in `joinup_community_content.module`. It seems to me that belongs more to `joinup_community_content` module. As is used only once, I don't even see why we need a service method?

```php
$fields = ['collection' => 'field_ar_state', 'solution' => 'field_is_state'];
$parent = $node->get('og_audience')->entity;
$moderation = $node->get($fields[$patent->bundle()])->value;
```

The same is true for `::getParentState()` and `::getParentContentCreationOption()`. All three are used **only in the community content module** and **only once**. I would remove them but if we plan to keep them as service methods let's move them there in `CommunityContentHelper` or in a new service. And, as I see, there's a high level of code reusability between the three methods. Maybe we can optimize a little bit.

Others (regarding the `JoinupGroupRelationInfo` service methods):

1. `::getGroupUsers()` not used anywhere. Should be removed.
1. `::getGroupMemberships()` used only in `::getGroupUsers()` but that is used nowhere :)
1. `::getUserGroupMembershipsByBundle()` not used anywhere. BTW, there's an identical `JoinupGroupManagerInterface::getUserGroupMembershipsByBundle()`. I wonder why is this duplicated? Should be removed.
1. `::getCollectionIds()` used only in one place but it seems to me that creating a service only to get all entities of a given bundle should not be in the scope of a service. It's just `\Drupal::entityQuery('rdf_entity')->condition('rid', 'collection')->execute()`. I would remove it.
1. Same for `::getSolutionIds()`, doesn't smell as something that should be a service. I would remove them both. 
1. `::getRdfEntityIdsByBundle()` just a helper for `::getCollectionIds()` and `::getSolutionIds()`. If we remove those, then this should go.
1. `::getOgMembershipStorage()` used only in `::getUserGroupMembershipsByBundle()` but that should be dropped (see above).
1. `::getContactInformationRelatedGroups()` belongs to `contact_information` module. Otherwise this adds `contact_information` dependency to `joinup_group`. This dependency exists now even is not listed `contact_information.info.yml`. Probably is fulfilled by other dependencies.


As a conclusion, I think we should drop this service instead of moving it around.

pfrenssen: OK I like this suggestion a lot. Removing it is even better than relocating it. I will have a go at this.

web/modules/custom/joinup_group/src/JoinupGroupRelationInfoInterface.php
pfrenssen: I looked into this and it seems this method is intended to be used in places where we deal with any type of node or even other entities. I see that in most cases where multiple entity types are supported it is first filtering on nodes though for efficiency.

For example if any code that deals with community content and other node types cannot use this because only group related content has the `og_audience` field. We cannot guarantee that every node type we will ever support in the future will have this field, or even that it will have this machine name.

For the code that deals exclusively with community content we can remove it though.

claudiu-cristea: > For example if any code that deals with community content and other node types cannot use this because only group related content has the `og_audience` field. We cannot guarantee that every node type we will ever support in the future will have this field, or even that it will have this machine name.

Yes, but we have no use cases. So, if we remove the actual usages, it's dead code. I wonder if we should build/keep an unused method just for a hypothetical, unlikely future usage.

pfrenssen: Hmmm looking deeper it seems that if this method is called on non-group entities then the current implementation will throw an exception inside `MembershipManager::getGroupIds()` so my assumption is wrong, this method _cannot_ be called on non-group content.

pfrenssen:
I am working on removing or replacing all the methods on `JoinupGroupRelationInterface` as suggested by @claudiu-cristea:

- [ ] getCollectionIds()
- [ ] getContactInformationRelatedGroups()
- [ ] getGroupMemberships()
- [ ] getGroupUsers()
- [ ] getParent()
- [ ] getParentContentCreationOption()
- [ ] getParentModeration()
- [ ] getParentState()
- [ ] getSolutionIds()
- [ ] getUserGroupMembershipsByBundle()

PR #2054, opened 2020-03-23
ISAICP-5712: Search result count for webtools analytics.

idimopoulos:


web/modules/custom/isa2_analytics/src/EventSubscriber/WebtoolsAnalyticsSubscriber.php
```diff
@@ -122,6 +122,9 @@ public function setSearchData(AnalyticsEventInterface $event) {
     if ($keys = $this->currentRequest->get('keys')) {
       $search_data = $event->getSearch();
       $search_data->setKeyword($keys);
+      if ($total_rows = $this->currentRequest->attributes->get('total_rows')) {
```
claudiu-cristea: Quoting from https://webgate.ec.europa.eu/fpfis/wikis/display/webtools/Europa+Analytics:

> Remark:
> 
> There is a difference between
> * not knowing the count of search results : the parameter count shouldn't be used
> * knowing that the count of search results is zero: the parameter count should be used with the value 0.

It's not very clear for me in what consist this difference. For this reason we need a real test with the EC Piwik server to see the results. I have no idea on how we can test this.

PR #2065, opened 2020-03-29
ISAICP-5413: Basic metadata for RDF entities.

idimopoulos:


web/modules/custom/joinup_seo/joinup_seo.module
```diff
@@ -62,3 +62,52 @@ function joinup_seo_metatags_alter(array &$metatags, array &$context) {
...
+
+  /** @var \Drupal\sparql_entity_storage\SparqlSerializerInterface $serializer */
+  $serializer = \Drupal::service('sparql_entity.serializer');
+  if (!$output = $serializer->serializeEntity($entity, 'jsonld')) {
+    return;
+  }
+
+  // Serializer returns an array so we remove the outer array enclosure tags
+  // to output the only element standalone. Works like reset().
+  $output = trim($output, '[]');
```
claudiu-cristea: Removing the outer enclosure is totally wrong. The result is an invalid JSON. I've tested the page http://joinup.test/collection/better-legislation-smoother-implementation after rebuilding.  You can see the results here https://jsonformatter.org/bdf1b8 (press Validate). So, I doubt that Google will parse this info if it cannot validate is as valid JSON.

I'm also questioning the output of some fields, such as:

```json
  "http://joinup.eu/search": [
    {
      "@value": "a:4:{s:6:\"fields\";a:3:{s:21:\"collection_event_type\";a:2:{s:6:\"weight\";i:0;s:6:\"region\";s:13:\"inline_facets\";}s:24:\"collection_policy_domain\";a:2:{s:6:\"weight\";i:0;s:6:\"region\";s:13:\"inline_facets\";}s:15:\"collection_type\";a:2:{s:6:\"weight\";i:0;s:6:\"region\";s:3:\"top\";}}s:7:\"enabled\";i:1;s:13:\"query_presets\";s:0:\"\";s:5:\"limit\";s:2:\"12\";}"
    }
  ],
```

Not that it's useless, but it might influence the Google results in a way that we don't want.

Other aspect is the list of references from the bottom:

```json
...
{
  "@id": "http://data.europa.eu/w21/d4932b01-c431-43af-95ee-9f2f68fd4668"
},
{
  "@id": "http://data.europa.eu/w21/f874db59-a770-4ae7-835a-a38f628e1a04"
},
{
  "@id": "http://joinup.eu/ontology/policy-domain#ict"
},
{
  "@id": "http://joinup.eu/ontology/policy-domain#ict-standardisation"
},
{
  "@id": "http://joinup.eu/ontology/policy-domain#ict-standardisation-in-egovernment"
},
{
  "@id": "http://publications.europa.eu/resource/authority/country/EUR"
},
{
  "@id": "http://www.w3.org/ns/dcat#Catalog"
}
```

What's the point of having these?

I would consider rescoping this ticket and take into consideration to output only ADMS fields and inject a compact, really useful, metadata blob in the document.

idimopoulos: I also think that we should not have these but we discussed this in the technical meeting and we said that we agreed to output them since we already do on export.

The last ids are entities that refer the current entity so it is kinda of entities that have some relation to the exported one.

As for the validation of the ADMS, I tested the data with live data in https://search.google.com/structured-data/testing-tool/u/0/ and it read them without issue and even pointed out all properties.
As for having fields polluting what google reads, google has some predefined data handling, like news and events, but when you provide a custom ontology, it will simply read them. Generally, I do not like the format that much but we discussed this and we agreed. That's why I kept it.

web/modules/custom/joinup_seo/joinup_seo.module
```diff
@@ -62,3 +62,52 @@ function joinup_seo_metatags_alter(array &$metatags, array &$context) {
...
+    'asset_release',
+    'asset_distribution',
+    'licence',
+  ];
+  // Owners and contact information entities are not viewable by anonymous users
+  // in their canonical route so there is no point attaching metadata to these
+  // pages.
+  if (!in_array($entity->bundle(), $allowed_bundles)) {
+    return;
+  }
```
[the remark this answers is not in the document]
idimopoulos: Sure, fine by me :) Removing the restriction.

web/modules/custom/joinup_seo/joinup_seo.module
idimopoulos: If you remember, I argued over outputting only ADMS fields but we went with printing out everything :/

claudiu-cristea: OK, here's my proposed approach:

# Decode `$output` and keep the first array item by performing a real `reset()` on a array. Chopping the JSON blob is not a nice.
# Encode the result.

In this way we get rid of the trailing references that don't make any sense to metadata and the resulting JSON is valid.

Consider a follow-up ticket to keep only relevant data.

PR #2068, opened 2020-03-30
ISAICP-5872: Provide a fix for the accounts with empty email.

idimopoulos:


web/modules/custom/joinup_core/joinup_core.post_update.php
```diff
@@ -1134,3 +1134,29 @@ function joinup_core_post_update_delete_orphaned_solutions() {
     $solution->delete();
   }
 }
+
+/**
+ * Implements hook_post_update_NAME().
```
claudiu-cristea: On (post)updates we write here a description of the operation we're performing. The first word should be an imperative verb (no "s" at the end of the word). See https://www.drupal.org/docs/develop/standards/api-documentation-and-comment-standards#updateN

web/modules/custom/joinup_core/joinup_core.post_update.php
```diff
@@ -1134,3 +1134,29 @@ function joinup_core_post_update_delete_orphaned_solutions() {
...
+    }
+  }
+
+  $user_storage = \Drupal::entityTypeManager()->getStorage('user');
+  $uids = $user_storage
+    ->getQuery()
+    ->notExists('mail')
+    ->execute();
+  // Remove the anonymous user.
+  unset($uids['0']);
```
claudiu-cristea: Could we add this as query condition?

web/modules/custom/joinup_core/joinup_core.post_update.php
```diff
@@ -1134,3 +1134,29 @@ function joinup_core_post_update_delete_orphaned_solutions() {
     $solution->delete();
   }
 }
+
+/**
+ * Fix broken accounts with empty emails.
+ */
+function joinup_core_post_update_fix_broken_accounts(&$sandbox) {
```
claudiu-cristea: * This belongs more to `joinup_user`.
* Parameter `$sandbox` is not used.
* Missing return type.

PR #2078, opened 2020-04-03
ISAICP-5568: Improve user profile layout v2.

papegaill:


web/themes/joinup/templates/user/user--group-header.html.twig
```diff
@@ -26,49 +26,69 @@
...
+                        <div>ICT</div>
+                        <div>Science and technology</div>
+                        <div>Collaboration</div>
+                        <div>Free and open source software in Public Administrations</div>
+                        <div>ICT standardisation in eGovernment</div>
+                        <div>Open government</div>
+                        <div>eGovernmentttt</div>
+                    </div>
+                  </span>
+                </div>
```
pfrenssen, line 86: This test data should be removed.

web/themes/joinup/templates/user/user--group-header.html.twig
```diff
@@ -26,49 +26,69 @@
...
+            <div class="mdl-grid mdl-grid--no-spacing">
+
+              <div class="mdl-cell mdl-cell--6-col mdl-cell--4-col-wide">
+                {% if content.field_user_organisation|render %}
+                  <div class="stats__wrapper stats__wrapper-organisation">
+                    <span class="stats__text bold">{{ 'Organisation'|t }}:</span>
+                    <span class="stats__text">{{ content.field_user_organisation }}</span>
+                  </div>
+                {% endif %}
+              </div>
```
pfrenssen, line 63: If the user doesn't enter an organisation then there is a strange void left in place here. We should probably not render the empty div in this case.

2020-04-06-115304_7680x2160_scrot

pfrenssen:
Looking good overall, found two small issues. We are also removing the country of the user but according to functional requirements this should be visible. Let's wait for an answer from the functional team on the JIRA issue before continuing with this.

web/themes/joinup/templates/user/user--group-header.html.twig
papegaill: Oups, updated ;)

web/themes/joinup/templates/user/user--group-header.html.twig
papegaill: Thank you for this feedback,PR updated accordingly.

papegaill:
Hi @pfrenssen thank you for your review, I updated the PR based on your remarks and I also added 'members of' and 'country ' fields as requested.

papegaill:
Hi @claudiu-cristea many thanks for review and feedback, I updated the PR accordingly.

claudiu-cristea:
@papegaill, excellent, thank you. In fact the failed test, showed that the label has been showed without having a field value. I've wrongly changed the test. Now I reverted the test change but pushed with `--force`, so you'll delete the local branch and checkout again from remote.

We're almost there. However...

> (...) Also the country value should be inline with the label, as this is a single value field

...this was not implemented.

papegaill:
@claudiu-cristea Thank you, Added css fixe to have country value inline with the label.

PR #2079, opened 2020-04-03
ISAICP-5258: Unlock Facets

pfrenssen:


web/profiles/joinup/src/Plugin/ArbitraryFacet/EventsArbitraryFacet.php
```diff
@@ -103,4 +103,11 @@ public function getFacetDefinition(): array {
     return $definition;
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function getCacheContexts() {
```
claudiu-cristea, line 110: We should use `\Drupal\Core\Cache\Cache::mergeContexts()` here.

pfrenssen:
Thanks!

PR #2081, opened 2020-04-05
ISAICP-5794: Configure site URL for Drush on Joinup environments

claudiu-cristea:


scripts/rpmbuild/make_joinup_build_uat.sh
```diff
@@ -35,10 +35,11 @@ mkdir -p ${JOINUP_DIR} || exit 1
 sudo ${COMPOSER_PATH} self-update || exit 1
 COMPOSER_DISCARD_CHANGES=true ${COMPOSER_PATH} install --no-interaction --prefer-dist || exit 1
 ./vendor/bin/phing compile-scss || exit 1
+./vendor/bin/phing setup-drush || exit 1
```
pfrenssen, line 38: Out of scope for this ticket, but these three steps are essentially the same as `phing build-dist`, which is used in the build scripts for the other environments. It is not done here because we want the dev dependencies. I see also that in the other scripts the composer installation is executed twice. We can consolidate and optimize this. I made a followup for this: https://webgate.ec.europa.eu/CITnet/jira/browse/ISAICP-5947

PR #2084, opened 2020-04-07
ISAICP-5374: Bundle specific classes

pfrenssen:


pfrenssen:
Aye I have taken this back on Friday since there is a suggestion on the upstream issue: https://www.drupal.org/project/drupal/issues/2570593

pfrenssen:
Addressed the upstream remarks. Back up for QA. Please also review the upstream patch https://www.drupal.org/project/drupal/issues/2570593, this is the most important part of this ticket.

pfrenssen:
Tagging because this is blocking #2101 - I want to add a `CollectionContentInterface` which is only possible if this is in.

web/modules/custom/joinup_bundle_class/src/JoinupBundleClassFieldAccessTrait.php
```diff
@@ -0,0 +1,58 @@
...
+
+namespace Drupal\joinup_bundle_class;
+
+use Drupal\Core\Field\FieldItemInterface;
+use Drupal\Core\TypedData\Exception\MissingDataException;
+
+/**
+ * Reusable methods for accessing fields in entity bundle classes.
+ */
+trait JoinupBundleClassFieldAccessTrait {
```
claudiu-cristea, line 13: Creating a module just to provide these wrappers feels excessive. For the event use-case the methods could simply use the entity/field API:

```php
public function getWebUrl(): ?Url {
  $field = $this->get('field_event_web_url')'
  return !$field->isEmpty() ? $field->getUrl() : NULL;
}

public function getLocation(): ?string {
  return $this->get('field_location')->value;
}
```
Note that in both cases you are already acknowledge the type of the fields (see the method's return typing). Knowing the field type you know the property that you want.

At least, I'm sure I'll never use them as I'm so used with the standard API. Probably I'll forgot that they exist after this QA. My recommendation is to not add this wrapper but I will not oppose adding it.

pfrenssen: But `isEmpty()` is not enough, there might be a `MissingDataException`. Also `EntityInterface::get()` can return NULL, so we need to check for it. So we will end up with the code from the trait anyway, but then duplicated in every field getter.

PR #2088, opened 2020-04-08
ISAICP-5949: Installation runs out of memory on Drupal 8.9

pfrenssen:
Tests are showing that the installation runs out of memory on Drupal 8.9.0-beta2:

> PHP Fatal error:  Allowed memory size of 1073741824 bytes exhausted (tried to allocate 122880 bytes) in /home/cphp/data/container/8cd9afaf-afa5-4d9a-81d3-d580bb6c7b0b/workspace/web/core/lib/Drupal/Core/Database/Statement.php on line 59

On Drupal 8.8.5 all tests pass without problems.


pfrenssen:
Using git bisect I have identified the first commit that exhibits the problem: [Issue #474684: Allow themes to declare dependencies on modules](https://www.drupal.org/project/drupal/issues/474684).

Couldn't find any reports about this so I created [Issue #3126003: Increased memory usage during install](https://www.drupal.org/project/drupal/issues/3126003)

pfrenssen:
Latest build is green, the problem is fixed!

PR #2090, opened 2020-04-11
Isaicp-5690: Short ID field for collections and solutions.

idimopoulos:


web/modules/custom/solution/config/optional/field.field.rdf_entity.solution.field_short_id.yml
```diff
@@ -0,0 +1,18 @@
...
+dependencies:
+  config:
+    - field.storage.rdf_entity.field_short_id
+    - rdf_entity.rdfentity.solution
+id: rdf_entity.solution.field_short_id
+field_name: field_short_id
+entity_type: rdf_entity
+bundle: solution
+label: 'Short ID'
+description: "Short URL is an abbreviated reference (It serves as a \"nickname\") for the solution, like an acronym. E.g. for a solution named \"European Commission\" it can be \"EuCo\". This value will be displayed in your browser's address bar.\r\n<br />This is not editable after the solution's initial publication.\r\n<br />The value must be between <strong>4 and 26 characters</strong> and can contain only <strong>alphanumeric characters and the dash(-) symbol</strong>. "
```
pfrenssen, line 12: We need to update this description as described above.

web/modules/custom/joinup_rdf/config/install/field.storage.rdf_entity.field_short_id.yml
```diff
@@ -0,0 +1,27 @@
...
+entity_type: rdf_entity
+type: string
+settings:
+  max_length: 255
+  is_ascii: false
+  case_sensitive: false
+module: core
+locked: false
+cardinality: 1
+translatable: true
```
pfrenssen, line 24: * `max_length` 26
* `is_ascii` `TRUE` (I assume)
* `translatable` `FALSE`

tests/features/joinup_group/short_id.feature
```diff
@@ -0,0 +1,130 @@
...
+Feature:
+  As an owner of a group
+  In order to make the URL of my group easier to read
+  I need to be able to provide a short ID to my group.
+
+  Scenario: Short IDs are unique in collections.
+    Given the following collections:
+      | title                     | short ID | state     |
+      | Scientific Studies        | SSt-2020 | validated |
+      | Scientific Sustainability | SST-2020 | validated |
```
pfrenssen, line 11: This is making 2 collections with the same case-insensitive short IDs, shouldn't this be impossible, even on an API level?

tests/features/joinup_group/short_id.feature
```diff
@@ -0,0 +1,130 @@
+@api @terms
+Feature:
+  As an owner of a group
+  In order to make the URL of my group easier to read
+  I need to be able to provide a short ID to my group.
```
pfrenssen, line 5: This describes that the short ID is intended to be used in the URL but the test doesn't look at the URL at any point. Also in the PR I cannot find the code that would actually affect the URL aliases?

Is the changing of the URL intended for a followup ticket or is this handled here? I cannot tell without having access to Jira. It would be strange to deploy this on production if the URLs are not actually using this short ID.

tests/features/joinup_group/short_id.feature
```diff
@@ -0,0 +1,130 @@
...
+
+    # Publish the collection.
+    When I am logged in as a moderator
+    And I go to the "Structural Solar Traces" collection edit form
+    And I press "Publish"
+    Then I should see the heading "Structural Solar Traces"
+
+    # Check that the "Short ID" field is not disabled for moderators.
+    When I click "Edit" in the "Entity actions" region
+    Then the following fields should not be disabled "Short ID"
```
pfrenssen, line 58: Can we also test that the moderator can change the value? Now we know that the field is not disabled, that's already a good thing, but maybe we get an error on save?

web/modules/custom/joinup_core/tests/src/ExistingSite/ShortIdAllowedValuesTest.php
```diff
@@ -0,0 +1,67 @@
...
+    }
+  }
+
+  /**
+   * Data provider for ::testShortIdAllowedValues().
+   */
+  public function shortIdAllowedValuesProvider(): array {
+    return [
+      'starts with dash' => ['-StartWithDash', FALSE],
+      'ends with dash' => ['EndWithDash-', FALSE],
```
pfrenssen, line 53: Let's document that `FALSE` means that it is considered valid, and `TRUE` is considered invalid.

web/modules/custom/joinup_group/src/Plugin/Validation/Constraint/UniqueShortIdInsensitiveValidator.php
```diff
@@ -0,0 +1,66 @@
...
+namespace Drupal\joinup_group\Plugin\Validation\Constraint;
+
+use Drupal\Core\Database\Database;
+use Symfony\Component\Validator\Constraint;
+use Symfony\Component\Validator\ConstraintValidator;
+
+/**
+ * Validates that a field is unique for the given entity type within a bundle.
+ */
+class UniqueShortIdInsensitiveValidator extends ConstraintValidator {
```
pfrenssen, line 14: Let's add a clear warning in the class docblock that this only is intended to work for the short ID field.

config/sync/field.field.rdf_entity.collection.field_short_id.yml
```diff
@@ -0,0 +1,19 @@
...
+    - field.storage.rdf_entity.field_short_id
+    - rdf_entity.rdfentity.collection
+id: rdf_entity.collection.field_short_id
+field_name: field_short_id
+entity_type: rdf_entity
+bundle: collection
+label: 'Short ID'
+description: "Short URL is an abbreviated reference (It serves as a \"nickname\") for the collection, like an acronym. E.g. for a collection named \"European Commission\" it can be \"EuCo\". This value will be displayed in your browser's address bar.\r\n<br />This is not editable after the collection's initial publication.\r\n<br />The value must be between <strong>4 and 26 characters</strong> and can contain only <strong>alphanumeric characters and the dash(-) symbol</strong>. "
+required: false
+translatable: true
```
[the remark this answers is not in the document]
idimopoulos: That is a VERY good point.

tests/features/joinup_group/short_id.feature
idimopoulos: The URL setup is not fixed. There is a follow up for this.

idimopoulos: Sorry I also cannot help you right now without Jira but I know there is a follow up.

tests/features/joinup_group/short_id.feature
idimopoulos: But this is field API. What kind of error would you expect? There is already a uniqueness check in the first time that the entity is saved. There is no reason why the behaviour would change in the second one.

web/modules/custom/joinup_group/src/Plugin/Validation/Constraint/UniqueShortIdInsensitiveValidator.php
idimopoulos: Added it to both the constraint class and the validator class.

web/modules/custom/collection/config/install/field.field.rdf_entity.collection.field_short_id.yml
```diff
@@ -0,0 +1,18 @@
...
+entity_type: rdf_entity
+bundle: collection
+label: 'Short ID'
+description: "Short URL is an abbreviated reference (It serves as a \"nickname\") for the collection, like an acronym. E.g. for a collection named \"European Commission\" it can be \"EuCo\". This value will be displayed in your browser's address bar.\r\n<br />This is not editable after the collection's initial publication.\r\n<br />The value must be between <strong>4 and 26 characters</strong> and can contain only <strong>alphanumeric characters and the dash(-) symbol</strong>. "
+required: false
+translatable: true
+default_value: {  }
+default_value_callback: ''
+settings: {  }
+field_type: string
```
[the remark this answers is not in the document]
idimopoulos: Well, the thing is that this ticket was up for QA right before I pick the refactoring one. I merged the other one wrongly I presume. But for example, I merged it properly for solutions and is now in optional. Moving this to optional as well.

web/modules/custom/joinup_group/joinup_group.module
```diff
@@ -73,15 +90,21 @@ function joinup_group_entity_type_alter(array &$entity_types) {
...
   // Add submit callback when the current bundle is in the list and when we are
   // in an affected operation type.
   $operation_types = ['add', 'default', 'propose'];
-  if (JoinupGroupHelper::isGroup($rdf) && in_array($form_object->getOperation(), $operation_types)) {
+  if (JoinupGroupHelper::isGroup($entity) && in_array($form_object->getOperation(), $operation_types)) {
     $form['actions']['submit']['#submit'][] = 'joinup_group_propose_message_submit';
   }
+
+  if (in_array($entity->bundle(), ['collection', 'solution'])) {
+    if (isset($form['field_short_id']) && $entity->hasGraph('default') && !\Drupal::currentUser()->hasPermission('edit short id after publication')) {
```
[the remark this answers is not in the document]
idimopoulos: I extracted the check to a variable and provided a comment to make it more clear.

web/modules/custom/joinup_group/joinup_group.module
```diff
@@ -30,6 +31,22 @@ use Drupal\views\Plugin\views\query\Sql;
 use Drupal\views\ViewExecutable;
 use Drupal\views\Views;
 
+/**
+ * Implements hook_entity_bundle_field_info_alter().
+ */
+function joinup_group_entity_bundle_field_info_alter(array &$fields, EntityTypeInterface $entity_type, string $bundle): void {
+  $bundles = ['collection', 'solution'];
+  if ($entity_type->id() === 'rdf_entity' && in_array($bundle, $bundles)) {
```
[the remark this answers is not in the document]
idimopoulos: The `joinup_group_entity_bundle_field_info_alter` takes the entity type as a parameter, not an entity, and the bundle as a separate parameter. We cannot use that method here.

web/modules/custom/joinup_test/tests/src/Kernel/JoinupKernelTestBase.php
```diff
@@ -37,6 +37,14 @@ abstract class JoinupKernelTestBase extends RdfKernelTestBase {
...
   ];
 
+  /**
+   * {@inheritdoc}
+   */
+  protected function bootEnvironment() {
+    parent::bootEnvironment();
+    $this->setUpSparql();
+  }
+
```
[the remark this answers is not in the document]
idimopoulos: I simply followed your example for this one. If you think this belongs to RdfKernelTestBase, then why don't all the rest of the tests extend the KernelTestBase and not the RdfKernelTestBase since all of them are using the rdf_entity module anyway.

idimopoulos:
I think I replied/fixed all remarks. Please, excuse me if I missed any.

web/modules/custom/joinup_group/joinup_group.module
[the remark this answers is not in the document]
pfrenssen: Fair point, I will update it to use the `JoinupGroupHelper::GROUP_BUNDLES` constant though so we don't duplicate the array of group bundle names.

web/modules/custom/joinup_test/tests/src/Kernel/JoinupKernelTestBase.php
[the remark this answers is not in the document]
pfrenssen: Because the `RdfKernelTestBase` is setting up the SPARQL connection. Now it is being set up twice.

tests/features/joinup_group/short_id.feature
pfrenssen: :+1:

pfrenssen:
All is looking good now. I fixed the missed remark of the shortened help text, and I enforced the short ID to be lowercase when entered. They will be lowercase when they are used in the URL, so it is better to already have these also lowercase in the UI. This way the user will not have the expectation that they will be somehow mixed case in the URL.

Once this comes back with a green test result this will be ready for UAT.

PR #2102, opened 2020-04-22
QA But not merge: Download administrator report for collections and solutions.

idimopoulos:
Due to the recent outage of Jira, this branch and its commits could not receive a proper number.

Please, QA, but not merge. 
When Jira is back, I will create a ticket and re-submit the commits (no code changes) so that proper log messages are passed into develop.

The scope of the ticket:
Yesterday, the FC asked us for a list of solution facilitators and administrators. We already have this for colllection facilitators but as a flat list without pagination and with export possibility of course.
We did not have this for solutions since it is not possible to directly load them into a single request.

I exported the list manually for solutions but then in the resulting email it was requested that this should be available for moderators.

Initially I created a form as a POC with a "Load more" button and 2 export buttons, 1 for the current list (all the items loaded up to the point the user clicks the button) and 1 for full download. However, this needed a fix in RDF entity and quite some more work so it was rejected as an idea in the standup.

I performed here the following changes(after the suggestions from the standup):
- The UI table is removed. The user does not see any report in the page. They can only download one.
- The route was converted into pointing to a form rather than a controller since the later is not strictly needed.
- I converted the page to work for both collections and solutions at the same time. This was based in a suggestion that everything is sortable in an excel. So we don't need to download 2 different reports as long as the bundle is in the results.
- I do not show the roles the user has. We are only downloading facilitators and administrators. That means that ALL of them is a facilitator. Instead, I present a flag of whether the user is an administrator or not.
- The main change is the approach to the download. We already have an user export list which provides a way to download the list of users from the website. What this does, is adding the list of users in a temporary file and serving it to the user at the very end of the batch process. This is not really useful for the following reasons:
-- The UI is ugly. The page serves the file at the end of the batch without redirecting, meaning that the user still sees the batch process without a way of navigating back to the page. 
-- It has no caching. The file is created and is given to the user with no ability to re-download. Any next attempt from a different moderator would immediately re build the whole list.
On those terms, I went slightly different. I created a mechanism where the user can construct the user list and a cache entry is being created with the results. Currently in Joinup, with more than 4000 administration memberships, the resulting csv entry is a bit more than 1MB so it is quite safe to store this in the database for a while.
Furthermore, the entry remains persistent for a day and the user gets details about when the dump was built and when it will be removed with the option to regenerate the data.

The reason I did not go with the temp files is that we cannot be sure that the file will persist for a required given time. Of course, I could not go with the public directory either as that would be a vulnerability.


idimopoulos:
Closing in favor of a proper numbered branch at https://github.com/ec-europa/joinup-dev/pull/2103

PR #2104, opened 2020-04-23
ISAICP-5900: Strict assignment is required.

idimopoulos:
The original scope of the ticket was that the distribution download button was not working in IE11.

Checking in the console, I found the following error
```
SCRIPT5045: Assignment to read-only properties is not allowed in strict mode
```
Searching a bit around, I found that in IE11 it is not allowed to have read only properties assigned a value and that the code should use an explicit method instead.

As an example from https://stackoverflow.com/a/39898755, in a strict typing (like we are using) you are not allowed to do 
```
this.faAddons.style = '';
```
because style is a read only property and should use the `setAttribute` method instead. The problem starts with only IE11 supporting this while browsers like Chrome allow this. I also could not find a way to force enable this in Chrome and FF.

I fixed this in the first commit and sadly, the Download button still didn't operate properly.
As it seems, the issue is generated in the following code:
```       
$this.on('click', function (event) {
     $(this).trigger('track_download.joinup');
}
```
The above code takes advantage of the link click event and throws a "track_download.joinup" event before the default functionality. The "track_download" event is responsible for showing the modal form for downloading. It uses AJAX to call for the form and display it in the screen.

*Expected behaviour:*
In JS, in order to disable the default functionality of the click event, we should add the `event.preventDefault();` after the triggering of the "track_download" event. However, we don't want that. So the expected behaviour is that after the form is shown, the click will continue, a redirect to the file will occur and the file will be served.

*Actual behaviour:*
The process in IE11 only ends in the triggering of the track download functionality. It does not continue with the rest of the default behaviour of the click event for some reason.. 
I could not find any relevant information and that is the main problem here. There is no indication on when this started. As far as I am concerned, IE11 never allowed this but I do remember this was working in Production :/ I tested out various cases, like checking which click element to click, and even got to a point where I was triggering a click event in the click event. That was causing the event to recurse up to a point but still was halting to its knees as soon as the modal appeared. No file served.
In any case, since this is an exclusive IE11 problem, I used a workaround and trigger the link clicking manually when the user is using IE11 to browse.


web/modules/custom/asset_distribution/js/download_tracking.js
```diff
@@ -30,8 +30,16 @@
           dialogType: 'modal'
         });
 
-        $this.on('click.joinup', function () {
+        $this.on('click', function (event) {
           $(this).trigger('track_download.joinup');
+          // IE11 miss behaves when clicking the link and the default behaviour
```
pfrenssen, line 35:
```suggestion
          // IE11 misbehaves when clicking the link and the default behaviour
```

pfrenssen:
This looks good to me. I cannot however test this on IE11 since I do not have a Windows computer, but as discussed during standup this morning I did a code review and this will be tested by the functional team on the UAT environment.

idimopoulos:
This has been tested in UAT and accepted. Merging.

PR #2108, opened 2020-04-27
ISAICP-5748: Provide a post update to fix the update date.

idimopoulos:
*Original scope of the ticket:*
```
Description: A user has contacted the support because a news item has changed date without apparent reason. The initial date is July and now it is from November. (see support ticket linked)
-- 
Link: [https://joinup.ec.europa.eu/collection/open-source-observatory-osor/news/digital-resilience]

Because this is not the first time that we have this bug, it would be appropriate to know the reason behind those sudden changes.
```

What happened:
In Joinup, we create a revision for every node entity update.
However, up to recently (last October) we had the visit count field as part of the entity which was updating the entity without creating a revision.
In last July we switched to publication_date module as well. We also had two more fixes in August and November for the publication dates.
However, the case reported above in the original post, relieved more cases where the publication date was wrong. There is a total of 514 cases where the publication date was later than the revision creation. 
Because if we are creating a revision in every single entity update, the proper publication date of the entity is the creation date of the revision, which is stored in the `revision_timestamp` of the node revision table.

The cases can be tracked with the following query:
```
SELECT MIN(nfr.vid), nfr.title, from_unixtime(nr.revision_timestamp), from_unixtime(nfr.published_at)
FROM node_field_revision as nfr
LEFT JOIN node_revision as nr ON nfr.vid = nr.vid
WHERE nfr.status = 1 AND nr.revision_timestamp < nfr.published_at
GROUP BY nfr.nid, nr.vid, nfr.title, nr.revision_timestamp, nfr.published_at
```
The post update does the following:
- Queries for the problematic nids and their current publication date as well as the creation date of the first published revision.
- Update all revisions' publication date that much the current publication date (in order to avoid overriding the unpublished versions default publication date) for the given nids.
- Copy over the publication date to the node_field_data table.

The last two steps are based on the original update from the publication_date module.

I did not provide tests as this is a result of the problematic update. The list of problematic entities is available on demand and will be provided to the Functional Contractors.

*For the QA:*
- Rebuild the environment on develop.
- Run the query above verify that there are 514 (with the latest database) cases.
- Checkout ISAICP-5748.
- Execute the updates.
- Run the query above again.
- Verify there are no results.

Why are direct queries used:
- We could not use Drupal API and Entity API to update the entities because we use the changed date all over around and we don't want any changes to occur to the changed date.
- We want to update the publication_date of all versions, not just the entity that is loaded through the API.


claudiu-cristea:
QAing this I discovered that previous attempts to fix this were partially failed. There a lot of nodes with wrong timestamps. I propose that we re-scope this ticket, here's a new scope draft of ISAICP-5748:

## Problem

The `published_at` field should reflect the time when a node has been published for the first time. Given a node with nid == 123, you can get that timestamp with this query:

```sql
-- Query #1 
SELECT nr.revision_timestamp
FROM node_field_revision AS nfr
INNER JOIN node_revision nr ON nfr.vid = nr.vid
WHERE nfr.status = 1
AND nfr.nid = 123
ORDER BY nfr.nid, nfr.vid
LIMIT 0, 1
```

The query to get all nodes with wrong `published_at` time is:

```sql
-- Query #2
SELECT nfd.nid, nfd.title, nr.revision_timestamp, nfd.published_at
FROM node_field_data nfd
INNER JOIN (
  SELECT nfr.nid, nfr.vid, nfr.status
  FROM node_field_revision AS nfr
  WHERE nfr.status = 1
  GROUP BY nfr.nid
  ORDER BY nfr.nid, nfr.vid
) subq ON nfd.nid = subq.nid
INNER JOIN node_revision nr ON subq.vid = nr.vid
WHERE nr.revision_timestamp <> nfd.published_at
ORDER BY nfd.nid
```

But there is another problem. In the past, before installing the `published_date` module, we had in place a `hook_node_presave()` that changed the `created` time to the time of the first published revision. Long story short: this was our `published_at` field. But this was wrong because the `created` time should never be altered, and this was the reason why we installed the `published_date` module. The query to get those nodes is:

```sql
-- Query #3
SELECT nfd.nid, nfd.title, nr.revision_timestamp, nfd.published_at
FROM node_field_data nfd
INNER JOIN (
  SELECT nfr.nid, nfr.vid, nfr.status
  FROM node_field_revision AS nfr
  GROUP BY nfr.nid
  ORDER BY nfr.nid, nfr.vid
) subq ON nfd.nid = subq.nid
INNER JOIN node_revision nr ON subq.vid = nr.vid
WHERE nr.revision_timestamp <> nfd.created
ORDER BY nfd.nid 
```

However, for nodes migrated from the legacy Drupal 6 site were migrated without history (with no revisions). The oldest revision is that which had been created during migration and its timestamp is the migration timestamp. This is introducing a problem: the node `created` time it’s not the same as the oldest revision timestamp. Also, as we don’t know the time when the node has been published we can assume that the data when it has been published for the first time is the `created` time.

## Proposal

I’m proposing a sequence of actions that should run in this order:

* Nodes with `nid < 700000`: Set the oldest revision `revision_timestamp` to node’s `created` time.
* Nodes with `nid >= 700000`: Set the node’s `created` time to the oldest revision `revision_timestamp` time.
* Nodes where the `published_at` is different than the oldest published revision `revision_timestamp` (Query #2): Set node’s `published_at` to the value of the oldest published revision `revision_timestamp`, including the node
revisions where the wrong `published_at` appears.

## Original report

…

claudiu-cristea:
After discussing with @idimopoulos, I will reject this ticket to implement something around my last comment.

idimopoulos:
I knew I had seen it somewhere. Drupal uses the MySQL flag ONLY_FULL_GROUP_BY that we were discussing since early version 8. https://www.drupal.org/project/drupal/issues/567148.
If we want to work without, we need to add the configuration to the settings file. https://www.drupal.org/project/drupal/issues/3098307#comment-13377679 has an example.

idimopoulos:
@claudiu-cristea I think I fixed them all with multiple queries according to the instructions. 
Keep in mind please that I removed tha cache clear and the reindexing. Even though I think most of the cases can easily be tracked by the latest post update (see PR changes), there are a LOT of records changed so I thought that maybe, it would be better to have a complete reset of the content index.

For the results I use the following query
```
SELECT nfr.nid, nfr.vid, nfr.published_at, nr.revision_timestamp
FROM node_field_revision nfr
INNER JOIN node_revision nr ON nfr.vid = nr.vid
WHERE nfr.vid = (
  SELECT MIN(subnfr.vid)
  FROM node_field_revision subnfr
  WHERE subnfr.status = 1 AND subnfr.nid = nfr.nid
  GROUP BY subnfr.nid
)
AND nfr.published_at <> nr.revision_timestamp
```
to verify that there are no revisions where they are published and the revision timestamp is different than the publication date.
Also, I use the query
```
SELECT nr.nid, nr.vid, nr.revision_timestamp, nfr.status, nfr.created, nfd.created as 'nfd.created', nfd.published_at as 'nfd.published_at', nfr.published_at as 'nfr.published_at'
FROM node_field_revision nfr
LEFT JOIN node_revision nr ON nfr.vid = nr.vid
LEFT JOIN node_field_data nfd ON nfd.nid = nr.nid 
WHERE nfr.nid > 700000
ORDER BY nfr.nid ASC, nfr.vid ASC
```
(which is altered easily for nids < 700000 as you can see) to manually check cases.
Your queries seem to be returning empty sets as well.
However, if you take a look in the results manually, for cases with nid > 700000, you will see that everything works smoothly.
The last commit fixes also one broken case, the 702423, where it seems that it has been unpublished and republished again with a new timestamp which is probably one of the problematic revisions.

idimopoulos:
In any case, @claudiu-cristea thanks for the effort in this ticket. It seems I missed quite "a few" (16k) cases.

web/modules/custom/joinup_core/joinup_core.post_update.php
```diff
@@ -1136,3 +1136,81 @@ function joinup_core_post_update_delete_orphaned_solutions() {
...
+  // time upon initial publication.
+  $query = <<<QUERY
+UPDATE {node_field_revision} nfr
+ INNER JOIN (
+    SELECT nid, MIN(revision_timestamp) as revision_timestamp
+    FROM node_revision
+    GROUP BY nid
+) nr ON nfr.nid = nr.nid
+SET nfr.created = nr.revision_timestamp
+WHERE nfr.nid = nr.nid AND nfr.nid >= 700000
```
claudiu-cristea, line 1174: So, we're joining on `nfr.nid = nr.nid` but the we also add this condition to WHERE?

web/modules/custom/joinup_core/joinup_core.post_update.php
```diff
@@ -1136,3 +1136,81 @@ function joinup_core_post_update_delete_orphaned_solutions() {
...
+    MIN(subnfr.vid) as vid,
+    MIN(subnr.revision_timestamp) as revision_timestamp
+  FROM {node_field_revision} subnfr
+  INNER JOIN {node_revision} subnr ON subnfr.vid = subnr.vid
+  WHERE status = 1
+  GROUP BY nid
+  ORDER BY vid
+) s
+SET r.published_at = s.revision_timestamp
+WHERE r.nid = s.nid AND r.vid >= s.vid;
```
claudiu-cristea, line 1205: I'm not sure I understand `AND r.vid >= s.vid`.

web/modules/custom/joinup_core/joinup_core.post_update.php
idimopoulos: This might be a leftover because I did switch from INNER JOIN to raw tables select and back.

web/modules/custom/joinup_core/joinup_core.post_update.php
idimopoulos: Consider this
| nid | vid | status | published_at |
| 1    | 1   | 0         | DEFAULT      |
| 1    | 2   | 1         | 1                    |
| 1    | 3   | 0         | 1                    |
| 1    | 4   | 1         | 1                    |

We had a case where in a node, an intermediate unpublished revision (vid 3 in the example above), had the default value instead of the publication timestamp.
The programmed functionality of the publication date is that once the entity is published, every revision takes the publication date regardless of the status as it _has_ been published in the past.
The `r.vid >= s.vid` means that every revision after the one that was initially published, should receive the publication date.

PR #2110, opened 2020-04-28
ISAICP-5664: Contact form disclaimer message.

idimopoulos:


claudiu-cristea:
The feature is implemented correctly. However, I think this needs a front-end follow-up as the text looks too close to the _Submit_ button. Also wouldn't it be better to put the text just before the button?
Contact___Joinup

PR #2114, opened 2020-05-04
SMEIMKT: EASME project

idimopoulos:


pfrenssen:
Maybe this should use a token with the site name instead of hard coding it everywhere. Then we don't need to redo this for the next projects that wants to reuse Joinup.

PR #2119, opened 2020-05-11
ISAICP-5964: New account cancellation policy

claudiu-cristea:


web/themes/joinup/joinup_theme.theme
```diff
@@ -1092,21 +1092,19 @@ function joinup_theme_preprocess_comment(&$variables) {
...
-        '#cache' => [
-          'tags' => Cache::mergeTags($user->getCacheTags(), $user->getEntityType()
-            ->getListCacheTags()),
-          'contexts' => $user->getCacheContexts(),
-          'max-age' => $user->getCacheMaxAge(),
-        ],
-      ];
+      $variables['profile_icon'] = ['#markup' => $url];
+      $cache_metadata
+        ->addCacheTags($user->getEntityType()->getListCacheTags())
```
idimopoulos, line 1106: As far as I understand you are trying to make the user icon in the comment cache-able but you are only passing the entity type list cache tags. But is this one even needed? The icon should not be affected if _any_ account changes, just the current one, so this should carry the `$user->getCacheTags()` instead. 
I am not sure if the user specific cache tags are invalidated when the user is deleted but I can assume they are.
I see for user 1 for example,
```
>>> $user->getCacheTags();
=> [
     "user:1",
   ]
>>> $user->getEntityType()->getListCacheTags();
=> [
     "user_list",
   ]
>>>
```
 So normally, if the `$user->getCacheTags()` are invalidated during account deletion then these should be enough. But actually, now with the new cancellation-no-deletion method, we don't even need to worry about the deletion.

PR #2122, opened 2020-05-12
ISAICP-5973: Fix entity with double changed time

claudiu-cristea:


claudiu-cristea:
I've fixed the failed test. However, as a 2nd round of QA, few remarks:

* When the country is missing from the profile, the _Country of origin:_ label still shows. It should be hidden for such cases.
Claudiu_Cristea___Joinup

* When the social media icons are totally missing, the _Member for:_ line is touching the bottom of the picture. See the above screenshot. I guess we have to remove the bottom margin from the social media icons and add it to the _Member for: …_, as top margin.

* When the country is present, I see that the label _Country of origin:_ is normal, not bold as the _Organisation:_ or _Domains:_. Also the country value should be inline with the label, as this is a single value field 
Claudiu_Cristea___Joinup

PR #2131, opened 2020-05-19
ISAICP-6003: Fix redirect cache

claudiu-cristea:


web/modules/custom/joinup_community_content/src/CommunityContentWorkflowAccessControlHandler.php
```diff
@@ -411,9 +373,16 @@ protected function getPermissionScheme(string $operation): array {
    *
    * @return \Drupal\node\NodeStorageInterface
    *   The storage handler.
+   *
+   * @throws \Drupal\Component\Plugin\Exception\InvalidPluginDefinitionException
+   *   Thrown if the entity type doesn't exist.
+   * @throws \Drupal\Component\Plugin\Exception\PluginNotFoundException
+   *   Thrown when an entity with a non-existing storage is passed.
```
idimopoulos, line 380: Just for the record, I kinda hate these exceptions in this place.. If the storage is wrong or the entity type doesn't exist, then something is very very broken in the website and it will not wait until the access control handler to throw the exception.. Unless it is a typo, which should not be a typo... Anyway, since it is defined, we should place them.

web/modules/custom/solution/src/Entity/Solution.php
```diff
@@ -20,9 +22,23 @@ class Solution extends Rdf implements SolutionInterface {
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getGroup(): RdfInterface {
     $field_item = $this->getFirstItem('collection');
     if ($field_item->isEmpty()) {
-      throw new MissingCollectionException();
+      throw (new MissingGroupException())->setEntity($this);
```
idimopoulos, line 41: What is the general usage of this exception now? I see that the `SolutionAffiliationFieldItemList` is throwing an exception on preSave, meaning that even on the API level, it is now impossible to save a group without a collection.
Furthermore:
- When the user simply removes the collection (for any reason) with `$entity->set('collection', NULL)` and then calls for `::getGroup` then the exception is not thrown as the condition `$field_item->isEmpty()` fails with "calling method `isEmpty` on NULL`.
- Even if this is fixed, the message is not 100% correct since it is using the ID of the entity. Most of the times, the user will run across this when trying to save a new entity and not an existing. As noted above, since there is a presave exception throw in the field item list, the only way to get this exception on an existing entity is if you deliberately unset the value and then call for the value.
- Even more, there is currently no usage of passing the entity into the exception other than to use it for the message above. The default message within the exception is not overridden to include the entity and the only usage is right above within the entity object that is related to the exception itself.

I get the following errors when trying with a new collection.
```
>>> $solution = \Drupal\rdf_entity\Entity\Rdf::create(['rid' => 'solution', 'label' => 'test', 'field_is_state' => 'validated']);
=> Drupal\solution\Entity\Solution {#7047
     #id: Drupal\Core\Field\FieldItemList {#7053},
     ...
     ...
     #releases_and_distributions: Drupal\solution\SolutionReleasesAndDistributionsFieldItemList {#7000},
   }
>>> $solution->save()
Exception with message 'Solution '' should have a parent collection.'
>>> $solution->getCollection();
PHP Error:  Call to a member function isEmpty() on null in /var/www/html/idimopoulos/joinup/web/modules/custom/solution/src/Entity/Solution.php on line 40
>>> $solution->getGroup();
PHP Error:  Call to a member function isEmpty() on null in /var/www/html/idimopoulos/joinup/web/modules/custom/solution/src/Entity/Solution.php on line 40
>>> 
```

And the following errors if I fix the condition with the `::isEmpty()`.
```
>>> $solution = \Drupal\rdf_entity\Entity\Rdf::create(['rid' => 'solution', 'label' => 'test', 'field_is_state' => 'validated']);
=> Drupal\solution\Entity\Solution {#7047
     #id: Drupal\Core\Field\FieldItemList {#7053},
     ...
     ...
     #releases_and_distributions: Drupal\solution\SolutionReleasesAndDistributionsFieldItemList {#7000},
   }
>>> $solution->save()
Exception with message 'Solution '' should have a parent collection.'
>>> $solution->getCollection();
Drupal/collection/Exception/MissingCollectionException with message 'Solution  missing a parent collection.'
>>> $solution->getGroup();
Drupal/joinup_group/Exception/MissingGroupException
>>> 
```

web/modules/custom/joinup_community_content/src/CommunityContentWorkflowAccessControlHandler.php
claudiu-cristea: It just annoys me to see them in the IDE

web/modules/custom/solution/src/Entity/Solution.php
claudiu-cristea: Good point! I've removed storing the entity in exception and removed the exception messages. Fixed also the ::isEmpty() issue.

PR #2143, opened 2020-05-27
ISAICP-5997: Environment variables

pfrenssen:


pfrenssen:
Ah this wasn't clear to me any more since it was never written down. I though that we had to duplicate the settings for docker. I removed the duplicated configuration.

PR #2145, opened 2020-05-28
ISAICP-5741: Add comparer licence text info modal.

papegaill:


tests/src/Traits/UtilityTrait.php
```diff
@@ -72,6 +72,40 @@ protected function assertNotVisuallyVisible(NodeElement $element): void {
...
+   * @param string $class
+   *   The class.
+   * @param bool $wrap
+   *   When true, wraps the expression between square brackets to be used
+   *   directly as selector. Defaults to true.
+   *
+   * @return string
+   *   The xpath selector.
+   */
+  protected function xpathHasClassSelector(string $class, bool $wrap = TRUE): string {
```
pfrenssen, line 87: This function name makes it seem like an assertion, you would expect it to return a boolean `TRUE` if the selector is present.

BTW I would not use functions to compose XPath selectors, this just makes it more opaque and confusing. Mink is also doing this and it is so difficult to read because you never see the full XPath and instead need to jump between functions to mentally assemble it.

web/modules/custom/joinup_communities/eupl/eupl.behat.inc
```diff
@@ -77,6 +79,43 @@ class EuplSubContext extends DrupalSubContextBase {
...
+   * @param string $spdx
+   *   The SPDX licence ID.
+   *
+   * @throws \Exception
+   *   Thrown if the info icon could not be found.
+   *
+   * @Given I click the info icon of the :spdx licence( table cell)
+   */
+  public function clickLicenceInfoIcon(string $spdx): void {
+    $xpath = '//td[' . $this->xpathHasClassSelector('licence-comparer__header', FALSE) . ' and contains(., "' . $spdx . '")]//span[' . $this->xpathHasClassSelector('icon--info', FALSE) . ']';
```
pfrenssen, line 94: Yeah this is what I mean, let's just put the XPath expression here instead of complicating things by splitting it across different functions.

idimopoulos: Done.

tests/src/Traits/UtilityTrait.php
idimopoulos: Done.

PR #2146, opened 2020-05-28
SMEIMKT-2788: Maximum nesting level on smart_trim

idimopoulos:
**This is ISAICP-6033 on our board.**

Related upstream patches:
https://www.drupal.org/project/smart_trim/issues/3013628

Steps to reproduce:
- Create a collection.
- Go to the creation page of a solution for the above collection.
- In the Description field, add more than 256 HTML elements (just add `<h4>Hello world</h4>` more than 256 times).
- Save. The page is saved properly.
- Go to the collection's page or to the Solutions overview page.

The page is broken due to a problem with maximum nesting level.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -310,6 +310,9 @@
                 "Add a static method that knows to index a given entity @see https://www.drupal.org/project/search_api/issues/3106342": "https://www.drupal.org/files/issues/2020-01-14/3106342-6.patch",
                 "Bulk Form field for SAPI @see https://www.drupal.org/project/search_api/issues/3117171": "https://www.drupal.org/files/issues/2020-04-01/3117171-13.patch"
             },
+            "drupal/smart_trim": {
+                "Maximum function nesting when reading the dom @see https://www.drupal.org/project/smart_trim/issues/301362": "https://www.drupal.org/files/issues/2020-05-28/maximum_function_nesting-3013628-D8-7.patch"
+            },
             "drupal/social_media_links": {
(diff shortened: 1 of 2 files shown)


pfrenssen:
This is a minor issue that will only pop up when XDebug is enabled in the environment and configured to limit the recursion depth.

I am not against merging this, but this is something that will never affect production.

composer.json
```diff
@@ -331,6 +331,9 @@
             "ec-europa/toolkit": {
                 "Variables in 'sequence-file' are not resolved @see https://github.com/ec-europa/toolkit/issues/273": "https://patch-diff.githubusercontent.com/raw/ec-europa/toolkit/pull/274.diff"
             },
+            "drupal/smart_trim": {
+                "Maximum function nesting when reading the dom @see https://www.drupal.org/project/smart_trim/issues/301362": "https://www.drupal.org/files/issues/2020-05-28/maximum_function_nesting-3013628-D8-7.patch"
+            },
```
pfrenssen, line 336: Let's move this up in the right alphabetical location.

composer.lock
```diff
@@ -17949,6 +17443,5 @@
     },
     "platform-overrides": {
         "php": "7.1.9"
-    },
-    "plugin-api-version": "1.1.0"
+    }
```
pfrenssen, line 17446: This seems to have been generated with an older version of Composer. Please regenerate this using the version of Composer that is bundled with Joinup.

pfrenssen:
After a final discussion with the team it was decided to not merge this right now since this is only affecting installations which use XDebug in production, and this is not the case for us. The upstream patch is RTBC so this fix will arrive organically through a module update at some point in the future. We don't need to maintain this patch in the meantime.

PR #2151, opened 2020-06-03
ISAICP-6029: Fix the predicates of the policy domain values.

idimopoulos:


idimopoulos:
Merged by me as it was approved already by Claudiu.

PR #2162, opened 2020-06-11
ISAICP-6046: Update dependencies.

pfrenssen:


composer.json
```diff
@@ -16,25 +16,25 @@
...
+        "drupal/digital_size_formatter": "~2.0",
         "drupal/editor_file": "~1.4",
         "drupal/email_registration": "^1.0",
         "drupal/embed_block": "^1.0",
         "drupal/entity_legal": "^2",
         "drupal/error_page": "^1.0",
         "drupal/facets": "~1.4",
-        "drupal/field_group": "^1.0",
-        "drupal/file_url": "^1.0-alpha7",
+        "drupal/field_group": "^3.0",
```
claudiu-cristea, line 36: We cannot update `field_group` because the `3.x` the vertical tabs description are not displayed & labels CSS styling is different. We'll do this update in ISAICP-5835.

pfrenssen:
I reverted both to the previously used versions. Created ticket to handle the Simplenews update: ISAICP-6059

There are some new patch updates, I had to exchange the patch for the Linkit module with a newer version since the previous one doesn't apply any more.

PR #2168, opened 2020-06-17
Isaicp 5695: Joinup core refactoring - tasty snacks

pfrenssen:


web/modules/custom/joinup_core/joinup_core.install
```diff
@@ -2,7 +2,7 @@
 
 /**
  * @file
- * Install, uninstall, schema and requirements hook for Joinup.
+ * Install, uninstall and requirements hook implementations for Joinup.
```
idimopoulos: Now it is not correct in terms of syntax.

pfrenssen:
Thanks!

PR #2169, opened 2020-06-18
ISAICP-6063: HOTFIX: Drupal vulnerabilities SA-CORE-2020-004, 005, 006

claudiu-cristea:


composer.json
```diff
@@ -23,8 +23,8 @@
         "drupal/config_ignore": "dev-3.x",
         "drupal/config_readonly": "^1.0",
         "drupal/config_update": "dev-1.x",
-        "drupal/core": "^8.8.6",
-        "drupal/core-composer-scaffold": "^8.8.6",
+        "drupal/core": "~8.8.8",
+        "drupal/core-composer-scaffold": "~8.8.8",
```
pfrenssen, line 27: Changing the `^` to `~` for patch releases will prevent us from updating to Drupal 8.9 but this is not so important for this hotfix. We should remove this again when merging back to develop.

PR #2175, opened 2020-06-22
ISAICP-5803: Pipeline step plugin reading triples from a file/URL

claudiu-cristea:


web/modules/custom/joinup_federation/src/Plugin/pipeline/Step/ImportFromFileStep.php
```diff
@@ -0,0 +1,104 @@
...
+      $container->get('entity_type.manager'),
+      $container->get('http_client')
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function execute() {
+    $resource = $this->getConfiguration()['resource'] ?? NULL;
```
idimopoulos, line 76: This is really not documented anywhere other than the exception below.

web/modules/custom/joinup_federation/src/Plugin/pipeline/Step/ImportFromFileStep.php
```diff
@@ -0,0 +1,104 @@
...
+
+/**
+ * Defines a step plugin that imports data from a local or remote file.
+ *
+ * @PipelineStep(
+ *   id = "file_import",
+ *   label = @Translation("Import from file"),
+ * )
+ */
+class ImportFromFileStep extends JoinupFederationStepPluginBase {
```
idimopoulos, line 26: So this takes either a file URI (as a local resource URI) or an external URI. Yet, it requires the devs to pre-configure the plugin with the final URL - meaning that in case of a URL change we would have to deploy a new file. Or in case of a local file update, we would still need to deploy a change because it might be that a `_1.rdf` is added to the file.
Furthermore, we have the ManualUpload step which is already doing this in a configurable way. Why not merge them both, use the file/remote_file widget we have already developed in order to make this also configurable?

idimopoulos: I discussed this with Claudiu and the step is intended for pipelines that run in an unmanaged mode. Thus, no interaction with the user should be needed to run this step.

web/modules/custom/joinup_federation/src/Plugin/pipeline/Step/ImportFromFileStep.php
claudiu-cristea: I think it misses the defaultConfiguration()

claudiu-cristea: Added `defaultConfiguration()` and documentation.

PR #2182, opened 2020-06-30
ISAICP-4133: Test that events that are currently taking place are sho…

pfrenssen:
…wn in the list of 'Upcoming events' and not in 'Past events'.

### DIFF
--- a/web/modules/custom/joinup_search/joinup_search.module
+++ b/web/modules/custom/joinup_search/joinup_search.module
@@ -170,7 +170,7 @@ function _joinup_search_search_api_has_condition($conditions, $field, $value, $o
  *
  * Do not index any content not written in English.
  */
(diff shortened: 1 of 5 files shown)


web/modules/custom/joinup_search/joinup_search.module
```diff
@@ -170,7 +170,7 @@ function _joinup_search_search_api_has_condition($conditions, $field, $value, $o
  *
  * Do not index any content not written in English.
  */
-function joinup_core_search_api_index_items_alter(IndexInterface $index, array &$items): void {
+function joinup_search_search_api_index_items_alter(IndexInterface $index, array &$items): void {
```
claudiu-cristea, line 173: Nice catch :)

PR #2184, opened 2020-07-01
ISAICP-5798: Replace 'Deletion request' with straight up 'Delete'

pfrenssen:


web/modules/custom/joinup_group/src/EventSubscriber/JoinupGroupOgSubscriber.php
```diff
@@ -41,4 +43,25 @@ public function provideOgGroupPermissions(OgPermissionEventInterface $event): vo
...
 
+  /**
+   * Gives moderators access to view, create, edit and delete all group content.
+   *
+   * @param \Drupal\og\Event\GroupContentEntityOperationAccessEventInterface $event
+   *   The event that fires when an entity operation is being performed on group
+   *   content.
+   */
+  public function moderatorsCanManageGroupContent(GroupContentEntityOperationAccessEventInterface $event): void {
+    $is_moderator = in_array('moderator', $event->getUser()->getRoles());
```
claudiu-cristea, line 54: I really don't like this! I'm repeating myself: don't check for roles, check for permissions. Isn't this equivalent to setting standard site-wide node permissions to `moderators`, such as `create news content`, `edit any document content`, etc? What I'm missing here? A site-wide permission should ***ALWAYS*** override an OG permission. It's possible that I don't get the purpose of this subscriber. If I'm right that the subscriber's job is to allow moderators to operate on group content and the site-wide permissions are not enough then It might be a bug in OG.

claudiu-cristea:
When I'm rebuilding with the PROD database, I'm getting:

```
 [error]  Drupal\Core\Config\ConfigImporterException: There were errors validating the config synchronization.
Entities exist of type <em class="placeholder">Message</em> and <em class="placeholder">Message template</em> <em class="placeholder">Solution request deletion</em>. These entities need to be deleted before importing.
Entities exist of type <em class="placeholder">Message</em> and <em class="placeholder">Message template</em> <em class="placeholder">Solution deletion - no approval</em>. These entities need to be deleted before importing.
Entities exist of type <em class="placeholder">Message</em> and <em class="placeholder">Message template</em> <em class="placeholder">Solution deletion approved</em>. These entities need to be deleted before importing. in Drupal\Core\Config\ConfigImporter->validate() (line 755 of web/core/lib/Drupal/Core/Config/ConfigImporter.php).
```

web/modules/custom/joinup_group/src/EventSubscriber/JoinupGroupOgSubscriber.php
pfrenssen: This was already like this, this code has just moved from `joinup_og_user_access_alter()` to here, unchanged. This was necessary to accommodate the latest OG code which no longer fires this hook for entity operations.

config/sync/collection.settings.yml
```diff
@@ -25,6 +49,8 @@ transitions:
     validated:
       - moderator
       - rdf_entity-collection-facilitator
+  # In the following to-state, the facilitator will not actually have permission to execute the transition if there is
+  # not already a published version of the collection.
   validated:
```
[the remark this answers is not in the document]
pfrenssen: Yeah I know, this is how the access system has been implemented years ago. It is not in scope of this ticket to fix that. I am just moving the documentation here from its original location in the collection `config/install` legacy config so it doesn't get lost.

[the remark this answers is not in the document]
pfrenssen: Created a followup to tackle this issue: [ISAICP-6154: Rework workflow transitions to be permission based rather than role based](https://citnet.tech.ec.europa.eu/CITnet/jira/browse/ISAICP-6154).

web/modules/custom/joinup_group/src/Entity/GroupTrait.php
```diff
@@ -0,0 +1,51 @@
...
+    if (!isset($this->memberships[$uid])) {
+      $og_membership_storage = $this->entityTypeManager()->getStorage('og_membership');
+
+      $query = $og_membership_storage
+        ->getQuery()
+        ->condition('uid', $uid)
+        ->condition('entity_type', $this->getEntityTypeId())
+        ->condition('entity_id', $this->id());
+
+      $membership_ids = $query->execute();
```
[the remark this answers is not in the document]
pfrenssen: I did this because of performance reasons, the OG method is inefficient since it loads all the user's memberships and filters them down. But I agree it is not our responsibility to do micro optimization for this. Let's go with the simple solution.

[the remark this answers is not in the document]
pfrenssen: I created a ticket in OG to solve the performance issue: https://github.com/Gizra/og/issues/696, I will simplify our code with a call to `MembershipManager::getMembership()`.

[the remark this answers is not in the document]
pfrenssen: I have updated the code. I have also removed the static cache since it is also not our responsibility to maintain a per-group membership cache.

web/modules/custom/joinup_group/src/EventSubscriber/JoinupGroupOgSubscriber.php
pfrenssen: I looked at this a bit closer and it seems indeed to be unnecessary, this was probably needed to work around some oddities in the access calculation, but after the recent improvements in `OgAccess` this should no longer be required.

I have tentatively removed this entire section, let's see what the tests say.

pfrenssen: Code is gone, tests are still green! Goodbye :grin:

web/modules/custom/joinup_group/src/Entity/GroupTrait.php
```diff
@@ -12,40 +12,15 @@
...
-      /** @var \Drupal\og\OgMembershipInterface|NULL $membership */
-      $membership = $og_membership_storage->load($membership_id);
-
-      $this->memberships[$uid] = $membership;
-    }
 
-    return $this->memberships[$uid];
+    /** @var \Drupal\og\MembershipManagerInterface $membership_manager */
+    $membership_manager = \Drupal::service('og.membership_manager');
+    return $membership_manager->getMembership($this, $uid, [OgMembershipInterface::STATE_ACTIVE]);
```
claudiu-cristea, line 23: With the previous code all memberships, regardless of their state, were returned. Now, I see we're limiting to the active states. Shouldn't we better use `OgMembershipInterface::ALL_STATES`?

pfrenssen: Yes you are right, we should also return blocked memberships, otherwise calling code cannot retrieve a blocked membership if they need it.

pfrenssen: I updated this, but I made it an option that can be passed by the calling code, and defaulting to not returning a blocked membership. I remember discussing this flag with @amitaibu at an earlier occasion and he convinced me that it is safer to exclude blocked users by default. The reasoning is as follows:

If blocked users are excluded by default, then if a programming mistake is made (e.g. the developer is not considering that a blocked membership might be returned) then blocked users are not going to be given access to some new feature by default. This is the most common case, blocked users are treated the same way as non-members for most functionalities. So the worst case scenario is that a blocked user is not given access to something they should have access to. This can be annoying but is not critical.

On the other hand, if we would default to returning a blocked membership, then if a programming mistake is made, then a blocked user would be accidentally given access to something they should __not__ have access to. This can lead to critical problems like information disclosure.

So it would be safer like this, and it is also consistent with `MembershipManagerInterface::getMembership()`.

Is this OK for you?

claudiu-cristea: @pfrenssen yes, makes totally sense

amitaibu: I don't remember that specific conversation, but yeah - it makes a lot of sense :smile_cat:

PR #2190, opened 2020-07-09
ISAICP-6107: Fatal errors when a collection with pinned content is deleted

pfrenssen:


pfrenssen:
I rebased this to have the test as the first commit so it can prove that it correctly detects the bug, but the order of commits is not displayed correctly in Github. For reference here is the failing test: https://app.continuousphp.com/git-hub/ec-europa/joinup-dev/build/585a8cbe-3b29-4177-befb-197ece9798c2/logs/e2cf0430-b24e-49b4-a15e-ed8d41469f40

pfrenssen:
Background information about the refactoring alongside the bugfix:

This bug is caused by putting too much trust the return value of `$entity->get('some entity reference field')->entity`. This is not guaranteed to return an entity, it can also return NULL.

When working on this I saw an earlier attempt to make this more bullet proof but it was not successful:

```
public function getGroup(): RdfInterface {
  $field_item = $this->getFirstItem(OgGroupAudienceHelperInterface::DEFAULT_FIELD);
  if (!$field_item || $field_item->isEmpty()) {
    throw new MissingGroupException();
  }
  return $field_item->entity;
}
```

Here a double check was being done: first check that `$field_item` itself is not empty, and then check with `$field_item->isEmpty()` that we have a value.

But even in this case it can return `NULL`, because if the referenced entity has been deleted, then the field value is considered not empty, but it cannot load the entity, and we get `NULL`.

We have this kind of code all over Joinup, we are putting too much trust into these magic Field API properties. So I decided to solve this by writing a method in a reusable trait that we can use to retrieve the entities. This even ended up not using the magic `->entity` property but rather `->referencedEntities()` which is more reliable since it doesn't return NULL but an empty array which is more robust to handle from the calling side:

```
  protected function getReferencedEntities(string $field_name): array {
    try {
      $item_list = $this->getEntityReferenceFieldItemList($field_name);
    }
    catch (\InvalidArgumentException $e) {
      $this->logException($e);
      return [];
    }

    return $item_list->referencedEntities();
  }
```

I also opted to catch exceptions and log an error for them where feasible, since recent experiences have shown that exceptions need to be avoided if possible. We don't want pages to break and cron jobs to be interrupted when some content is deleted and our code is not prepared for it. Let's log an error and allow Drupal to continue its job.

PR #2196, opened 2020-07-14
ISAICP-4698: Refactor code that notifies user when their account is changed by a moderator

pfrenssen:


pfrenssen:
I started by completing the test coverage for this feature before starting on the refactoring, so that we can assess that this is still working correctly after the refactoring. The tests can be found here:

* `user/profile.feature`, test `A moderator can navigate to any user's profile and edit it`: covers the edit form case.
* `user/roles.feature`, test `A moderator can assign the moderator role`: covers the bulk operation case.

PR #2198, opened 2020-07-15
6072: Glossary filter

claudiu-cristea:


web/modules/custom/collection/src/Plugin/Filter/Glossary.php
```diff
@@ -0,0 +1,248 @@
...
+    }
+
+    return $map;
+  }
+
+  /**
+   * Stores the collection context.
+   */
+  protected function setCollection(): void {
+    if ($this->collection = $this->ogContext->getGroup()) {
```
pfrenssen: This is not 100% reliable since these texts can also be shown outside of a collection (e.g. in the search results which have no group context). With the render cache in play this could lead to unexpected results - for example if some content is rendered as a tile and is first displayed in the search results it will be cached without having its glossary terms replaced. Then when the cached render array is later used inside the collection context the terms will still be missing.

I am going to let this pass though, for the following reasons:

- I could not replicate this in local testing. Search results and other places where tiles are shown (e.g. Keep Up To Date) correctly show the content without glossary terms. Possibly there is a render context that causes variations to be cached.
- This is a [known problem](https://www.drupal.org/project/drupal/issues/226963) in Drupal for a very long time, and will be incredibly hard to fix, way out of scope for this PR.
- I created a followup to look into this in some more depth, we can skip it for now: ISAICP-6116

web/modules/custom/collection/src/Plugin/Filter/Glossary.php
```diff
@@ -0,0 +1,248 @@
...
+    if (!isset($this->collection)) {
+      return $result;
+    }
+
+    // This collection has no glossary term entries.
+    if (!$replacements = $this->getReplacementsMap()) {
+      return $result->addCacheableDependency($this->cacheMetadata);
+    }
+
+    $pattern = '/\b(' . implode('|', array_keys($replacements)) . ')\b/';
```
pfrenssen, line 111: This regex might become a problem when glossaries grow to thousands of terms, but I'm going to let this pass since this is probably never going to happen.

web/modules/custom/collection/src/Plugin/Filter/Glossary.php
```diff
@@ -0,0 +1,248 @@
...
+  /**
+   * Finds-out if a piece of text is inside a HTML link.
+   *
+   * @param \DOMText $text_node
+   *   The text node to be checked.
+   *
+   * @return bool
+   *   If the given text node is inside a HTML link.
+   */
+  protected function isLinkText(\DOMText $text_node): bool {
```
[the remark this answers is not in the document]
claudiu-cristea: Great point! I think there are other places where we're using this lib.

web/modules/custom/collection/src/Entity/GlossaryTermInterface.php
```diff
@@ -0,0 +1,41 @@
...
+declare(strict_types = 1);
+
+namespace Drupal\collection\Entity;
+
+use Drupal\node\NodeInterface;
+
+/**
+ * Interface for glossary terms in Joinup.
+ */
+interface GlossaryTermInterface extends NodeInterface {
```
claudiu-cristea, line 12: Shouldn’t GlossaryTermInterface extend also GroupContentInterface?

web/modules/custom/collection/src/Plugin/Filter/Glossary.php
```diff
@@ -194,17 +194,17 @@ protected function getReplacementsMap(): array {
...
           'url' => $glossary->toUrl()->toString(),
-          'summary' => $this->getSummaryOrTrimmedDefinition($glossary),
+          'summary' => $glossary->getSummary(),
         ];
 
         $map[$glossary->label()] = $link;
         // Link also the abbreviation, if any.
-        if (!$glossary->get('field_glossary_abbreviation')->isEmpty()) {
-          $map[$glossary->get('field_glossary_abbreviation')->value] = $link;
+        if ($glossary->hasAbbreviation()) {
```
claudiu-cristea, line 211: ❤️👍😍

web/modules/custom/collection/src/Plugin/Filter/Glossary.php
claudiu-cristea: OK, let's see the feedback. The customer is doesn't have yet a 100% opinion on the business logic.

web/modules/custom/collection/src/Entity/GlossaryTermInterface.php
pfrenssen: I only added the methods strictly needed in scope of this PR, but yes we should add this when the need arises.

PR #2188, opened 2020-07-08
ISAICP-6023: Remove the Eurostat solutions

pfrenssen:
This was requested by the Eurostat team.

PR #2207, opened 2020-07-27
ISAICP-6023

idimopoulos:
Original work and discussion took place in https://github.com/ec-europa/joinup-dev/pull/2188
The ticket was then requested to be merged against master so a separate branch and PR was constructed in order to 1. not delete the original discussion and 2. resolve the conflicts.


idimopoulos:
I am merging this after our discussion with @pfrenssen that we should revert the exception logging that was removed due to changing the base branch.

PR #2209, opened 2020-07-28
ISAICP-6127: Update develop after release v1.62.1

idimopoulos:


web/modules/custom/joinup_core/joinup_core.post_update.php
```diff
@@ -7,10 +7,99 @@
...
+
+/**
+ * Remove a selected set of solutions and 1 document from Eurostat.
+ */
+function joinup_core_post_update_0106300(array &$sandbox): void {
+  $entity_type_manager = \Drupal::entityTypeManager();
+
+  // Gather the IDs of the solutions that need to be removed, according to
+  // client specifications.
+  if (!isset($sandbox['ids'])) {
```
pfrenssen, line 20: This update has run when 1.62.1 was deployed so it can be removed.

web/modules/custom/joinup_core/joinup_core.post_update.php
```diff
@@ -7,10 +7,99 @@
...
+  $queue_worker->process();
+
+  $sandbox['#finished'] = !empty($sandbox['ids']) ? $sandbox['progress'] / $sandbox['total'] : 1;
+}
+
 /**
  * Fix the last updated time of node entities.
  */
-function joinup_core_post_update_0106300(&$sandbox) {
+function joinup_core_post_update_0106301(&$sandbox) {
```
pfrenssen, line 102: We can keep this on `_0106300` since this will be the first update to run when deploying 1.63.0

PR #2210, opened 2020-08-02
ISAICP-6085: Featured menu items for the support menu.

idimopoulos:


tests/src/Context/WhatsNewContext.php
```diff
@@ -0,0 +1,79 @@
...
+   *
+   * @Then the :link_title link should be featured as what's new
+   */
+  public function assertSupportMenuLinkWhatsNewClass(string $link_title): void {
+    foreach ($this->getFeaturedSupportLinks() as $element) {
+      if ($element->getText() === $link_title) {
+        return;
+      }
+    }
+    throw new \Exception("No link with title '{$link_title}' has been found marked as featured in the support menu.");
```
claudiu-cristea, line 49: This makes the IDE yell that docblock contains no`@throws` tag. And IDE is right because we should start doing better with our custom step definition. When testing, there's a big difference between a test ending with **error** vs. **failed**. From https://phpunit.readthedocs.io/en/9.2/textui.html:

> PHPUnit distinguishes between failures and errors. A failure is a violated PHPUnit assertion such as a failing `assertSame()` call. An error is an unexpected exception or a PHP error. Sometimes this distinction proves useful since errors tend to be easier to fix than failures.

So, when the expectation is not met we should use `ExpectationFailedException`. Or when an element that we want to test is missing, we should use `ElementNotFoundException`. This allows better understanding of what's going on as...

> If you have a big list of problems, it is best to tackle the errors first and see if you have any failures left when they are all fixed.

tests/src/Context/WhatsNewContext.php
```diff
@@ -0,0 +1,79 @@
...
+   *
+   * @param string $link_title
+   *   The link title.
+   *
+   * @Then the :link_title link should not be featured as what's new
+   */
+  public function assertSupportMenuLinkNotWhatsNewClass(string $link_title): void {
+    foreach ($this->getFeaturedSupportLinks() as $element) {
+      if ($element->getText() === $link_title) {
+        throw new \Exception("The link with title '{$link_title}' has been found marked as featured but should not.");
```
claudiu-cristea, line 63: Same.

web/modules/custom/whats_new/src/WhatsNewHelper.php
```diff
@@ -0,0 +1,94 @@
...
+  public function hasFlagEnabledMenuLinksForEntity(EntityInterface $entity): bool {
+    $entity_uris = [
+      // Support both the alias and the internal URL. The alias can occur if the
+      // user copies it from the URL and the internal URL can occur if the user
+      // selects the entity from the autocomplete field.
+      'internal:' . $entity->toUrl()->toString(),
+      "entity:node/{$entity->id()}",
+    ];
+    $query = $this->entityTypeManager->getStorage('menu_link_content')->getQuery()
+      ->condition('menu_name', 'support')
```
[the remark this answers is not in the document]
idimopoulos: Indeed this is only meant for the support menu. At least for now. If needed however, it is easy to switch and add the menu. The support menu is anyway used as a hardcoded menu in the whats_new module.

PR #2214, opened 2020-08-06
ISAICP-6148: Error message from requirements check on updatedb

claudiu-cristea:


claudiu-cristea:
Hm... this outputs something like:

```
[error]   (Currently using Incompatible modules The following modules are installed, but they are incompatible with PHP 7.1.33:<ul><li>Schema.org Article</li><li>Schema.org Event</li><li>Schema.org Metatag</li></ul>Review the <a href="https://www.drupal.org/docs/8/update/troubleshooting-database-updates"> suggestions for resolving this incompatibility</a> to repair your installation, and then re-run update.php.) 
```

Is there a way to transform HTML in console readable text?

PR #2221, opened 2020-08-20
ISAICP-5939: New author role for trusted content authorship.

idimopoulos:


config/sync/core.entity_view_display.rdf_entity.collection.about.yml
```diff
@@ -139,17 +140,15 @@ content:
...
       format: default
-      format_custom_true: ''
       format_custom_false: ''
-    third_party_settings:
-      template_suggestion:
-        template_suggestion: fieldset
-    type: boolean
-    region: content
+      format_custom_true: ''
+    third_party_settings: {  }
```
pfrenssen, line 151: These changes are causing a regression in the about page:

moderation-label

The label is now shown, but this duplicates the label that is already there.

config/sync/field.field.rdf_entity.collection.field_ar_moderation.yml
```diff
@@ -11,7 +11,7 @@ id: rdf_entity.collection.field_ar_moderation
 field_name: field_ar_moderation
 entity_type: rdf_entity
 bundle: collection
-label: Moderation
+label: Moderated
```
pfrenssen, line 14: Ah I might be mistaken in my comment above, I possibly mixed up my screenshots. I agree that this should be "Moderated" and not "Moderation".

config/sync/og.og_role.rdf_entity-collection-author.yml
```diff
@@ -0,0 +1,34 @@
...
+  - 'delete own event content'
+  - 'delete own news content'
+  - 'delete own news content'
+  - 'edit own custom_page content'
+  - 'edit own discussion content'
+  - 'edit own document content'
+  - 'edit own event content'
+  - 'edit own news content'
+  - 'post comments'
+  - 'view own unpublished content'
```
pfrenssen, line 33: All of these permissions are already present in the standard member role, so they should not be duplicated here.

web/modules/custom/joinup_community_content/joinup_community_content.module
```diff
@@ -133,6 +133,14 @@ function joinup_community_content_workflow_selector(EntityInterface $entity): ?s
     return NULL;
   }
 
+  $user_id = \Drupal::currentUser()->id();
+  /** @var \Drupal\og\OgMembershipInterface $membership */
+  if ($membership = \Drupal::getContainer()->get('og.membership_manager')->getMembership($parent_entity, $user_id)) {
+    if ($membership->hasRole($parent_entity->getEntityTypeId() . '-' . $parent_entity->bundle() . '-author')) {
+      return 'node:post_moderated';
+    }
+  }
```
pfrenssen, line 142: This is hacky so it is a good idea to explain why it is necessary that we mess with the post_moderated state depending on the membership.

web/modules/custom/joinup_core/joinup_core.post_update.php
```diff
@@ -48,3 +48,38 @@ function joinup_core_post_update_0106400(): void {
...
+INSERT { ?entity_id <http://joinup.eu/collection/new_role> "rdf_entity-collection-author" }
+WHERE { ?entity_id <http://joinup.eu/collection/moderation> "false"^^<http://www.w3.org/2001/XMLSchema#boolean> }
+QUERY;
+
+    $connection->query($enabled);
+    $connection->query($disabled);
+  }
+
+  \Drupal::entityTypeManager()->getStorage('rdf_entity')->resetCache($ids);
+}
```
pfrenssen, line 85: This is not clear to me, can you explain why this is needed? The non-moderated collections don't need the author role added to function like expected so this update hook seems to be useless.

It is also not asked in the ticket that we should assign the author role to all members of collections that are not moderated.

pfrenssen (changes requested):

web/modules/custom/joinup_community_content/joinup_community_content.module
idimopoulos: I have added a commend though this is not really hacky. This is the whole meaning of the ticket. Having a role that behaves as a post moderated role in any group.

pfrenssen:
This is OK if tests come back green.

pfrenssen (approved):

pfrenssen (approved):

PR #2227, opened 2020-08-27
ISAICP-6173: Clear cached values before checking the changes.

idimopoulos:


tests/src/Traits/MailConfigTrait.php
```diff
@@ -34,6 +34,28 @@ protected function checkMailConfigOverride(): void {
...
+   * Clears the config values from both cache and $_GLOBALS array.
+   *
+   * This ensures that when values are overridden or changed in the settings.php
+   * file during the test run, the values are being read anew from storage.
+   * Config factory sets overrides both in cache and in the $_GLOBALS array so
+   * cleaning them would require to remove them from both indexes.
+   *
+   * When we are changing the settings values before running behat tests, the
+   * values are already loaded into cache and $_GLOBALS so checking for
+   * overrides will still fail.
```
pfrenssen, line 47: If this is needed because the settings are being changed, then it's better to call this code at the time the settings change, instead of at the time the settings are read. Otherwise this will be a problem every time we read from the settings file and forget to call this method.

pfrenssen:
I have looked into this. The current approach is not valid since it is not addressing the root cause. It is only affecting the `$config` part of the settings file, and is ignoring everything else, like settings, error handlers, inclusion of `settings.local.php` etc.

So the problem is that the Drupal Extension is bootstrapping Drupal before we have the chance to replace the `settings.php` file:

* The file replacement happens in `FeatureContext::addBehatSpecificDrupalSettings()` which is a `@beforeSuite` hook.
* Right before the `beforeSuite` hooks are firing DrupalExtension bootstraps Drupal. This is happening inside `Reader::findSubContextClasses()` which needs to query Drupal for the available modules so it can decide which subcontexts to load.
* We cannot yet disable subcontexts since we are using them, but this is a deprecated feature, so for the sake of experiment I went deeper here, and disabled subcontexts by adding the `subcontexts.autoload: false` option to `behat.yml`. This successfully manages to start firing the `beforeSuite` hooks without bootstrapping Drupal, but then a second problem arises:
* We have other `beforeSuite` hooks that rely on Drupal being bootstrapped at the moment they fire. Some examples are `JoinupEuLoginContext::disableLimitedAccess()` and `FeatureContext::disableAntibotForSuite()`.
* We cannot determine the order in which hooks are fired in Behat, so we cannot reliably ensure that our hook to alter the Drupal settings will fire before the others. There is a feature request for this though (https://github.com/Behat/Behat/issues/62) but implementing this is out of scope for this ticket.

So what I would propose to solve this:

* In scope of this ticket, let's make a task runner command that replaces the settings file at the start. This is less convenient but it will at least allow us to run tests with a single command. I'm not sure if it is possible to restore the settings to `site-clean` after the tests.
* Let's make a followup to make it possible to replace the settings files by running the Behat command itself, like was intended now, but this would require us to implement the hook ordering in Behat.

claudiu-cristea:
> * let's make a task runner command that replaces the settings file at the start.

Oh, this is very bad. I often run a single scenario directly from PhpStorm and I would the suite hooks to do their job.

pfrenssen:
This will not work, like I explained in my previous comment Drupal is already bootstrapped _before_ the `beforeSuite` hooks are firing. I explored this option only because I know this feature will be deprecated at some point in the future, so this could be a possible solution once DrupalExtension 5.x is released. We can force it now by setting `subcontexts.autoload: false` but then we lose all the third party subcontexts, so tests using Honeypot, CAS mock server, digest messages, groups etc will no longer work.

BTW that comment from @everzet was from a simpler time, when projects were small, had only one `FeatureContext` file and were not using any third party extensions. This did not anticipate the present day reality. We have easily 100+ hooks scattered across 50+ contexts provided by us and a dozen third party Drupal modules.

pfrenssen:
In a private discussion with @claudiu-cristea we came up with a plan: let's try to disable the discovery of the third party subcontexts and include them directly in our `behat.yml` file. If this works we can possibly shuffle the contexts in a way that will allow to read the updated `settings.php` file before any of the other hooks are fired.

pfrenssen:
Created a PR with the second attempt, but it appears our trick is not working: #2228

pfrenssen:
Closing in favor of #2228.

PR #2228, opened 2020-08-31
ISAICP-6173 - attempt II

pfrenssen:


pfrenssen:
Created a sister PR that will apply all the patches that add Behat context classes in addition to the now deprecated subcontext classes to contrib modules: https://github.com/ec-europa/joinup-dev/pull/2235
This will ensure that none of these patches are causing regressions.

claudiu-cristea (approved):
Nice work, I've only added a follow-up (see my inline comments)

PR #2245, opened 2020-09-08
ISAICP-5965: remove gulp and prototype folders.

papegaill:


claudiu-cristea (commented):
We need to exclude the scss/ also when building the distribution (i.e. in .gitattributes).

claudiu-cristea (approved):

PR #2251, opened 2020-09-15
ISAICP-5677: Show the external button in distribution tiles.

idimopoulos:


pfrenssen (approved):
Looking good. There is one edge case which is not tested (an external distribution can also be shown on the solution page if it is the only distribution in a release), but this is of minor importance and not specifically demanded in the ticket description. We can handle this in a followup if needed.

PR #2252, opened 2020-09-16
ISAICP-6143: Licence comparer plugin type

pfrenssen:


idimopoulos:
Very nice testing :D It will be a pain to change them in the future if needed, but it is much better having more cases :)

PR #2253, opened 2020-09-17
ISAICP-6197: Update old aliases.

idimopoulos:


claudiu-cristea (changes requested):
# Problems

## Duplicate aliases (fixed by me)

I found a big mess in our `path_alias` table. There are many EN duplicated aliases. I have no idea how we went into this. You can find the duplicates with this query:

```sql
SELECT MAX(id) AS valid_id, COUNT(*) AS duplicates_count, path
FROM path_alias
WHERE langcode = 'en'
GROUP BY path
HAVING duplicates_count > 1
ORDER BY duplicates_count DESC
```

I've added a new post update to cleanup the table before we recompute the aliases.

## Wrong non-EN aliases (not fixed)

After running the updates, I've noticed >500 non-English aliases that were wrongly computed. It seems that a token is not resolved. You can find them by running:

```sql
SELECT * FROM path_alias WHERE alias LIKE '/rdf_entity/%'
``` 

## Creating redirects is superfluous (fixed by me)

See the inline comment.

## Node aliases should be computed as well (not fixed)

See the inline comment.

## Not all RDF entity aliases are computed (not fixed)

After running updates, try this:

```sql
SELECT * FROM path_alias WHERE path LIKE '/rdf_entity/%' AND alias REGEXP '^/(solution|release|distribution)/.*'
```

idimopoulos:
Closing this PR in favor of a new upcoming one against develop.

PR #2259, opened 2020-09-23
ISAICP-6197: Recompute path aliases

claudiu-cristea:


web/modules/custom/joinup_core/joinup_core.post_update.php
```diff
@@ -48,3 +48,103 @@ function joinup_core_post_update_0106401(): void {
...
+  if (empty($sandbox['entity_ids'])) {
+    $sandbox['entity_ids']['rdf_entity'] = $rdf_storage->getQuery()->condition('rid', 'collection')->execute();
+    $sandbox['entity_ids']['rdf_entity'] += $rdf_storage->getQuery()->condition('rid', 'solution')->execute();
+    $sandbox['entity_ids']['rdf_entity'] += $rdf_storage->getQuery()->condition('rid', 'asset_release')->execute();
+    $sandbox['entity_ids']['rdf_entity'] += $rdf_storage->getQuery()->condition('rid', [
+      'collection',
+      'solution',
+      'asset_release',
+    ], 'NOT IN')->execute();
+    $sandbox['entity_ids']['node'] = [];
```
claudiu-cristea, line 119: Oops! After running the post update I'm doing:

```sql
SELECT * FROM path_alias WHERE alias NOT LIKE '/collection/%' AND path LIKE '/node/%'
```

...and I'm still getting ~6.600 records.

web/modules/custom/joinup_core/joinup_core.post_update.php
```diff
@@ -48,3 +48,103 @@ function joinup_core_post_update_0106401(): void {
...
+
+  $alias_generator = \Drupal::getContainer()->get('pathauto.generator');
+  foreach ($storage->loadMultiple($entity_ids) as $entity) {
+    if ($entity->bundle() === 'asset_release' && $entity->field_isr_release_number->isEmpty()) {
+      // There are a number of releases that do not have a release number, even
+      // though it is a mandatory field. Aliases fail to be created for these
+      // entities. These will be handled in ISAICP-6217.
+      // @see https://citnet.tech.ec.europa.eu/CITnet/jira/browse/ISAICP-6217
+      continue;
+    }
```
claudiu-cristea, line 141: Good catch

claudiu-cristea (commented):
Some remarks are inline code.

# Previous QA remarks not addressed

## Wrong non-EN aliases (not fixed)

After running the updates, I've noticed >500 non-English aliases that were wrongly computed. It seems that a token is not resolved. You can find them by running:

```sql
SELECT * FROM path_alias WHERE alias LIKE '/rdf_entity/%'
``` 

## Node aliases should be computed as well (not fixed)

See the inline comment.

web/modules/custom/joinup_core/joinup_core.post_update.php
idimopoulos: :( When you forget to revert test changes..... :(

idimopoulos:
I have removed the calculation of the translation aliases. I left it to be handled in a separate issue. Please, check the parent Jira ticket for reference.

PR #2260, opened 2020-09-23
ISAICP-6146: Licence compatibility check page controller

pfrenssen:


claudiu-cristea (commented):
The ticket description states:

> The controller should only visible to users with the "view compatibility documents" permission. This permission is initially only granted to moderators. It will be extended to all users once the documents are prepared on production.

However I don't see this implemented.

As we've been discussed I've also renamed the plugins for better DX.

Other remarks are inline, in code.

PR #2263, opened 2020-09-25
ISAICP-5775: Present policy domain in the solution header.

idimopoulos:


web/modules/custom/joinup_core/src/Plugin/Field/FieldFormatter/EntityReferenceCustomLinkFormatter.php
```diff
@@ -0,0 +1,235 @@
...
+   */
+  public function __construct(ContainerInterface $container, $plugin_definition, FieldDefinitionInterface $field_definition, array $settings, $label, $view_mode, array $third_party_settings, Token $token) {
+    parent::__construct($container, $plugin_definition, $field_definition, $settings, $label, $view_mode, $third_party_settings);
+    $this->token = $token;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function create(ContainerInterface $container, array $configuration, $plugin_id, $plugin_definition) {
```
claudiu-cristea, line 65: Missing `:self` as return type.

web/modules/custom/joinup_core/src/Plugin/Field/FieldFormatter/EntityReferenceCustomLinkFormatter.php
```diff
@@ -0,0 +1,235 @@
...
+      $configuration['view_mode'],
+      $configuration['third_party_settings'],
+      $container->get('token')
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function defaultSettings() {
```
claudiu-cristea, line 81: Missing return strict type.

web/modules/custom/joinup_core/src/Plugin/Field/FieldFormatter/EntityReferenceCustomLinkFormatter.php
```diff
@@ -0,0 +1,235 @@
...
+      'label' => '[term:label]',
+      'query_parameters' => '',
+      'limit' => -1,
+    ] + parent::defaultSettings();
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function settingsForm(array $form, FormStateInterface $form_state) {
```
claudiu-cristea, line 93: Missing return strict type.

web/modules/custom/joinup_core/src/Plugin/Field/FieldFormatter/EntityReferenceCustomLinkFormatter.php
```diff
@@ -0,0 +1,235 @@
...
+      '#default_value' => $this->getSetting('limit'),
+    ];
+
+    return $elements;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function settingsSummary() {
```
claudiu-cristea, line 134: Missing return strict type.

web/modules/custom/joinup_core/src/Plugin/Field/FieldFormatter/EntityReferenceCustomLinkFormatter.php
```diff
@@ -0,0 +1,235 @@
...
+      '%limit' => $this->getSetting('limit'),
+    ]);
+
+    return $summary;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function viewElements(FieldItemListInterface $items, $langcode) {
```
claudiu-cristea, line 156: Missing return strict type.

web/modules/custom/joinup_core/src/Plugin/Field/FieldFormatter/EntityReferenceCustomLinkFormatter.php
```diff
@@ -0,0 +1,235 @@
...
+      $return[$name] = $this->token->replace($value, [$entity->getEntityTypeId() => $entity]);
+    }
+
+    return $return;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function checkAccess(EntityInterface $entity) {
```
claudiu-cristea, line 223: Missing return strict type.

web/modules/custom/joinup_core/src/Plugin/Field/FieldFormatter/EntityReferenceCustomLinkFormatter.php
```diff
@@ -0,0 +1,235 @@
...
+    return $entity->access('view label', NULL, TRUE);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function isApplicable(FieldDefinitionInterface $field_definition) {
```
claudiu-cristea, line 230: Missing return strict type.

web/modules/custom/joinup_core/src/Plugin/Field/FieldFormatter/EntityReferenceCustomLinkFormatter.php
```diff
@@ -0,0 +1,235 @@
...
+    $limit = $this->getSetting('limit');
+    if ($limit !== -1 && $limit < count($entities_to_view)) {
+      // Due to the fact that delta is not yet implemented in SPARQL, in case
+      // not all terms are shown, sort the terms alphabetically. This will keep
+      // the results predictable and consistent across the website.
+      usort($entities_to_view, function (EntityInterface $entity_a, EntityInterface $entity_b): int {
+        return $entity_a->label() <=> $entity_b->label();
+      });
+      $entities_to_view = array_splice($entities_to_view, 0, (int) $this->getSetting('limit'));
+    }
```
claudiu-cristea, line 168: This part makes this formatter so custom, so "only Joinup thing" that I don't see the reason why it was made so configurable.

Also, conceptually, `$this->getEntitiesToView()` should be in business to perform these operations of limiting and sorting, as its name pretends.

web/modules/custom/joinup_core/src/Plugin/Field/FieldFormatter/EntityReferenceCustomLinkFormatter.php
```diff
@@ -0,0 +1,235 @@
...
+    $return = [];
+    foreach ($query_parameters as $query_parameter) {
+      [$name, $value] = explode('|', $query_parameter, 2);
+      $name = trim($name);
+      $value = rtrim($value, "\r");
+      if (empty($name)) {
+        continue;
+      }
+      $return[$name] = $this->token->replace($value, [$entity->getEntityTypeId() => $entity]);
+    }
```
claudiu-cristea, line 215: I think it's more safe to normalize the blob list in this way:

```php
$query_parameters = array_filter(array_map('trim', explode("\n", str_replace("\r", "\n", $setting)))));
```

web/modules/custom/joinup_core/src/Plugin/Field/FieldFormatter/EntityReferenceCustomLinkFormatter.php
```diff
@@ -0,0 +1,235 @@
...
+  protected function checkAccess(EntityInterface $entity) {
+    return $entity->access('view label', NULL, TRUE);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function isApplicable(FieldDefinitionInterface $field_definition) {
+    // This formatter is only available for taxonomy terms.
+    return $field_definition->getFieldStorageDefinition()->getSetting('target_type') == 'taxonomy_term';
```
claudiu-cristea, line 232: This is also an evidence that the formatter is very specialized. Or do you have in mind other use cases?

web/themes/joinup/templates/rdf_entity/rdf-entity--solution--group-header.html.twig
```diff
@@ -19,6 +19,9 @@
         <div class="stats__row">
           {{ content.downloads }}
         </div>
+        <div class="stats__row">
+          {{ content.field_policy_domain }}
+        </div>
```
claudiu-cristea, line 24: Off topic: I hate this. This is a wrong way to build blocks/pages. We should rely on entity displays and get some help from a Display Suite, Layout Builder or other tool. We should not add content pieces in this way.

web/modules/custom/joinup_core/src/Plugin/Field/FieldFormatter/EntityReferenceCustomLinkFormatter.php
```diff
@@ -0,0 +1,235 @@
...
+        '#url' => $url,
+        '#options' => $url->getOptions(),
+      ];
+    }
+
+    return $elements;
+  }
+
+  /**
+   * Processes the query parameters and encodes the url parameters.
```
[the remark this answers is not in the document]
idimopoulos: I will have to add an entry in the codesniffer for some common keywords :D

web/modules/custom/joinup_core/src/Plugin/Field/FieldFormatter/EntityReferenceCustomLinkFormatter.php
idimopoulos: It is mostly configurable because you guys like things to be configurable in the UI and stored in YAML files instead of hardcoding logic here and there.
Indeed this part was a bit unavoidable because tests would not be consistent (I could not get CPHP to work) which means the resulting limited array would also not be consistent and predictable. 
Of course there was a simpler way to test this by simply checking 7 links, but I thought that the predictability was also an issue.

Apart from that, as we discussed in the chat - adding here for future reference - I understand that this whole formatter is specific, but the formatter itself is mainly allowing the user to use tokens and control the output as link. I have put in place some limitations because they cover our use case now. In the future this can be extended or have a base class which this class will extend. The configurability will come almost out of the box. This for example can be a class that in the future extends a "Custom text formatter" which is not a link but still allows the user to output the field as a custom token supporting text.

web/modules/custom/joinup_core/src/Plugin/Field/FieldFormatter/EntityReferenceCustomLinkFormatter.php
idimopoulos: The same with the relevant answer above.

web/themes/joinup/templates/rdf_entity/rdf-entity--solution--group-header.html.twig
idimopoulos: This was a temporary hack.. To be honest, also for me, It takes me quite some time to understand why some things are not visible in the page even though they exist in the entity display. I have created a follow up ticket for the FE design, you can request the change there. Please, disregard the FE part of the ticket. It already does not show according to the designs.

config/sync/core.entity_view_display.rdf_entity.solution.group_header.yml
```diff
@@ -89,6 +89,19 @@ content:
...
+  field_policy_domain:
+    type: entity_reference_custom_link
+    weight: 4
+    region: content
+    label: inline
+    settings:
+      path: 'internal:/search'
+      label: '[term:name]'
+      query_parameters: "f[0]|policy_domain:[term:tid]\r\nf[1]|type:solution"
+      limit: '7'
```
claudiu-cristea: The `limit` should be an integer. Then, why it was saved as a string? Simply because your formatter settings are missing a schema. This is the reason why this hack [`c1b8b52` (#2263)](https://github.com/ec-europa/joinup-dev/pull/2263/commits/c1b8b524d7b6b30e735a8d04af8052d7b83c8434) was necessary. In fact that was only hiding the real problem.

claudiu-cristea (approved):

PR #2264, opened 2020-09-29
Dye, GrumPHP commit sniffer, die!

claudiu-cristea:


pfrenssen (approved):

pfrenssen:
I agree, it is pointless to do all these time consuming checks on every commit. Having them run on CI is sufficient.

PR #2268, opened 2020-10-04
ISAICP-6220: Hide away the keywords field from the collection page.

idimopoulos:


claudiu-cristea (approved):
Look good and works as expected.

PR #2277, opened 2020-10-08
ISAICP-6239: Releases are shown along the solutions in the federation wizard

idimopoulos:


claudiu-cristea (approved):
Good catch, effective fix!

PR #2282, opened 2020-10-14
ISAICP-6207: Orphaned triples in the graph.

idimopoulos:


config/sync/sparql_entity_storage.mapping.rdf_entity.rdf_graph.yml
```diff
@@ -26,7 +26,7 @@ base_fields_mapping:
       format: t_literal
   created:
     value:
-      predicate: 'http://purl.org/dc/terms/modified'
+      predicate: 'http://purl.org/dc/terms/issued'
```
claudiu-cristea, line 29: Nice catch. However, what happens if, at the time this is released, we already have have `rdf_graph`entities? Maybe we don't have in this moment, but we cannot anticipate at the release time.

web/modules/custom/joinup_core/joinup_core.install
```diff
@@ -76,20 +78,37 @@ function joinup_core_requirements($phase): array {
...
-        'severity' => REQUIREMENT_ERROR,
-        'value' => $error,
-      ];
+      $requirements['joinup_core_forward_revisions']['severity'] = REQUIREMENT_ERROR;
+      $requirements['joinup_core_forward_revisions']['value'] = $error;
+    }
+
+    // Regression requirement to ensure that no orphaned triples are leftover
+    // after various actions that interact with SPARQL storage.
+    $requirements['joinup_core_orphaned_triples'] = [
```
claudiu-cristea, line 97: As we agreed, we'll have only one status report requirement, in `joinup_rdf` module, called “RDF content integrity” (or a better name, if you find). Each type of data integrity violation will be added to this requirement, as a bullet or other categorisation markup. The logic encapsulated in the helper should be also moved in there.

web/modules/custom/solution/tests/src/Kernel/SolutionAffiliationTest.php
```diff
@@ -187,6 +187,43 @@ public function affiliationProvider(): array {
...
+WHERE {
+  GRAPH ?g {
+    ?s ?p <{$solution->id()}> .
+    FILTER NOT EXISTS { ?s a ?type } .
+  }
+}
+QUERY;
+
+    $results = \Drupal::getContainer()->get('sparql.endpoint')->query($query);
+    $this->assertEqual($results[0]->count->getValue(), 0);
```
claudiu-cristea, line 224: The expectation comes first. If the values are of the same type use `assertSame()`, otherwise `assertEquals()`. `assertEqual()` is deprecated.

web/modules/custom/solution/tests/src/Kernel/SolutionAffiliationTest.php
```diff
@@ -187,6 +187,43 @@ public function affiliationProvider(): array {
...
+SELECT COUNT(*) as ?count
+WHERE {
+  GRAPH ?g {
+    ?s ?p <{$solution->id()}> .
+    FILTER NOT EXISTS { ?s a ?type } .
+  }
+}
+QUERY;
+
+    $results = \Drupal::getContainer()->get('sparql.endpoint')->query($query);
```
claudiu-cristea, line 223: In Kernel tests we use `$this->container->get(..)`

claudiu-cristea (changes requested):
Sending back to fix the QA remarks

config/sync/sparql_entity_storage.mapping.rdf_entity.rdf_graph.yml
idimopoulos: I have an update path to fix the values.

web/modules/custom/joinup_rdf/joinup_rdf.install
```diff
@@ -21,34 +22,16 @@ function joinup_rdf_requirements(string $phase): array {
...
-
-  if (!empty($errors)) {
-    $error = t('Leftover graphs were found: :graphs', [
-      ':graphs' => implode(', ', $errors),
-    ]);
-    $requirements['joinup_rdf_adms_validation_graphs']['severity'] = REQUIREMENT_ERROR;
-    $requirements['joinup_rdf_adms_validation_graphs']['value'] = $error;
+  if ($errors = \Drupal::getContainer()->get('joinup_rdf.requirements_helper')->getRequirementErrors()) {
+    $requirements['joinup_rdf_content_integrity']['severity'] = REQUIREMENT_ERROR;
+    $requirements['joinup_rdf_content_integrity']['value'] = implode("\n", $errors);
```
claudiu-cristea, line 34: The output of this concatenation looks very bad.

Status_report___Joinup

It looks like it was never manually tested in the UI 🙃 Why not using real HTML markup? Basic markup, such as `<ul>` and `<li>`, is something that even `drush core:requirements` knows how to handle .

Just setting the `value` to a render array with `'#theme' => 'item_list'` and removing the hardcoded dash should do it.

web/modules/custom/solution/src/SolutionAffiliationFieldItemList.php
```diff
@@ -174,7 +174,9 @@ protected function solutionInOfficialGraph(): bool {
   }
 
   /**
-   * Returns the graph ids that the collection exists in.
+   * Returns the graph ids that the collections exists in.
```
claudiu-cristea, line 177: (if you already touched this)  Yes, acronyms should be uppercased.

web/modules/custom/joinup_core/joinup_core.post_update.php
```diff
@@ -148,10 +148,44 @@ function joinup_core_post_update_0106503(): void {
...
+QUERY;
+
+  \Drupal::getContainer()->get('sparql.endpoint')->query($query);
+}
+
 /**
  * Clean up the validation graphs.
  */
-function joinup_core_post_update_0106504(array &$sandbox): void {
+function joinup_core_post_update_0106506(array &$sandbox): void {
```
claudiu-cristea, line 188: Looks a little bit odd to add the new (post)updates before the ones already committed in develop except the case when one can influence the other. Is this the case?

I noticed this because I've run the all the updates against `develop` and dump the DB in `./tmp/mysql.sql` just o avoid running all path alias updates every time. Because of the changing of the order in this PR, suddenly I saw some updates, such as _Clean up orphaned triples_ not running. The reason was that the `0106505` already ran with `develop`. For this reason I think it's better, as much as we can, to preserve the order that is coming naturally from the development flow.

web/modules/custom/joinup_rdf/src/RequirementsHelper.php
```diff
@@ -0,0 +1,124 @@
...
+  /**
+   * Returns a list of error messages for the joinup_rdf requirements.
+   *
+   * @return array
+   *   An array of string messages.
+   */
+  public function getRequirementErrors(): array {
+    $errors = [];
+    if ($count = $this->getOrphanedTriples()) {
+      $errors[] = $this->translation->formatPlural($count, '- :count orphaned triple was detected.', '- :count orphaned triples were found.', [
```
claudiu-cristea, line 54: Nice that we're using plural formatter,.

claudiu-cristea (changes requested):
Still need some attention.

web/modules/custom/joinup_core/joinup_core.post_update.php
```diff
@@ -148,10 +148,44 @@ function joinup_core_post_update_0106503(): void {
   }
 }
 
+/**
+ * Fix the creation date for the rdf graphs.
```
claudiu-cristea, line 152: s/rdf/RDF

web/modules/custom/joinup_rdf/joinup_rdf.install
idimopoulos: They look much better now :) 

Selection_137
Selection_138

web/modules/custom/joinup_rdf/src/RequirementsHelper.php
```diff
@@ -0,0 +1,134 @@
...
+    $this->translation = $translation;
+  }
+
+  /**
+   * Returns a list of error messages for the joinup_rdf requirements.
+   *
+   * @return array
+   *   An array of string messages.
+   */
+  public function getRequirementErrors() {
```
claudiu-cristea, line 51: We should say that we're returning the list of error messages as a render array. Also the return type is still `: array`

claudiu-cristea (approved):

PR #2288, opened 2020-10-17
ISAICP-6241: RTF order of tiles.

idimopoulos:


pfrenssen:
I don't think this is the right way to solve this random failure. The user scenarios should describe the expected behavior of the front end, not the actual behavior, and definitely not the wrong behavior. According to the specifications solutions should always be shown at the top of the list, before all other content. But now the scenario is describing that community content should be shown first if it is newer than a solution. So it is conflicting with the functional specification. This is like changing a bug into feature.

pfrenssen (approved):

PR #2305, opened 2020-11-02
ISAICP-6099: Featuring content should not affect last update timestamp

pfrenssen:


pfrenssen:
I tried to do the update process in one step as was done in ISAICP-6098, but this was problematic: I was getting errors from the autoloader that the `FeaturedContentInterface` was not found when accessing the Drupal API in a normal update hook. This happens because this interface is defined in a new module, and this module is only enabled _after_ the normal update hook runs. This is a good reminder that we are not allowed to access the Drupal API in a normal update hook, only in post-update hooks.

This also indicates that we got lucky in ISAICP-6098.

web/modules/custom/joinup_core/joinup_core.post_update.php
```diff
@@ -6,3 +6,31 @@
...
+
+    foreach ($storage->loadMultiple($entity_ids) as $entity) {
+      if ($entity instanceof FeaturedContentInterface) {
+        $entity->feature();
+        $count[$entity_type_id]++;
+      }
+    }
+  }
+
+  return 'Featured entities: ' . http_build_query($count, '', ', ');
```
claudiu-cristea, line 35: Hehe, what a usage of `http_build_query()` 👍

tests/features/featured_entities.feature
```diff
@@ -122,16 +122,6 @@ Feature: Featuring content site-wide
...
-    # Due to the bug described in ISAICP-4352, we have to temporary add the next
-    # two lines. Manually testing or running this Behat test locally doesn't
-    # require this cache clear but it fails on the Continuous PHP bot for some
-    # obscure reasons that are very hard to be tracked.
-    # TODO: Remove these two lines in ISAICP-4849.
-    # See https://citnet.tech.ec.europa.eu/CITnet/jira/browse/ISAICP-4352.
-    # See https://citnet.tech.ec.europa.eu/CITnet/jira/browse/ISAICP-4849.
-    And the cache has been cleared
-    And I reload the page
-
```
claudiu-cristea, line 134: ❤️

claudiu-cristea (approved):
Nice work and so happy to see the contextual links cache issue fixed.

web/modules/custom/joinup_core/joinup_core.post_update.php
pfrenssen: I found this trick on StackOverflow :grin:

PR #2318, opened 2020-11-10
ISAICP-6182: Apply comments threading

claudiu-cristea:


config/sync/simple_block.simple_block.anonymous_comment_links.yml
```diff
@@ -0,0 +1,11 @@
...
+langcode: en
+status: true
+dependencies:
+  config:
+    - filter.format.content_editor
+id: anonymous_comment_links
+title: 'Anonymous comment links'
+content:
+  value: '<a href="/caslogin">Login</a> or <a href="/user/register">create an account</a> to comment.'
+  format: content_editor
```
pfrenssen, line 11: The wording and URIs here are subtly different from the ones shown inside the comments themselves.

I'm wondering if it is not overkill to repeat this so often, probably this block is enough and we don't need to show the links for each comment.

edit: This is a functional decision, so I posted this question in the functional ticket.

tests/src/Context/JoinupCommunityContentContext.php
```diff
@@ -260,4 +263,124 @@ public function pinCommunityContentInGroup(AfterNodeCreateScope $scope) {
...
+   * regardless of their indent.
+   *
+   * @param string $comment_number
+   *   The '1 based' number of the comment in the page.
+   * @param string $markup
+   *   The piece of markup to be checked.
+   *
+   * @Then (the )comment #:comment_number should contain the markup :markup
+   */
+  public function assertCommentContainsMarkup(string $comment_number, string $markup): void {
```
pfrenssen, line 280: It is actually possible to type hint the `$comment_number` parameter to the correct type `int`!
Behat has a nice feature that performs type inspection and will cast automatically to the requested type if possible. :+1: 

This can also be done for all other step definitions that are added in this PR that work with integer values.

tests/src/Context/JoinupCommunityContentContext.php
```diff
@@ -260,4 +263,124 @@ public function pinCommunityContentInGroup(AfterNodeCreateScope $scope) {
...
+   * Asserts that a comment has given indent.
+   *
+   * The comment is identified by its '1 based' number. Numbers are starting
+   * from 1 and are assigned in the order comments are shown on the page,
+   * regardless of their indent. The indent is a '0 based' integer.
+   *
+   * @param string $comment_number
+   *   The '1 based' number of the comment in the page.
+   * @param string $indent
+   *   The '0 based' indent value.
```
pfrenssen, line 296: Very good idea to make this very clear in the documentation about the distinction between the 1-based comment numbers and 0-based indentation values. It could get confusing otherwise!

pfrenssen (approved):

pfrenssen:
Note that the test coverage is missing for videos, but this is in line with our existing coverage. Videos are a very niche feature on Joinup, and not well tested. I decided not to reject this PR for this reason.

pfrenssen (approved):

PR #2326, opened 2020-11-20
ISAICP-6154: Switch to permission based check for community content and groups.

idimopoulos:


web/modules/custom/state_machine_permissions/src/StateMachinePermissions.php
```diff
@@ -0,0 +1,105 @@
...
+   * @param string $bundle
+   *   The entity bundle.
+   * @param array $field_names
+   *   An array of state field names.
+   *
+   * @return array
+   *   An array of permissions keyed by the permission id where each entry
+   *   contains the 'label' key and the description.
+   */
+  public function generateWorkflowPermissionsForBundle($entity_type_id, $bundle, array $field_names) {
```
pfrenssen: Since this method is not called anywhere outside of the scope of this class, this can be a protected method. Let's also add full type hint information.

pfrenssen (changes requested):

pfrenssen (approved):

PR #2333, opened 2020-11-30
ISAICP-3913: Remove demo modules from the project.

idimopoulos:


idimopoulos:
I could not find any reference to the demo_content or demo_users in the project. The phing targets that were including these modules are long gone.

pfrenssen (approved):

PR #2334, opened 2020-12-01
ISAICP-6259: Provide digest notifications for solutions.

idimopoulos:


web/modules/custom/joinup_subscription/src/EventSubscriber/CollectionContentSubscriptionSubscriber.php
```diff
@@ -5,57 +5,16 @@
...
-   *   The Joinup message delivery service.
-   * @param \Drupal\Core\Logger\LoggerChannelFactoryInterface $logger
-   *   The logger channel factory.
-   */
-  public function __construct(EntityTypeManagerInterface $entityTypeManager, JoinupMessageDeliveryInterface $joinupMessageDelivery, LoggerChannelFactoryInterface $logger) {
-    $this->entityTypeManager = $entityTypeManager;
-    $this->messageDelivery = $joinupMessageDelivery;
-    $this->loggerFactory = $logger;
-  }
+class CollectionContentSubscriptionSubscriber extends GroupContentSubscriptionSubscriberBase implements EventSubscriberInterface {
```
claudiu-cristea, line 17: I see no dependency of this module on `collection` module. However, I think this subscriber should live in `collection` module. Then we don't need any inter-module dependencies.

tests/src/Context/JoinupSubscriptionContext.php
```diff
@@ -610,56 +617,26 @@ public function assertGroupContentSubscriptionEmailSections(string $username, Ta
...
    * @return array
    *   An array of emails found.
    *
    * @throws \Exception
    *   Thrown if no emails are found or no user exists with the given data.
    */
-  protected function getGroupSubscriptionEmailsByEmail(string $email_address): array {
+  protected function getGroupSubscriptionEmailsByEmail(string $email_address, string $type): array {
     $emails = [];
+    $type = ucfirst($type);
```
claudiu-cristea, line 636: I don't like the `ucfirst()` thing. This is to be avoided especially on multilingual sites. Each language has its rules. Why not just reading the bundle's label?

```php
\Drupal::entityTypeManager()->getStorage('rdf_type')->load($type)->label()
```

web/modules/custom/joinup_subscription/joinup_subscription.module
```diff
@@ -50,26 +50,49 @@ function joinup_subscription_entity_base_field_info(EntityTypeInterface $entity_
...
+    // Set the bundle label as a temporary variable in the notifier as a dynamic
+    // property since it does provide any methods to cary information over.
+    // This will allow us to alter the mail subject and include the group bundle
+    // in it.
+    //
+    // We are accessing information the same way as we do later in the sorting
+    // function so it is safe to assume that values do exist.
+    //
+    // @see \joinup_subscription_mail_alter()
+    $notifier->bundle_label = ucfirst($message->get($fields[$message->getTemplate()->id()])
```
claudiu-cristea, line 90: Same here, better read the standard label than use `ucfirst()`.

web/themes/joinup/joinup_theme.theme
```diff
@@ -1630,6 +1630,20 @@ function joinup_theme_preprocess_rdf_entity(&$variables) {
  */
 function joinup_theme_preprocess_rdf_entity__collection__digest_message_header(&$variables) {
   $variables['url'] = $variables['rdf_entity']->toUrl('canonical', ['absolute' => TRUE])->toString();
+  $variables['bundle_label'] = ucfirst($variables['rdf_entity']->get('rid')->entity->getSingularLabel());
```
claudiu-cristea, line 1633: Same `ucfirst()` thingie.

web/themes/joinup/joinup_theme.theme
```diff
@@ -1630,6 +1630,20 @@ function joinup_theme_preprocess_rdf_entity(&$variables) {
...
+ * Implements hook_preprocess_HOOK().
+ *
+ * Adapts the output of the collection header in collection content subscription
+ * digest messages:
+ * - Changes the link to the solution page to an absolute URL so this can be
+ *   opened in email messages.
+ */
+function joinup_theme_preprocess_rdf_entity__solution__digest_message_header(&$variables) {
+  $variables['url'] = $variables['rdf_entity']->toUrl('canonical', ['absolute' => TRUE])->toString();
+  $variables['bundle_label'] = ucfirst($variables['rdf_entity']->get('rid')->entity->getSingularLabel());
```
claudiu-cristea, line 1646: Same `ucfirst()`.

web/themes/joinup/templates/rdf_entity/rdf-entity--solution--digest-message-header.html.twig
```diff
@@ -0,0 +1,32 @@
...
+        <td height="25" align="left" valign="top">&nbsp;</td>
+        <td align="left" valign="top">&nbsp;</td>
+        <td align="left" valign="top">&nbsp;</td>
+      </tr>
+      </tbody>
+    </table></td>
+</tr>
+<tr>
+  <td align="left" valign="top" class="separator">&nbsp;</td>
+</tr>
```
claudiu-cristea, line 32: Needs an empty newline.

claudiu-cristea (changes requested):
Overall looks good. I still have some points.

web/modules/custom/joinup_subscription/src/EventSubscriber/GroupContentSubscriptionSubscriberBase.php
```diff
@@ -0,0 +1,225 @@
...
+  /**
+   * Returns the entity ID of the group the given entity belongs to.
+   *
+   * @param \Drupal\Core\Entity\ContentEntityInterface $entity
+   *   The entity for which to return the group ID.
+   *
+   * @return string
+   *   The collection or solution ID.
+   */
+  abstract protected function getGroupId(ContentEntityInterface $entity): string;
```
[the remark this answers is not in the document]
idimopoulos: It might be a solution as well so this should cover both entity types.

web/modules/custom/joinup_subscription/joinup_subscription.module
idimopoulos: Here, I am reading the singular label of the bundle type but it comes in all short. I am just making the first character capital.

web/modules/custom/joinup_subscription/src/EventSubscriber/CollectionContentSubscriptionSubscriber.php
idimopoulos: I tried it and it does create a dependency to the notification module. I will move it to joinup_notifications where the rest of notifications lie in too.

pfrenssen (commented):
I have not finished the review, but a huge merge conflict is preventing me from building this. I will continue review after solving the conflict.

web/modules/custom/joinup_subscription/src/EventSubscriber/SolutionContentSubscriptionSubscriber.php
```diff
@@ -0,0 +1,55 @@
...
+use Drupal\Core\Entity\ContentEntityInterface;
+use Drupal\joinup_notification\NotificationEvents;
+use Drupal\joinup_subscription\DigestFormatter;
+use Drupal\solution\Entity\SolutionInterface;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+/**
+ * Subscriber for compiling solution content subscription digest messages .
+ */
+class SolutionContentSubscriptionSubscriber extends GroupContentSubscriptionSubscriberBase implements EventSubscriberInterface {
```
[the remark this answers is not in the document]
pfrenssen: Agreed. These event subscribers are a perfect way to decouple modules. They will only fire if the module that defines the event is enabled.

In this case I think that it is possibly not needed to make separate implementations for solutions and collections. We can leverage bundle classes to wrap the implementation differences between solutions and collections.

pfrenssen (approved):

web/modules/custom/joinup_subscription/joinup_subscription.module
```diff
@@ -111,8 +131,9 @@ function joinup_subscription_mail_alter(array &$message): void {
     $notifier = $message['params']['notifier'] ?? NULL;
     if ($notifier instanceof MessageNotifierInterface) {
       $interval = $notifier->getPluginDefinition()['title'];
-      $message['subject'] = t('Joinup: @interval digest message', [
+      $message['subject'] = t('Joinup: @interval:bundle_label digest message', [
         '@interval' => $interval,
+        ':bundle_label' => ' ' . $notifier->bundle_label ?? '',
```
pfrenssen, line 136: This is suspicious, what is this bundle label that is appended to the title? It looks like this change is intended to differentiate between different digest messages that are sent. Like a "collection digest message" and a "solution digest message".

This is not what we want. There should only be 1 digest sent to a user (for example the "Joinup weekly digest message") and this should include all messages, regardless of the bundle.

web/modules/custom/joinup_subscription/joinup_subscription.module
```diff
@@ -47,29 +50,59 @@ function joinup_subscription_entity_base_field_info(EntityTypeInterface $entity_
...
+    // property since it does not provide any methods to carry information over
+    // either. This will allow us to alter the mail subject and include the
+    // group bundle in it.
+    //
+    // We are accessing information the same way as we do later in the sorting
+    // function so it is safe to assume that values do exist.
+    //
+    // @see \joinup_subscription_mail_alter()
+    $message = reset($messages);
+    $notifier->bundle_label = $message->getSubscribedGroup()->get('rid')->entity->label();
```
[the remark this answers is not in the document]
pfrenssen: The functional requirements have been updated: we no longer send multiple digests but will stick with the original design and will send a single digest containing both the collection and solution content.

pfrenssen (approved):

PR #2347, opened 2020-12-07
ISAICP-6320: Remove deprecated methods in JoinupGroupHelper

pfrenssen:


web/modules/custom/custom_page/src/Plugin/search_api/processor/IncludeCustomPageContent.php
```diff
@@ -111,10 +111,12 @@ public function preprocessIndexItems(array $items): void {
...
-            $custom_page_paragraph_body_list = $paragraph->get('field_body');
-            foreach ($custom_page_paragraph_body_list as $list_item) {
-              if (!$list_item->isEmpty()) {
-                $body_field->addValue(check_markup($list_item->value, $list_item->format));
+            if ($paragraph->hasField('field_body')) {
+              $custom_page_paragraph_body_list = $paragraph->get('field_body');
+              foreach ($custom_page_paragraph_body_list as $list_item) {
+                if (!$list_item->isEmpty()) {
+                  $body_field->addValue(check_markup($list_item->value, $list_item->format));
+                }
```
pfrenssen, line 119: This is out of scope for this ticket, but I fixed this because during testing I got an `InvalidArgumentException: Field 'field_body' is unknown` here. It seems that this code was written when we only had the simple paragraph type. The new paragraph types like accordion etc don't have this field.

idimopoulos (approved):
Nice work :) The bundle classes is a really good improvement.

PR #2354, opened 2021-01-07
ISAICP-5747: Remove instances of the *site_pinned field.

idimopoulos:


tests/features/pin_to_front.feature
```diff
@@ -128,63 +128,6 @@ Feature: Pinning content to the front page
       | discussion   | Discussion |
       | news         | News       |
 
-  Scenario Outline: Moderators can pin and unpin collections and solutions to the front page.
-    When I am an anonymous user
-    And I am on the homepage
-    And I click "<header link>" in the "Header" region
-    Then I should see the following tiles in the correct order:
```
pfrenssen, line 135: Why is this test being removed? As far as I can see this is a refactoring that should not affect current functionality. 

Having collections and solutions pinned to the front page is an important feature for the moderators. We should keep this test to ensure that this works as expected.

Edit: I see that a new ticket ISAICP-6328 has been created to put this test back, but that seems like the wrong way to approach this. Let's keep the test, and block this ticket on ISAICP-6328 instead. If we remove this test now we cannot know for sure that this refactoring does not affect the current functionality.

pfrenssen (changes requested):
We shouldn't remove a test in a refactoring PR.

config/sync/rdf_entity.rdfentity.rdf_graph.yml
```diff
@@ -5,7 +5,7 @@ dependencies: {  }
 name: 'RDF Graph'
 rid: rdf_graph
 description: ''
-label_singular: RDF graph
-label_plural: RDF graphs
+label_singular: 'RDF graph'
+label_plural: 'RDF graphs'
 label_count:
-  default: "1 RDF graph\x03@count RDF graphs"
+  default: !!binary MSBSREYgZ3JhcGgDQGNvdW50IFJERiBncmFwaHM=
```
pfrenssen, line 11: I checked this and this value remains the same but is now represented in binary base64 encoded format because of a non-standard character present in the string.

pfrenssen (approved):

PR #2355, opened 2021-01-07
ISAICP-5806: Deduplicate WorkflowStatePermission plugins

pfrenssen:


pfrenssen:
:green_heart:

pfrenssen:
While working on this I noticed that the workflow state permission plugins were using an obsolete pattern of retrieving group memberships directly from the OG membership manager. We now have `GroupInterface::getMembership()` which is preferable because it binds the group related functionality directly to the group entities, making it unneeded to know the implementation details (i.e. that we use Organic Groups for this).

I had a look around the rest of the code and fixed a bunch of instances where this same outdated pattern was used while waiting for the tests to become green again. This hunt for a green build took a bit longer than expected, so there are quite a few unrelated classes fixed :)

claudiu-cristea (approved):

PR #2368, opened 2021-01-21
ISAICP-6273: Integrate Webtools eTrans service

pfrenssen:


web/modules/custom/joinup_core/joinup_core.module
```diff
@@ -201,3 +211,56 @@ function joinup_core_module_implements_alter(array &$implementations, $hook) {
...
+ * are suitable for machine translation.
+ *
+ * This is handled using a custom access check rather than configuring block
+ * visibility conditions in the block UI since we need to check conditions
+ * across several domains. This is not possible in the UI because the conditions
+ * are compared using "AND" logic.
+ *
+ * @see https://www.drupal.org/project/drupal/issues/923934
+ */
+function joinup_core_block_access(BlockInterface $block, string $operation, AccountInterface $account): AccessResultInterface {
```
claudiu-cristea, line 228: It would have been nice and elegant if the whole logic here was a _visibility condition_.

web/modules/custom/joinup_core/joinup_core.module
```diff
@@ -201,3 +211,56 @@ function joinup_core_module_implements_alter(array &$implementations, $hook) {
...
+  if ($operation !== 'view' || $block->getPluginId() !== 'oe_webtools_etrans') {
+    return AccessResult::neutral();
+  }
+
+  // Check if we are on a route that is suitable for machine translation.
+  $eligible_route_names = [
+    // The "About" page of a collection or solution.
+    'entity.rdf_entity.about_page',
```
claudiu-cristea, line 236: How about the Overview page, which is the canonical and landing page?

config/sync/block.block.webtools_etrans.yml
```diff
@@ -0,0 +1,23 @@
...
+region: content
+weight: -25
+provider: null
+plugin: oe_webtools_etrans
+settings:
+  id: oe_webtools_etrans
+  label: 'OpenEuropa Webtools ETrans'
+  provider: oe_webtools_etrans
+  label_display: '0'
+  render_as: link
```
claudiu-cristea, line 22: In the ticket's wireframes I've noticed that they are requiring a button rather than a link. I had a discussion with Filip and he confirmed.

claudiu-cristea (commented):
At least 2 remarks from my review require a discussion, the "Overview" page question and moving the access logic in a condition plugin. I'm returning the ticket to clarify these aspects.

web/modules/custom/joinup_core/joinup_core.module
pfrenssen: I was in doubt about this. A visibility condition is available to all blocks and cannot be hidden from the UI. Since this is hardcoded logic I wanted to limit it only to this block.

web/modules/custom/joinup_core/joinup_core.module
pfrenssen: The documentation of eTrans says that it is recommended to use it on pages that are content focused, not on pages which are mainly intended to link to other content. For the overview pages I mainly see these linking to solutions and community content.

It is trivial to add it though. I am sure the functional team will come up with some additional routes where they can be added.

config/sync/block.block.webtools_etrans.yml
pfrenssen: I suppose the reason for this is to make sure that search engines don't follow it? I gathered from the mockups that this was intended to be a link styled as a button. I can change it.

web/modules/custom/joinup_core/joinup_core.module
pfrenssen: I discussed this today with @claudiu-cristea and we decided to keep the hook.

web/modules/custom/joinup_core/joinup_core.module
pfrenssen: This has been escalated to the business analyst team. They decided to show the button only on the About page for now, leaving the option open to add this in the future to the collection/solution homepage if users request it.

config/sync/block.block.webtools_etrans.yml
pfrenssen: This was discussed with the business analysts. They also opt for the button. I will change it and update tests.

claudiu-cristea (approved):

PR #2378, opened 2021-01-31
PHP 8 compatibility - round 1

pfrenssen:
This PR does a first round of PHP 8 compatibility fixes, working towards a successful build and installation.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -71,26 +71,26 @@
         },
         {
             "name": "caxy/php-htmldiff",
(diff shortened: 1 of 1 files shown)


pfrenssen:
So this is immediately blocked on the fact that we are still building for PHP 7.1. It's going to be quite impossible to make a build that will support both PHP 8 and PHP 7.1. We need to update to PHP 7.3 or 7.4 first.

PR #2382, opened 2021-02-05
ISAICP-6086: Create a Joinup Acceptance module

claudiu-cristea:


web/modules/custom/acceptance/src/Form/DeliverDigestMessagesForm.php
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildForm(array $form, FormStateInterface $form_state): array {
+    $form['deliver'] = [
+      '#type' => 'submit',
+      '#value' => $this->t('Deliver message digests'),
+      '#submit' => [[$this, 'deliverDigestMessages']],
```
idimopoulos, line 68: Is there a reason to generate an explicit submit handler instead of using the default `submitForm`?

web/modules/custom/acceptance/src/Form/DeliverDigestMessagesForm.php
```diff
@@ -0,0 +1,132 @@
...
+   *   The form state API object.
+   */
+  public function deliverDigestMessages(array &$form, FormStateInterface $form_state): void {
+    $this->expireDigestMessages();
+    do {
+      $this->expireMessageDigestNotifiers();
+      message_digest_cron();
+      $this->processQueue();
+    } while ($this->countAllUndeliveredDigestMessages());
+  }
```
idimopoulos, line 88: There is no indication that a process is completed. We should add a status message. Since the `::countAllUndeliveredDigestMessages()` returns the count of messages to be delivered, we could even show the number of digest messages sent. But even a "Process complete" success message should be fine.

idimopoulos (changes requested):

web/modules/custom/acceptance/src/Form/DeliverDigestMessagesForm.php
claudiu-cristea: Yes. This form is open to support also other _acceptance tasks_ in the future. So, I've anticipated that new buttons or form elements might be added, with a different functionality. Thus, I didn't want to use the form submit for this particular action. Each future button or form element will add their submit logic, separately, in their submit callbacks.

True, that the form & block & permissions are telling about message delivery. That I'll change to reflect the idea of generic _acceptance tasks_ UI.

web/modules/custom/acceptance/src/Form/DeliverDigestMessagesForm.php
claudiu-cristea: OK, then, at least we should also say something on queue error.

PR #2385, opened 2021-02-09
ISAICP-6366: Document how to set up a local development environment

pfrenssen:


idimopoulos (approved):
Since this is working for you, I am accepting it as is.

PR #2388, opened 2021-02-15
ISAICP-6350: Glossary: Highlight only the first instance

claudiu-cristea:


tests/features/glossary/glossary.feature
```diff
@@ -243,3 +243,80 @@ Feature: As a moderator or group facilitator I want to be able to add, edit and
...
+    And I should see the link "Evermore"
+    But I should not see the link "battle"
+    And I should not see the link "BATTLE"
+    And I should not see the link "everMore"
+    And I should not see the link "EVERmore"
+
+    When I click "Settings" in the "Entity actions" region
+    And I uncheck the box "Highlight only the first term appearance"
+    And I press "Save"
+    And I go to the "Collection With Glossary" collection
```
pfrenssen, line 309: After submitting the form it would be better to redirect the user automatically back to the collection homepage. The user can be a bit lost on how to return, and the back button shows the page from the browser cache, which doesn't have any changes applied.

pfrenssen (approved):

tests/features/glossary/glossary.feature
claudiu-cristea: I fully agree.

pfrenssen (approved):

PR #2390, opened 2021-02-16
ISAICP-6313: Hide date range filters when there are no results

pfrenssen:


pfrenssen:
Yeah this is not looking good. I have a working PoC but it is super hacky and not fit for production. It also makes things like licences and user profiles disappear from the search.

PR #2391, opened 2021-02-16
ISAICP-6278: Unify collection & solution subscribe forms

claudiu-cristea:


web/modules/custom/joinup_subscription/src/Form/UnsubscribeFromAllForm.php
```diff
@@ -55,90 +56,82 @@ class UnsubscribeFromAllForm extends ConfirmFormBase {
...
 
   /**
    * {@inheritdoc}
    */
-  public function getCancelUrl() {
-    return Url::fromRoute('joinup_subscription.my_subscriptions', [
-      'subscription_type' => $this->bundle,
-    ]);
+  public function getCancelUrl(): Url {
+    return Url::fromRoute('joinup_subscription.my_subscriptions');
```
idimopoulos, line 106: The only remark I found is this cancel URL. When a moderator performs the unsubscribe-from-all action in the profile of another user, the cancel URL as well as unsubscribing the user from all groups, redirects the user to their own subscription page instead of the user they were performing the action at.

idimopoulos (changes requested):

idimopoulos (approved):

PR #2410, opened 2021-02-26
ISAICP-5958: Update group content URL aliases

claudiu-cristea:


web/modules/custom/joinup_group/tests/src/ExistingSite/GroupContentUpdateUrlAliasTest.php
```diff
@@ -0,0 +1,437 @@
...
+    ]));
+
+    // Edit the solution and add a short ID.
+    $edit = ['field_short_id[0][value]' => 'abc123'];
+    $this->drupalPostForm($this->entity['rdf_entity']['solution']['1']->toUrl('edit-form'), $edit, 'Publish');
+    $this->assertSession()->pageTextContains("You've added a solution short ID: abc123. It will take some time until the solution content URLs will be updated.");
+
+    // Process both queues.
+    $this->drush('queue:run', ['joinup_group:group_update']);
+    $this->drush('queue:run', ['joinup_group:group_content_update']);
```
idimopoulos, line 149: The branch is not working mainly of these two lines, or I simply did not find a way of how to do it in a "normal scenario". Explaining. Normally, what I would expect, is that when I update a group's Short ID, and I see the message that `It will take some time until the <group type> content URLs will be updated`, this will happen in the upcoming cron requests.
**Disclaimer**: I was able to update the aliases by running the `drush queue:run "joinup_group:group_content_update"` command.
What I did in the ticket is that I updated a Short ID and ran multiple times the cron job. There is no cron hook that processes these queues and no URL is updated after multiple runs.
Behat scenario is also msising but it is OK I guess since this is also a test.

idimopoulos (changes requested):

web/modules/custom/joinup_group/tests/src/ExistingSite/GroupContentUpdateUrlAliasTest.php
claudiu-cristea: What you mean by the _branch is not working_?

> There is no cron hook that processes these queues and no URL is updated after multiple runs.

It shouldn't be any cron hook. After processing all cron hooks, cron is processing the active queues. See https://git.drupalcode.org/project/drupal/-/blob/8.9.x/core/lib/Drupal/Core/Cron.php#L145. It might be possible that you have other errors on cron that interrupts cron to finish. I had this case and I bypassed those cron hooks.

PR #2418, opened 2021-03-13
ISAICP-5971: Update the password_policy module.

idimopoulos:


composer.json
```diff
@@ -56,7 +56,7 @@
         "drupal/og_menu": "^1.0-alpha4",
         "drupal/paragraphs": "^1.12",
         "drupal/paragraphs_jquery_ui_accordion": "^1.4",
-        "drupal/password_policy": "3.0-alpha5",
+        "drupal/password_policy": "3.x-dev",
```
pfrenssen, line 59: Let's make sure to stay on a stable version. Password policy is a very brittle subject where a single commit with a small change or addition can break all our tests.

As is mentioned in the Jira ticket we should wait with this until they have released a new version (8.x-3.0-beta2 or higher).

pfrenssen (changes requested):

PR #2423, opened 2021-03-19
ISAICP-6416: Allow an EC namespace to contain a '/'.

idimopoulos:


claudiu-cristea (commented):
I have some, more or less, formal remarks, but I've fixed them in [`d00cf1d` (#2423)](https://github.com/ec-europa/joinup-dev/pull/2423/commits/d00cf1d46de5f1f8f3c665eb82bce890e4ab1942)

web/modules/custom/joinup_core/src/PathProcessor/IdRedirectUuidProcessor.php
```diff
@@ -0,0 +1,38 @@
...
+ * @see \Drupal\joinup_core\Controller\IdRedirect::redirectToRdfEntity()
+ */
+class IdRedirectUuidProcessor implements InboundPathProcessorInterface {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function processInbound($path, Request $request): string {
+    if (strpos($path, '/data/') === 0 && !$request->query->has('uuid')) {
+      preg_match('#/data/(?<namespace>[^/]+)/(?<uuid>.+)#', $path, $matches);
```
claudiu-cristea, line 30: Also, we can make the regex more exact by adding `^` at the beginning an `$` at the end.

PR #2432, opened 2021-03-25
ISAICP-6222: Explain member permissions

pfrenssen:


web/modules/custom/joinup_group/src/Controller/GroupMembershipPermissionsInformationController.php
```diff
@@ -0,0 +1,259 @@
...
+    }
+
+    $build['close'] = [
+      '#type' => 'link',
+      '#title' => $this->t('Got it'),
+      '#url' => Url::fromRoute('entity.rdf_entity.member_overview', [
+        'rdf_entity' => $rdf_entity->id(),
+      ]),
+      '#attributes' => [
+        'class' => [
```
idimopoulos, line 244: According to what we have said in the past, all these should go to the theme, not hardcoded in the controller. The same applies for all classes above with the table cell data.

idimopoulos (changes requested):

idimopoulos (approved):

PR #2449, opened 2021-04-15
ISAICP-6431: Replace "policy domain" with "topic".

idimopoulos:


pfrenssen:
In the "Topic" facet on the advanced search page the topics are not shown, but instead we are seeing the term IDs.

term-ids

I'm seeing this after rebuilding from the production database. I have tried to do a full Solr reindex but this doesn't help.

web/modules/custom/joinup_core/joinup_core.deploy.php
```diff
@@ -16,10 +16,33 @@
 
 use Drupal\asset_distribution\Entity\DownloadEvent;
 
+/**
+ * Backup the field policy domain node field.
+ */
```
pfrenssen, line 21: This is in fact restoring the backup.

pfrenssen:
I found a few files still containing references to policy domains, I assume these are intended to be deleted in a followup?

* `./config/sync/taxonomy.vocabulary.policy_domain.yml`
* `./config/sync/sparql_entity_storage.mapping.taxonomy_term.policy_domain.yml`

This one can probably be fixed in this PR:
* `./web/themes/joinup/templates/field/field--solution-header-policy-domain.html.twig`

pfrenssen (changes requested):

idimopoulos:
> I found a few files still containing references to policy domains, I assume these are intended to be deleted in a followup?
> 
> * `./config/sync/taxonomy.vocabulary.policy_domain.yml`
> * `./config/sync/sparql_entity_storage.mapping.taxonomy_term.policy_domain.yml`
> 
> This one can probably be fixed in this PR:
> 
> * `./web/themes/joinup/templates/field/field--solution-header-policy-domain.html.twig`

I have updated the template. 
For the other two lines, please, check the upstream ticket for more information.

pfrenssen (dismissed):

web/modules/custom/joinup_core/joinup_core.post_update.php
```diff
@@ -14,3 +14,21 @@
...
+
+/**
+ * Backup the field policy domain node field.
+ */
+function joinup_core_post_update_0107000(&$sandbox) {
+  $queries = [
+    'ALTER TABLE `node__field_policy_domain` RENAME TO `node__field_topic_backup`;',
+    'CREATE TABLE `node__field_policy_domain` LIKE `node__field_topic_backup`;',
+    'ALTER TABLE `node_revision__field_policy_domain` RENAME TO `node_revision__field_topic_backup`;',
+    'CREATE TABLE `node_revision__field_policy_domain` LIKE `node_revision__field_topic_backup`;',
```
claudiu-cristea, line 26: I wonder why we didn't use here the database API? E.g.:

```php
\Drupal::database()->schema()->renameTable('node__field_policy_domain', 'node__field_topic_backup');
```

It's safer because we might have a different version of MySQL on other environments but the API unifies the usage.

web/modules/custom/joinup_core/joinup_core.deploy.php
```diff
@@ -16,10 +16,57 @@
 
 use Drupal\asset_distribution\Entity\DownloadEvent;
 
+/**
+ * Restore the field policy domain node field into the new topic field.
+ */
+function joinup_core_deploy_0107000(&$sandbox) {
```
claudiu-cristea, line 22: Normally we should add new functions after existing ones. Also this lacks strict typing for $sandbox and function return.

claudiu-cristea (changes requested):
I was trying to check if we don't have leftovers in the database, so I did a MySQL dump and did a text search against the whole dump. Here are my findings:

* The `batch` table contains occurrences of "policy_domain" and/or "policy-domain". But, looking at each row timestamp, it seems these are very old records. I've learned here https://www.drupal.org/project/drupal/issues/2803061 that such stale rows are never cleaned. Instead of fixing the _policy domain_ occurrences, let's add a `batch.storage` service with `::cleanup()` method in our `joinup_core_cron()`.
* Table `cachetags`: We need to rename `config:block.block.policy_domain_facet`, `config:facets.facet.collection_policy_domain`, `config:facets.facet.policy_domain`, `config:facets.facet.solution_policy_domain`.
* The `key_value` entries for `pathauto_state.taxonomy_term` related to policy domain, that I mentioned in my comment and Slack were not removed.
* Tables `node__field_topic` and `node_revision__field_topic`: let's rename also the `field_policy_domain_target_id` index and the table comment/description.
* Tables `paragraph__field_content_listing` and `paragraph_revision__field_content_listing` there are queries that are using the old IDs.
* The `field_user_professional_domain` was not fixed as IDs conversion.

web/modules/custom/joinup_core/joinup_core.deploy.php
idimopoulos: I know and I understand this. This is the standard procedure. However, the problem comes from the fact that I am renaming the topic table in the last step of post updates. If there was to be another deploy hook to run before mine and attempt to update an entity, it would break the deployment due to a missing field table. That is why I wanted my post update to run last, and the deploy to run first. I will add a comment in the hook.

idimopoulos:
Changes are approved but I cannot press the approve button.

PR #2454, opened 2021-04-21
EPIC-6378: Hierarchical topic select in search. 

idimopoulos:


idimopoulos:
This branch is branched off of ISAICP-6431 as it is blocked on renaming the policy domain to topic.

PR #2484, opened 2021-05-27
ISAICP-6525: Adopt EC Europa QA conventions

pfrenssen:


pfrenssen:
This is bypassing GrumPHP. As we experienced already in the past the way GrumPHP sets up a dynamic configuration of PHP_CodeSniffer is not compatible with our ruleset. The same problems we saw in the past with the inclusion and exclusion of filename patterns is still persisting.

The second tool that GrumPHP is running is PHPMD which has no added value for us and only causes false positives.

Both are disabled now, and we have a green pipeline.

PR #2487, opened 2021-06-01
ISAICP-6532: Adhere to QA coding standard rules

pfrenssen:


web/modules/custom/acceptance/joinup_acceptance.info.yml
```diff
@@ -1,8 +1,10 @@
 name: 'Joinup Acceptance'
-type: module
 description: 'Tasks to performed in the acceptance environment.'
-core_version_requirement: ^8 || ^9
+type: module
 package: Joinup
+core_version_requirement: ^8.9 || ^9
```
claudiu-cristea, line 5: Drupal 9.0 is EOL this months. I think it's safe to use `^8.9 || ^9.1`.

web/modules/custom/acceptance/joinup_acceptance.info.yml
```diff
@@ -1,8 +1,10 @@
...
+php: 7.1
```
claudiu-cristea, line 6: I don't think we should add this. We should rely on Drupal core PHP minimum version. An exception could be the case when we need in a module version higher than the core min version. E.g. Drupal 8.9 works with PHP 7.1 but the module needs 7.3 in order to work. I guess we don't have such case.

web/modules/custom/joinup_user/joinup_user.install
```diff
@@ -17,7 +17,25 @@ use Drupal\joinup_user\JoinupUserViewsData;
...
+  $internal_path = "public://default_images";
+  if (is_file($file_path) && \Drupal::service('file_system')->prepareDirectory($internal_path, FileSystemInterface::CREATE_DIRECTORY | FileSystemInterface::MODIFY_PERMISSIONS)) {
+    $file_system = \Drupal::service('file_system');
+    $destination = file_build_uri($file_system->basename($file_path));
+    $file_path = $file_system->copy($file_path, $destination, FileSystemInterface::EXISTS_REPLACE);
+  }
+  /** @var \Drupal\file\FileInterface $file */
+  $file = File::create(['uri' => $file_path]);
+
+  // The uuid is the one assigned to the user photo field so there is no need
```
claudiu-cristea, line 34: s/uuid/UUID. See https://www.drupal.org/drupalorg/style-guide/content#abbreviations

web/modules/custom/acceptance/joinup_acceptance.info.yml
pfrenssen: It is a QA coding standard rule that this PHP version has to be defined in the info file, so adding it is in scope of this PR. I agree that it is a bit stupid that we define the default value, but this is maybe something we should propose for the QA coding standards? For me personally it doesn't matter if this line is there or not.

We need minimum PHP 7.1 because we use nullable types and return types. We will for sure adopt 7.4 functionality as soon as we can :D

PR #2491, opened 2021-06-04
ISAICP-6530: Clean orphan Solr document entries

claudiu-cristea:


pfrenssen:
Hmm I am getting the following error:

```
GuzzleHttp\Exception\ClientException: Client error: `GET http://localhost:8983/solr/drupal_published/update?stream.body=%3Cdelete%3E%3Cquery%3E-hash%3Atb6juv%3C%2Fquery%3E%3C%2Fdelete%3E&commit=true&wt=json` resulted in a `400 Bad Request`
```

I will try to see if this is perhaps related to my Solr version.

pfrenssen:
I cannot test this easily on my system. I am using Solr 7 which works fine for standard usage in Joinup, but throws the error I mentioned above. I tried installing Solr 6 but this is not compatible any more with the version of OpenJDK that I am using. Strangely enough I am getting the error "Your current version of Java is too old to run this version of Solr", but this is probably just a wrong error message.

Unless absolutely necessary I would like to avoid downgrading Java since a lot of software I use (like PHPStorm) uses it. Maybe we can run this in the UAT environment after hours?

idimopoulos (approved):

PR #2493, opened 2021-06-08
ISAICP-6539: Use POST instead of GET with the Solr update endpoint

claudiu-cristea:


pfrenssen (approved):
The `solr:empty` task now runs reliably on Solr 7 (tested with 7.7.3). I couldn't get the `restore_solr.sh` script to work locally. I got an error `org.apache.solr.common.SolrException: Exception while restoring the backup index`, but I can see on the UAT environment that the command works normally there. Approving.

PR #2506, opened 2021-06-24
ISAICP-6548: Switch to OE Bootstrap as the base theme

pfrenssen:


composer.json
```diff
@@ -360,6 +361,9 @@
             "drupal/views_autosubmit": {
                 "The views plugin options are missing schema @see https://www.drupal.org/i/3195213": "https://git.drupalcode.org/project/views_autosubmit/-/merge_requests/2.diff"
             },
+            "openeuropa/oe_bootstrap_theme": {
+                "Preserve link attributes in navigation menus @see https://github.com/openeuropa/oe_bootstrap_theme/pull/38": "https://patch-diff.githubusercontent.com/raw/openeuropa/oe_bootstrap_theme/pull/38.diff"
+            },
```
pfrenssen, line 366: Note that this should not be reviewed. This patch will not be accepted because the navigation template is currently being refactored by the FE team.

This was discussed and we will keep the patch for now so this PR is not blocked. Once their work is completed we will be notified. We can then remove this patch and do a new full test and provide feedback whether it is working.

PR #2509, opened 2021-06-28
ISAICP-6433: Change notification when saving content as draft.

saidatom:


claudiu-cristea (changes requested):
Your first Joinup PR look promising. Good! However, I have few remarks.

web/modules/custom/joinup_workflow/joinup_workflow.module
```diff
@@ -246,3 +266,26 @@ function joinup_workflow_is_motivation_required($entity_type_id, $workflow_id, $
...
+  if ($node->get('field_state')->getString() == 'draft') {
+    \Drupal::messenger()->deleteByType('status');
+    $args = [
+      '@type' => node_get_type_label($node),
+      '%title' => $node->getTitle(),
+      '%state' => $node->get('field_state')->getString(),
+      ':user' => Url::fromRoute('user.page')->toString(),
+      ':url' => $node->toUrl('canonical')->toString(),
+    ];
+    \Drupal::messenger()->addStatus(t('@type <a href=":url">%title</a> has been created as %state. You can find it in the section "My unpublished content" located in your <a href=":user">My account page</a>, or in the aforementioned section under the Collection/Solution it was created in.', $args));
```
claudiu-cristea, line 289: I think the request intent was to have either _Collection_ or _Solution_, but not _Collection/Solution_. I think, but I am not sure, that you can get that with:

```
$node->get('oe_audience')->entity->get('rid')->entity->label()
```

web/modules/custom/joinup_workflow/joinup_workflow.module
```diff
@@ -246,3 +266,26 @@ function joinup_workflow_is_motivation_required($entity_type_id, $workflow_id, $
...
+ * @param \Drupal\Core\Form\FormStateInterface $form_state
+ *   The form state object.
+ */
+function joinup_workflow_message_form_submit_alter(array &$form, FormStateInterface $form_state) {
+  $node = $form_state->getFormObject()->getEntity();
+  if ($node->get('field_state')->getString() == 'draft') {
+    \Drupal::messenger()->deleteByType('status');
+    $args = [
+      '@type' => node_get_type_label($node),
+      '%title' => $node->getTitle(),
```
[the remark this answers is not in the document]
saidatom: Instead of use $node->getTitle() i use $node->toLink()->toString() because generates the HTML link to an entity and i can remove the argument ':url'.

web/modules/custom/joinup_workflow/joinup_workflow.module
```diff
@@ -267,6 +250,13 @@ function joinup_workflow_is_motivation_required($entity_type_id, $workflow_id, $
   return !empty($required_transitions[$entity_type_id][$workflow_id]) && in_array($transition_id, $required_transitions[$entity_type_id][$workflow_id]);
 }
 
+/**
+ * Implements hook_form_node_form_alter().
```
claudiu-cristea, line 254: The correct docs is `Implements hook_form_BASE_FORM_ID_alter()`

web/modules/custom/joinup_workflow/joinup_workflow.module
```diff
@@ -277,15 +267,29 @@ function joinup_workflow_is_motivation_required($entity_type_id, $workflow_id, $
...
-      '%title' => $node->getTitle(),
-      '%state' => $node->get('field_state')->getString(),
-      ':user' => Url::fromRoute('user.page')->toString(),
-      ':url' => $node->toUrl('canonical')->toString(),
-    ];
-    \Drupal::messenger()->addStatus(t('@type <a href=":url">%title</a> has been created as %state. You can find it in the section "My unpublished content" located in your <a href=":user">My account page</a>, or in the aforementioned section under the Collection/Solution it was created in.', $args));
+
+  if ($node instanceof CommunityContentInterface) {
+    $messenger = \Drupal::messenger();
+    if ($node->get('field_state')->value == 'draft') {
```
claudiu-cristea, line 273: Let's use strict comparison (===)

web/modules/custom/joinup_workflow/joinup_workflow.module
```diff
@@ -277,15 +267,29 @@ function joinup_workflow_is_motivation_required($entity_type_id, $workflow_id, $
...
+        '@type' => node_get_type_label($node),
+        '%title' => $node->toLink()->toString(),
+        '%state' => $node->get('field_state')->getString(),
+        ':user' => Url::fromRoute('user.page')->toString(),
+        '@audience' => $node->get('og_audience')->entity->get('rid')->entity->label(),
+      ];
+
+      foreach ($status_messages as $message) {
+        if ($message instanceof MarkupInterface) {
+          if ((string) $message != (string) t('@type %title has been created.', $args)) {
```
claudiu-cristea, line 286: As we're not using strict comparison, I think we can drop the string casting at least on one side.

claudiu-cristea (changes requested):
Looks good. However, I see the test are failing and I've managed to reproduce the failure locally but didn't investigate. I also have some nitpicks, see my inline comments.

web/modules/custom/joinup_workflow/joinup_workflow.module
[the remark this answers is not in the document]
claudiu-cristea: @saidatom initially I was reluctant to your solution because tokens such as `%title` used to be sanitized and HTML markup ends by being escaped. However, after testing manually I was surprised to find that it just works. I've tracked down just to get an explanation, and, yes, the fact the secret is that `$node->toLink()->toString()` doesn't actually return a string, but a `GeneratedLink` instance and that implements `MarkupInterface`. That means in this line https://git.drupalcode.org/project/drupal/-/blob/9.2.x/core/lib/Drupal/Component/Render/FormattableMarkup.php#L260 the HTML is not escaped but preserved. Nice!

web/modules/custom/joinup_workflow/joinup_workflow.module
```diff
@@ -246,3 +249,47 @@ function joinup_workflow_is_motivation_required($entity_type_id, $workflow_id, $
...
+
+/**
+ * Change notification when saving content as draft.
+ *
+ * @param array $form
+ *   The form array.
+ * @param \Drupal\Core\Form\FormStateInterface $form_state
+ *   The form state object.
+ */
+function joinup_workflow_message_form_submit_alter(array &$form, FormStateInterface $form_state) {
```
claudiu-cristea, line 268: Let's add return typing

web/modules/custom/joinup_workflow/joinup_workflow.module
```diff
@@ -246,3 +249,47 @@ function joinup_workflow_is_motivation_required($entity_type_id, $workflow_id, $
 
   return !empty($required_transitions[$entity_type_id][$workflow_id]) && in_array($transition_id, $required_transitions[$entity_type_id][$workflow_id]);
 }
+
+/**
+ * Implements hook_form_BASE_FORM_ID_alter().
+ */
+function joinup_workflow_form_node_form_alter(&$form, FormStateInterface $form_state, $form_id) {
```
claudiu-cristea, line 256: Missing return typing and argument typing.

claudiu-cristea (approved):
Great! Looks good, approved

PR #2510, opened 2021-07-01
ISAICP-6479: Add a "Explore" block to homepage.

saidatom:


web/modules/custom/joinup_front_page/src/Plugin/Block/ExploreItemBlock.php
```diff
@@ -0,0 +1,203 @@
...
+   * @param mixed $plugin_definition
+   *   The plugin implementation definition.
+   * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entity_manager
+   *   The entity manager.
+   * @param \Drupal\Core\Messenger\MessengerInterface $messenger
+   *   The messenger.
+   */
+  public function __construct(array $configuration, $plugin_id, $plugin_definition, EntityTypeManagerInterface $entity_manager, MessengerInterface $messenger) {
+    parent::__construct($configuration, $plugin_id, $plugin_definition);
+    $this->entityManager = $entity_manager;
```
pfrenssen, line 64: In Drupal 8 the `EntityManager` is deprecated, and is replaced with `EntityTypeManager`. In Drupal 9 the `EntityManager` is completely removed. Since we are going to move to D9 soon it's a good idea to already adopt its practices where possible. This will minimize the number of instances where we will have to replace deprecated terminology.

web/modules/custom/joinup_front_page/src/Plugin/Block/ExploreItemBlock.php
```diff
@@ -0,0 +1,203 @@
...
+   * @param string $type
+   *   The type of content.
+   *
+   * @return array
+   *   List of content with view mode "explore_item".
+   *
+   * @throws \Drupal\Component\Plugin\Exception\InvalidPluginDefinitionException
+   * @throws \Drupal\Component\Plugin\Exception\PluginNotFoundException
+   */
+  private function getContent($type): array {
```
pfrenssen, line 151: In Joinup we are using the Drupal philosophy that every class should be reusable and extendable so we avoid the use of private methods. This can be a protected one. Also let's add the `string $type` type hint.

web/modules/custom/joinup_front_page/src/Plugin/Block/ExploreItemBlock.php
```diff
@@ -0,0 +1,203 @@
...
+    $content = [];
+    $entity = $this->entityManager->getStorage('node');
+    $view_builder = $this->entityManager->getViewBuilder('node');
+    $query = $entity->getQuery();
+
+    $ids = $query->condition('status', 1)
+      ->condition('type', $type)
+      ->sort('created')
+      ->pager(12)
```
pfrenssen, line 160: It's better to use `->range(0, 12)` rather than `->pager(12)`. This has less overhead and doesn't risk to pollute the query arguments with pager information.

web/modules/custom/joinup_front_page/src/Plugin/Block/ExploreItemBlock.php
```diff
@@ -0,0 +1,203 @@
...
+   */
+  private function getRdfEntity($type): array {
+    $rdf = [];
+    $entity = $this->entityManager->getStorage('rdf_entity');
+    $view_builder = $this->entityManager->getViewBuilder('rdf_entity');
+    $query = $entity->getQuery();
+
+    $ids = $query->condition('rid', $type)
+      ->sort('created')
+      ->pager(12)
```
pfrenssen, line 191: Here also, better to use `->range(0, 12)` because using a pager will needlessly keep track of the pager state in the `PagerManager`.

web/modules/custom/joinup_front_page/src/Plugin/Block/ExploreItemBlock.php
```diff
@@ -0,0 +1,203 @@
...
+      ->execute();
+
+    $entities = $entity->loadMultiple($ids);
+
+    foreach ($entities as $node) {
+      $rdf[] = $view_builder->view($node, $this->viewMode);
+    }
```
pfrenssen, line 198: Nitpick: since these are RDF entities, these are not nodes ;)

web/modules/custom/joinup_front_page/src/Plugin/Block/ExploreItemBlock.php
```diff
@@ -0,0 +1,203 @@
...
+   * @param string $plugin_id
+   *   The plugin_id for the plugin instance.
+   * @param mixed $plugin_definition
+   *   The plugin implementation definition.
+   * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entity_manager
+   *   The entity manager.
+   * @param \Drupal\Core\Messenger\MessengerInterface $messenger
+   *   The messenger.
+   */
+  public function __construct(array $configuration, $plugin_id, $plugin_definition, EntityTypeManagerInterface $entity_manager, MessengerInterface $messenger) {
```
pfrenssen, line 62: Here the `$plugin_id` can be type hinted as a string.

web/modules/custom/joinup_front_page/src/Plugin/Block/ExploreItemBlock.php
```diff
@@ -0,0 +1,203 @@
...
+  public function getCacheTags() {
+    $tags = [
+      'node:news',
+      'node:events',
+      'rdf_entity:solution',
+      'rdf_entity:collection',
+    ];
+
+    return Cache::mergeTags(parent::getCacheTags(), $tags);
+  }
```
pfrenssen, line 137: I understand the reasoning behind this but it seems dangerous for a block that is shown on the front page. Having this vary by the list cache tags of 4 of our most popular content types means that the front page will be invalidated every single time a change is made to any of these types. I can imagine that this will at times cause the cache to be invalidated multiple times per minute.

Let's keep this for now, but we need to discuss this with the business analyst team to find a good solution, for example a simple time based cache. I will make a comment in the Jira ticket to get some feedback.

If we end up switching to a time based cache this will become very interesting to test :sweat_smile:

tests/features/homepage.feature
```diff
@@ -188,3 +188,39 @@ Feature: Homepage
...
+      | Easier transformations      | Ecosystem                | Clash of jonnys  | EU and European Policies                   | validated | 2021-03-29T10:18:00Z | Archaeological evidence shows that by 10000 BCE all societies employed ecologically transformative land use practices including burning hunting species propagation domestication cultivation have left long-term legacies across the biosphere.     |
+    And event content:
+      | title                        | body               | collection      | solution               | state            |
+      | Thick-target collisional     | Evaporation        | Clash of jonnys | Lantern International  | validated        |
+      | Conductive cooling losses    | Single-loop        | Clash of jonnys | Proton Global          | validated        |
+      | Source of SXR plasma supply  | Fast electrons     | Clash of jonnys | Shiny Shan             | validated        |
+      | Stars forming in material    | Colossal material  | Clash of jonnys | Shiny Shan             | validated        |
```
pfrenssen, line 217: We should probably have 12+ of each type, and also mix in some that are in various states in addition to `validation` (eg `draft` or `proposed`).

This way we can check that our filters work correctly and that only published content ends up on the homepage.

tests/features/homepage.feature
```diff
@@ -188,3 +188,39 @@ Feature: Homepage
...
+    When I am on the homepage
+    Then I should see the button "Solutions" in the "Explore block"
+    And I should see the button "Collections" in the "Explore block"
+    And I should see the button "News" in the "Explore block"
+    And I should see the button "Events" in the "Explore block"
+    And I should see the link "See more" in the "Explore block"
+    When I click "See more"
+    Then I should be on the advanced search page
+    And the "Solutions" content checkbox item should be selected
```
pfrenssen, line 226: Let's extend these tests to make sure the correct content is showing up in the right format: the title, shortened body text, publication date (in the right date format) and content type.

A good way to handle this without having to write an insanely long and complex is to do a spot check of 1 or 2 items per content type. Also check that some content in unpublished state is not showing up unexpectedly.

See for example `assertInTheSpotlight()` for a Behat step definition that does an in-depth check of every rendered element of the "In The Spotlight" block.

pfrenssen (changes requested):
This is great work! Review remarks below.

pfrenssen:
This is looking good. There is a failure but this is caused by ISAICP-6440. I have fixed it.

We still need a decision on how often to update the block. I have asked the functional team about this. If we do not get an answer in time, let's just put in a max age of 24 hours to replace the cache tags.

PR #2513, opened 2021-07-05
ISAICP-6521: Longer glossary terms conflict with substrings.

idimopoulos:


pfrenssen (changes requested):
The test doesn't seem to prove the bug. It is passing also if I revert the fix.

PR #2518, opened 2021-07-09
ISAICP-6414: Users shouldn't be able to edit/delete content after being blocked or removed from a collection/solution.

saidatom:


pfrenssen:
We have to enable the `node_access_strict` flag in OG in this ticket so that users with global permissions to edit their own content can not bypass the newly introduced restriction on editing their own content in groups where they are blocked.

Because of this we will need to duplicate a number of permissions from the global Drupal roles to the OG roles. For example a permission such as `'transition own node document from validated to validated using the community_content - node:post_moderated workflow'` now also needs to be granted on group level.

Since these permissions are currently being reworked in ISAICP-6312, I would propose to pause the work on this ticket for a while until that ticket is in, since this will otherwise conflict with the work done in that ticket.

saidatom:
Looks good. Approved.

PR #2519, opened 2021-07-09
ISAICP-6448: Sort search results by last updated date, when blank search query is performed.

saidatom:


web/modules/custom/joinup_search/joinup_search.module
```diff
@@ -395,3 +396,25 @@ function joinup_search_views_pre_execute(ViewExecutable $view) {
...
+  if (empty($user_input['keys'])) {
+    unset($form['sort_by']['#options']['relevance']);
+
+    if ($user_input['sort_by'] === 'relevance') {
+      $user_input['sort_by'] = 'last-updated-date';
+    }
+    $form_state->setUserInput($user_input);
+  }
+
+  return $form;
```
pfrenssen, line 419: In a form alter hook the `$form` is passed by reference, so it is not needed to return it.

web/modules/custom/joinup_search/joinup_search.module
```diff
@@ -395,3 +396,25 @@ function joinup_search_views_pre_execute(ViewExecutable $view) {
   // Selecting the entity_created option itself won't affect the query.
   $view->getQuery()->sort('entity_created', 'DESC');
 }
+
+/**
+ * Implements hook_form_FORM_ID_alter().
+ */
+function joinup_search_form_views_exposed_form_alter(array &$form, FormStateInterface $form_state, string $form_id) {
+  $view = $form_state->getStorage('view');
```
pfrenssen, line 404: This method doesn't take an argument, it can just be `->getStorage()`. I saw this during review because PHPStorm highlights this when you enable the inspection "Parameters number mismatch declaration".

pfrenssen (changes requested):

web/modules/custom/joinup_search/joinup_search.module
```diff
@@ -395,3 +396,25 @@ function joinup_search_views_pre_execute(ViewExecutable $view) {
...
+  if ($view['view']->id() !== 'search') {
+    return;
+  }
+
+  $user_input = $form_state->getUserInput();
+  if (empty($user_input['keys'])) {
+    unset($form['sort_by']['#options']['relevance']);
```
pfrenssen, line 411: This removes the "Sort by relevance" entirely from the dropdown. This has an unwanted side effect, since if a user clicks on "Advanced search" with the intention to search on keywords, then the default option (and the one that makes the most sense) is not available to them.

When a user arrives on the search page, then the options should still include "Relevance" and this should still be set as the default option, so that when a user enters their keywords they will get relevant search results.

The expectation of the user is that they will get relevant results after entering their keywords. As it is implemented now, the user will need to take a bunch of additional steps: enter keywords, click search button, then when the results are shown they have to choose the "Relevance" option from the dropdown and launch the search again.

So I think the idea of this ticket is that the _results_ should be shown in last updated order when users arrive on this page, but the actual form should remain untouched.

pfrenssen (approved):

PR #2528, opened 2021-07-19
ISAICP-6468: Content listing paragraph is added w/empty presets or content filter.

saidatom:


saidatom:
Looks good. Approved.

PR #2538, opened 2021-08-04
ISAICP-6632: Update the EIRA_SKOS file.

saidatom:


pfrenssen (changes requested):
There is a problem with the update - there are solutions currently in Joinup that are using deprecated EIRA terms which have been removed in the new version. After running the update these solutions have lost their "Solution type" which means the data model is not complete.

We should map these obsolete EIRA terms to new ones if a replacement has been suggested, but in the case no replacement is available it is not clear how to proceed. Probably we need to make an inventory of solutions using deprecated terms and contact them / update them manually.

PR #2540, opened 2021-08-09
ISAICP-6638: Fix notice about missing `menu_name` theme variable

pfrenssen:


pfrenssen:
This turned out to be caused by a core bug in the Toolbar module: [Issue #3227370: Toolbar menu theme override omits the 'menu_name' variable](https://www.drupal.org/project/drupal/issues/3227370).

The reason why this was not affecting us before is because the new theme provides an overridden template for the toolbar menu: `menu--toolbar.html.twig`. By providing this template we are switching the theme rendering engine from "theme_engine" to "module", and the variables are now sourced from `toolbar_theme()` where previously they were sourced from `drupal_common_theme()`. Unfortunately `toolbar_theme()` omits declaring the `menu_name` variable. It is fixed in the upstream issue.

saidatom (approved):

PR #2542, opened 2021-08-10
ISAICP-6579: Facets summary Search view.

saidatom:


config/sync/core.entity_view_display.entity_subqueue.discover_topics.default.yml
```diff
@@ -26,7 +26,7 @@ content:
       field_formatter_range:
         limit: 0
         offset: 0
-        reverse: false
+        order: 0
```
idimopoulos, line 29: In the upstream version of field_formatter_range, between version 1.1 and 1.2, there was a BC change that caused issues with our configuration. The change is that they removed the property reverse (boolean) and replaced it with the property order (integer).

tests/features/joinup_search/group_search.feature
```diff
@@ -221,7 +221,7 @@ Feature: Search inside groups
     When I enter "ground" in the search bar and press enter
     Then the option with text "Chalet construction   (2)" from select facet "collection/solution" is selected
     And the page should show the tiles "Ground plan, Natural materials"
-    And the page should show the following chip:
+    And the page should show the following chip in the "Search bar":
```
idimopoulos, line 224: The errors were caused because there was styling applied in the ticket. The chip of the group now appeared twice in the page, one in the search bar, as it is meant, and one in the new chips. Probably one should be removed, however, since the tests are asserting for the chips in the search bar, I made it a bit more specific.

saidatom:
Looks good. Approved.

PR #2550, opened 2021-08-17
ISAICP-6652: Fix 'See more topics' modal doesn't open on Chrome.

saidatom:


web/modules/custom/joinup_front_page/joinup_front_page.module
```diff
@@ -12,8 +12,21 @@ declare(strict_types = 1);
  */
 function joinup_front_page_preprocess_block(&$variables): void {
   if (\Drupal::currentUser()->isAuthenticated()) {
-    // The libraries are available for authenticated users.
-    return;
+    // The libraries are available for all users except the
+    // ones only with authenticated role.
```
pfrenssen, line 16: This documentation is not clear, it assumes the reader knows exactly what this is about. Let's update this so we have a good explanation why we need to alter a certain block.

web/modules/custom/joinup_front_page/joinup_front_page.module
```diff
@@ -12,8 +12,21 @@ declare(strict_types = 1);
...
+      'administrator',
+      'rdf_graph_manager',
+    ];
+
+    $user_roles = \Drupal::currentUser()->getRoles();
+    foreach ($roles as $role) {
+      if (in_array($role, $user_roles)) {
+        return;
+      }
+    }
```
pfrenssen, line 29: Is it possible to skip this whole role filtering? Is this perhaps needed to work around some bug or conflict?

It seems to me that all users regardless of role should be able to interact with the modal, the Discover Topics modal is intended for every user on Joinup.

web/modules/custom/joinup_front_page/joinup_front_page.module
pfrenssen: Here is a proposal for updated documentation:

```
/**
 * Implements hook_preprocess_HOOK().
 *
 * The 'Discover topics' block which is shown on the front page is created using
 * the Views module. The block includes a 'See more topics' link which opens a
 * modal dialog with the full list of available topics. Include the necessary JS
 * so that all users can interact with the modal.
 */
```

web/modules/custom/joinup_front_page/joinup_front_page.module
saidatom: After several tests with different user permissions we do not need this rule.

pfrenssen (approved):

PR #2552, opened 2021-08-18
ISAICP-6593: Theme "In The Spotlight" block.

gingapa:


web/themes/ventuno/src/scss/components/_homepage.scss
```diff
@@ -0,0 +1,77 @@
+// Temporarily hide homepage title
+#block-ventuno-pagetitle {
+  display: none;
+}
+
```
pfrenssen: Is it OK for you if I hide the pagetitle block on the homepage? Then we don't need this workaround.

pfrenssen (approved):

web/themes/ventuno/src/scss/components/_homepage.scss
gingapa: Yes perfect!

PR #2554, opened 2021-08-18
ISAICP-6598: Style HIGHLIGHTED SOLUTION block.

gingapa:


pfrenssen:
The solution type is not styled, this should be in a muted grey according to the design. The date which is in the mockup does not exist by the way, solutions are intended to be eternal :)

pfrenssen (approved):

PR #2571, opened 2021-08-26
ISAICP-6584: Integrate Facets Form & Date Range Widget.

saidatom:


tests/src/Context/JoinupSearchContext.php
```diff
@@ -374,6 +462,47 @@ public function checkCheckboxFacet(string $option, string $facet_type): void {
...
+        $node_element->click();
+        return;
+      }
+    }
+
+    throw new \Exception("The option '{$option}' was not found in the '{$facet_type}' facet.");
+  }
+
+  /**
+   * Facets form inputs.
```
idimopoulos, line 493: Please, add a better description here.

tests/src/Context/JoinupSearchContext.php
```diff
@@ -374,6 +462,47 @@ public function checkCheckboxFacet(string $option, string $facet_type): void {
...
+   *
+   * @param string $name
+   *   The label of the input.
```
idimopoulos, line 496: Also, input is not really a good word for action buttons. I mean, I had to look at the method to understand what it does.

tests/src/Traits/TraversingTrait.php
```diff
@@ -350,6 +350,40 @@ protected function findFacetByAlias(string $alias, ?NodeElement $region = NULL,
...
+   *   The facet node element.
+   *
+   * @throws \Exception
+   *   Thrown when the facet form is not found in the designated area.
+   */
+  protected function findFacetFormByAlias(string $alias, ?NodeElement $region = NULL, string $html_tag = '*'): NodeElement {
+    if ($region === NULL) {
+      $region = $this->getSession()->getPage();
+    }
+    $facet_id = self::getFacetFormIdFromAlias($alias);
```
idimopoulos, line 377: To be honest I don't like the fact that we are duplicating the documentation as well as most of the method in order to look in a different list with a different data-drupal property.
I would prefer to somehow unify them. But again, this is also optional. It won't be too much of an improvement.

idimopoulos: Actually, there is a good way to improve it. I see that the `getFacetFormIdFromAlias` is used only here, and the `getFacetIdFromAlias` is used only in 3 places. Furthermore, in all cases, the ID is retrieved to generate an xpath with the corresponding data-drupal property.
We can unify the above with a parameter of the type of the facet (e.g. facet or facet-form) and have it return the xpath (including the data-drupal property) directly instead of simply returning the identifier.
I mean, this is just one idea, but we can find ways to unify them.

idimopoulos: If the ticket is way over board, you can ignore this bit.

web/modules/custom/search_api_field/src/Plugin/facets/facet_source/SearchApiFieldDeriver.php
```diff
@@ -61,6 +61,7 @@ public static function create(ContainerInterface $container, $base_plugin_id) {
    * {@inheritdoc}
    */
   public function getDerivativeDefinitions($base_plugin_definition) {
+    $plugin_derivatives = [];
```
idimopoulos, line 64: This should be defined in its scope, meaning outside the foreach below. There is no reason to initialize all variables in the top of the method.

tests/src/Context/JoinupSearchContext.php
```diff
@@ -311,6 +347,39 @@ public function assertSelectFacetOptionSelected(string $option, string $select):
     }
   }
 
+  /**
+   * Asserts a selected option in the select facets form.
+   *
+   * @param string $option
+   *   Text value of the option to find.
+   * @param string $select
+   *   CSS selector of the select field.
```
idimopoulos, line 356: This is not really a CSS selector, is it? It is more of a title of the field.

idimopoulos (changes requested):

idimopoulos (approved):

PR #2573, opened 2021-08-26
ISAICP-6595 Add swipe functionality to Explore block.

gingapa:


pfrenssen:
This works great in small and medium sized devices, but when I use screens of size `lg` and larger (>=992 px wide) the scrolling doesn't behave naturally: when I swipe the rightmost item all the way to the left, the slider doesn't get repositioned on this item, but instead scrolls back to the 2nd item.

This effect gets worse when the screen gets bigger - if I am on an `xl` screen of >= 1200px then there are 5 or more items shown. If I scroll the 5th item all the way to the left, then the slider almost completely reverses my action.

On sizes `sm` and `md` this is not the case, the scrolling works naturally.

pfrenssen:
Thanks to the input of @theosuja and @gingapa I have made some tweaks. It is not easy to get this working fully naturally on both touch and mouse input devices, but this is a good compromise. I put some details about in the commit message.

This now looks good from a technical perspective. Including in the acceptance environment so it can be approved by the business analyst team.

pfrenssen (approved):

PR #2631, opened 2021-10-20
ISAICP-6765: “Highlighted Solution” will become “Highlighted Content”.

saidatom:


pfrenssen:
Let's also make sure that the entity queue only accepts nodes of types `news`, `document` and `event`. It should not accept `discussion` or `video` for example.

